PR #613, opened 2024-01-13
fix: fix keyboard onChange not change value

### LINKED ISSUE #601: `onChangeEnd` of `Slider` or `RangeSlider` is not executed

@yamada-ui/react

### Package Version

1.1.4

### Description

When I manipulated the value of `Slider` or `RangeSlider` with the keyboard, I expected `onChangeEnd` to be executed every time the value changed, but it did not.

### Link to Reproduction

https://hirotomoyamada.github.io/yamada-ui/?path=/story/components-forms-slider--on-change-end

### Steps to Reproduce

1. Change the value with your keyboard.
2. `onChangeEnd` is not executed.

### Browser

Chrome

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

Yes

None

### Additional Information

```tsx
useUpdateEffect(() => {
  const { eventSource, values } = latestRef.current

  if (eventSource === "keyboard") onChangeEnd(values)
}, [values, onChangeEnd])
```

It seems that the cause is that `eventSource` is always `null`.

quangnmwork:
Closes #601 

## Description

> Fix #601  for Slider component

## Current behavior (updates)

> no

## New behavior

> no

> no

## Additional Information

**Root cause:**

Obsviously,main cause is the `eventSource` in  `useUpdateEffect` is null. 

It happen because when value change (or component is updated). The value of `latestRef` will be re-init. 

So that's why we need create a useState for eventSource value for preventing re-init data.

**After apply state for eventSource :**

https://github.com/hirotomoyamada/yamada-ui/assets/78476413/386a7e8e-b190-4e94-b9eb-7a4847259d91


hirotomoyamada:
@quangnmwork 

Thanks for the PR.

I believe using `useRef` is appropriate for this update since it does not affect rendering. What are your thoughts on this?

Additionally, if you switch to `useRef`, please name it `eventSourceRef`.

I also think that the `RangeSlider` component will need a similar update.

quangnmwork:
@hirotomoyamada 

Hey thanks for the reply. Seem like I have found where the bug. 

It's look like it's cause from the `useLatestRef`

```
const useLatestRef = <T extends any>(value: T) => {
  const ref = useRef<T | null>(null)

  ref.current = value

  return ref as MutableRefObject<T>
}
```

When value change : 

```
useLatestRef({
    min,
    max,
    step,
    value,
    isInteractive,
    eventSource: null as 'pointer' | 'keyboard' | null,
    focusThumbOnChange,
  })
```

Then this code `ref.current = value` will be executed. Then eventSource will be assign to null again that's why in useUpdateEffect the value of event source is always null.

So I think create a computed state as `computedValue` will solve this. 

Or , we will create something like this `const sourceRef = useRef({ eventSource: '' })` for controlling the event source value. 

What do you think about it? . I will be happy if you have any feedback about my reply. 🙇

hirotomoyamada:
@quangnmwork 

Hey, thanks for the response.

I agree with your approach.

Could you remove `eventSource` from `useLatestRef` and create a new `eventSourceRef` to control it?

```tsx
const eventSourceRef = useRef<"pointer" | "keyboard" | null>(null)
```

```tsx
eventSourceRef.current = "pointer"
```

```tsx
  useUpdateEffect(() => {
    const { values } = latestRef.current

    if (eventSourceRef.current === "keyboard") onChangeEnd(values)
  }, [values, onChangeEnd])
```

quangnmwork:
@hirotomoyamada Hi, I have updated the PR. 

Also in the range slider `useUpdateEffect` I found that the values is always mutate and cause maximum call stacks size like this:

image

I have fixed that and change into `[values[0], values[1], onChangeEnd]`.

hirotomoyamada:
thanks!

PR #771, opened 2024-02-21
fix(pin-input): some fixes about the asynchronous tests

koralle:
## Description

Applied the following three fixes to the asynchronous tests in `@yamada-ui/pin-input`:

1. Added the necessary `await act(async() => {...})` where updates to React's state were missing.
3. Added the missing `await waitFor(() => {...})` for assertions in`test(async () => {})` statements.
4. Replaced with `getByXXXX` with `findByXXXX` for asynchronous tests.


taroj1205 (approved):
Approved

PR #874, opened 2024-02-29
build: remove `@swc/core` from workspace-root

koralle:
## Description

No packages depend on `@swc/core` directly, so I removed it.


hoshico (approved):
LGTM

PR #1004, opened 2024-03-19
Stop uploading test coverages to Codecov temporarily

koralle:
## Description

There are many files with coverage not exceeding 80%, so we are not fully benefiting from Codecov. On the contrary, it is having a negative impact on the development speed. I would like to suspend the uploads of coverage to Codecov until the tests are adequately in place.


imaimai17468 (approved):
I think that's good.
I'll leave it up to you if you want to leave it commented out or remove it altogether.

PR #1007, opened 2024-03-19
[Storybook] Upgrade to v8

koralle:
## Description

Upgraded Storybook to v8.

Please run `pnpm storybook` and confirm that Storybook launches.

Yes (Only Storybook. There is no impact on the build.)

## Additional Information

https://storybook.js.org/docs/migration-guide


imaimai17468 (approved):
We have confirmed that the storybook works.
I think it is the correct code formatting for the other updated files as well.

PR #1063, opened 2024-03-25
fix(scroll-area): scroll-area is now displayed correctly in Safari

### LINKED ISSUE #757: ScrollArea does not hide scroll-bar on Safari with type prop "hover" and "scroll"

@yamada-ui/react

### Package Version

1.2.5

### Description

The same bug as this issue occurs in safari.

#745

https://github.com/hirotomoyamada/yamada-ui/assets/60034520/75d8aef9-7141-41c6-91ef-4f4aba22edf5

### Link to Reproduction

https://hirotomoyamada.github.io/yamada-ui/?path=/story/components-data-display-scrollarea--with-type

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

Safari

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

None

None

### Additional Information

_No response_

AVAM16:
Closes #757 

## Description

The bug was that in Safari the scrollbar did not appear when type==="hover" and the scroll area was hovered, and not appear when type==="scroll" and the scroll area was scrolled. The cause of this issue is that Safari does not visually update the css of the page.

## Current behavior (updates)

I had to force a re-render of the page, assure that the scroll position did not reset to 0 after the re-render and avoid infinite loops when type==="scroll".

## New behavior

The scrollbar now behaves correctly in the Safari Browser


AVAM16:
The test is probably not going to work due to the new infinite-scroll-area class.  The new exports in react/src/index.ts should be removed in order for pnpm test to work.

.changeset/calm-books-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@yamada-ui/scroll-area": major
```
setodeve, line 2: Could you fix like `"@yamada-ui/scroll-area": patch`?

setodeve:
> The test is probably not going to work due to the new infinite-scroll-area class. The new exports in react/src/index.ts should be removed in order for pnpm test to work.

Thanks for contributing!
Please ignore the matter of pnpm test for now.

setodeve (approved):

PR #1132, opened 2024-04-04
fix(chart): apply `className` to the chart component

### LINKED ISSUE #1076: Attach `className` to the chart component
### Description

Due to the rechart specification, some components do not have a className attached.
- `Cursor`
- `RadarChart` component
  - `PolarGrid`
  - `Dot`
  - `PolarRadiusAxis`
    - `tick` of `PolarRadiusAxis`
  - `PolarAngleAxis`
    - `tick` of `PolarAngleAxis`

`className` is not attached and cannot be styled.

### Proposed Solution or API

This will be fixed when the `rechart v2.12.4 ` is released.

### Alternatives

_No response_

None

### Additional Information

_No response_

108yen:
Closes #1076 

## Description

Due to the rechart specification, some components do not have a className attached.
`rechart v2.12.4` has been released with the bug fixed and the style applied with `className`
In addition, the `strokeDasharray` functionality has been added to `RadarChart`.

## Current behavior (updates)

Some components cannot pass a `className` and therefore cannot be styled.

## New behavior

- It can be styled by `className`
- `strokeDasharray` prop is available in `RadarChart`

No


packages/components/charts/src/area-chart.tsx
illionillion, line 1: @108yen 

How about writing it this way?

```diff
-- {withLegend ? (
++ {withLegend && (
  <Legend
    content={({ payload }) => (
      <ChartLegend
        className="ui-area-chart__legend"
        payload={payload}
        onHighlight={setHighlightedArea}
        {...computedLegendProps}
      />
    )}
    {...getLegendProps()}
  />
-- ) : null}
++ )}
```

108yen: no.
Do not use `&&` outside of conditional statements.

illionillion: Is it because the behavior changes depending on whether null or false is returned when the bool value is false?

PR #1137, opened 2024-04-05
fix: fixed Use theme tokens for styling values in Reorder component

### LINKED ISSUE #1095: If a theme is applied, use the theme's token in the `Reorder` component

@yamada-ui/react

### Package Version

1.3.0

### Description

Currently, styling values in components ignore theme tokens. If a theme is applied, you should use the theme's token.

Yamada UI's style system provides fallback functions. If the value of the token in the first argument does not exist, it returns the second argument, the fallback value.

```diff
const {
  className,
  orientation = "vertical",
-  gap = "1rem",
+  gap = "fallback(4, 1rem)",
  onChange,
  onCompleteChange,
  children,
  ...rest
} = omitThemeProps(mergedProps)
```

### Link to Reproduction

_No response_

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

None

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

Yes

None

### Additional Information

_No response_

PoPodada:
Closes #1095

## Description

The Reorder component was previously ignoring theme tokens when applying  styling values. This commit address the issue by ensuring that theme tokens are used when a theme is applied.

## Current behavior (updates)

The Reorder component was previously ignoring theme tokens when applying styling values.

## New behavior

 Yamada UI's style system's fallback functions are utilized. If the value of the token in the first argument does not exist, it falls back to the second argument, which serves as the fallback value.

No

### DIFF
--- a/.changeset/two-ears-listen.md
+++ b/.changeset/two-ears-listen.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/reorder": patch
+---
+
+change gap value in reorder component
--- a/packages/components/reorder/src/reorder.tsx
+++ b/packages/components/reorder/src/reorder.tsx
@@ -64,7 +64,7 @@ export const Reorder = forwardRef<HTMLUListElement, ReorderProps>(
     const {
       className,
       orientation = "vertical",
-      gap = "1rem",
+      gap = "fallback(4,1rem)",
       onChange,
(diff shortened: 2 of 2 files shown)


setodeve (approved):
Thanks you for contributing !
Approved !

PR #1172, opened 2024-04-09
fix: theme-token-closeButton

### LINKED ISSUE #1083: If a theme is applied, use the theme's token in the `CloseButton` component

@yamada-ui/react

### Package Version

1.3.0

### Description

Currently, styling values in components ignore theme tokens. If a theme is applied, you should use the theme's token.

Yamada UI's style system provides fallback functions. If the value of the token in the first argument does not exist, it returns the second argument, the fallback value.

```diff
position: "relative",
overflow: "hidden",
outline: 0,
display: "flex",
justifyContent: "center",
alignItems: "center",
flexShrink: 0,
...styles,
...__css,
- ...(isRounded ? { borderRadius: "9999px" } : {}),
+ ...(isRounded ? { borderRadius: "fallback(full, 9999px)" } : {}),
```

### Link to Reproduction

_No response_

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

None

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

Yes

None

### Additional Information

_No response_

Hinata0607:
Closes #1083

## Description

If a theme is applied, that use the theme's token.

## Current behavior (updates)

If the value of the token in the first argument does not exist, it returns the second argument, the fallback value.

## New behavior

If the value of the token in the first argument does not exist, it returns the second argument, the fallback value.

No


nakayan5:
@Hinata0607 
this change is not major, patch is right

nakayan5 (approved):
@Hinata0607 
great! thanks

Issue #1229, opened 2024-04-19
Stories are not updated even after updating the `theme` folder


hirotomoyamada:
### Which package has an issue?

@yamada-ui/react

### Package Version

1.3.4

### Description

The story does not update even if you change the style of a file in the `theme` folder, for example a component.

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

Chrome

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Do you know how to fix the issue?

None

### Are you willing to participate in fixing this issue and create a pull request with the fix?

None

hirotomoyamada:
https://github.com/storybookjs/storybook/issues/26881

PR #1230, opened 2024-04-19
Force reload when specific folder is updated

hirotomoyamada:
Related #1229

## Description

Force reload when specific folder is updated.

This is a temporary measure. :(

No


imaimai17468 (approved):
LGTM

PR #1239, opened 2024-04-20
Simplified a11y rules configuration

hirotomoyamada:
## Description

Simplified a11y rules configuration.

No


taroj1205 (approved):
LGTM

PR #1116, opened 2024-04-01
[Markdown] Ensure rendering Alert component

koralle:
Closes #379 

## Description

This PR will enable the Note syntax to be rendered as an Alert component even when line breaks are used within it.

## Current behavior (updates)

When I wrote Markdown like the following, it did not render correctly as an Alert component.

```markdown
:::note 

This is the first text.

This is the second text.

:::
```

## New behavior

The above Markdown renders as intended;

image

## Is this a breaking change (Yes/No):

Yes

PR #1526, opened 2024-05-13
Fix/use dedicated methods for `null`, `undefined` checks

koralle:
## Description

* When I created PR #1116  previously, I recalled that `isNull`/ `isUndefined` were not being used for checking null and undefined.
* In this PR, I have fixed it to use `isNull` / `isUndefined`.


komura-c (approved):
LGTM!

PR #1606, opened 2024-05-19
fix: use `@testing-library/user-event` to implement keyboard event in `Menu` component test

### LINKED ISSUE #1605: Focus event is not working as expected in the `Menu` component test

@yamada-ui/react

### Package Version

1.1.4

### Description

When I execute the component test on main branch, some of the `Menu` component tests that use keyboard events fail.
(There's a high possibility that the failure happens only in my environment as it's on main branch.)

- packages/components/menu/tests/menu.test.tsx > <Menu /> > keydown events for Enter
- packages/components/menu/tests/menu.test.tsx > <Menu /> > keydown events for ArrowUp
- packages/components/menu/tests/menu.test.tsx > <Menu /> > keydown events for ArrowDown

image

### Link to Reproduction

_No response_

### Steps to Reproduce

1. Go to './'
2. execute `pnpm install`
3. execute `pnpm test menu`

### Browser

None

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

None

None

### Additional Information

As additional information, I got `menuButton` focused at least when I used `menuButton.focus()` instead of using fireEvent in the above code, like so: (If `fireEvent.focus(menuButton)` is used, even `menuButton` won't be considered as focused by Vitest.)
```tsx
  test("keydown events for ArrowDown", async () => {
    const { getByRole } = render(
      <Menu>
        <MenuButton>Menu</MenuButton>
        <MenuList>
          <MenuItem>Add item</MenuItem>
          <MenuItem>Edit item</MenuItem>
        </MenuList>
      </Menu>,
    )

    const menuButton = getByRole("button", { name: "Menu" })

    // instead of fireEvent.focus(menuButton), 
    menuButton.focus()
    await act(() => fireEvent.keyDown(menuButton, { key: "ArrowDown" }))
    expect(screen.getByText("Add item")).toHaveFocus()
  })
```

However, I also could clarify that the focus event is happening as expected in browser, like so:

https://github.com/yamada-ui/yamada-ui/assets/74392116/993abffe-f6c9-4f6b-9193-45a46037fc20

Anyhow, I'm still investigating why this happens at least in my environment. I'm not fully confident that this happens to everyone, so it'd be awesome if someone gives information in this issue.

saku-1101:
Closes #1605

## Description

This fix lets `Menu` component test passed as this updated way of testing can correctly detect `tab-index=-1` is not focusable.

## Current behavior (updates)

> Please describe the current behavior that you are modifying

- Execute `pnpm test menu`
- The test result will be:
image

- One of the test case's failed results shows:
image

- This means that even the element that has `tab-index=-1` attribute can be focused with the current test code, which is `<ul />`.
- However, if it isn't in the testing environment, the element which has `tab-index=-1` attribute is correctly ignored, and `Menu` component is working correctly in the production, so the test code is the one to be correted.
 
## New behavior

> Please describe the behavior or changes this PR adds

- With the change of this PR, the element which has `tab-index=-1` attribute will be successfully ignored since the test code now uses `@testing-library/user-event` to implement keyboard event instead of using fireEvent from `@testing-library/react`.

image

No

## Additional Information
Nothing.

### DIFF
--- a/.changeset/blue-monkeys-sniff.md
+++ b/.changeset/blue-monkeys-sniff.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/menu": major
+---
+
+Fix `Menu` component test
--- a/packages/components/menu/tests/menu.test.tsx
+++ b/packages/components/menu/tests/menu.test.tsx
@@ -234,7 +234,7 @@ describe("<Menu />", () => {
   })
 
   test("keydown events for ArrowDown", async () => {
-    const { getByRole } = render(
+    const { user } = render(
       <Menu>
         <MenuButton>Menu</MenuButton>
         <MenuList>
@@ -244,16 +244,15 @@ describe("<Menu />", () => {
       </Menu>,
     )
 
-    const menuButton = getByRole("button", { name: "Menu" })
-
-    fireEvent.focus(menuButton)
-
-    await act(() => fireEvent.keyDown(menuButton, { key: "ArrowDown" }))
+    // focus the menu button
+    await user.keyboard("[Tab]")
+    // open the menu
+    await user.keyboard("[ArrowDown]")
     expect(screen.getByText("Add item")).toHaveFocus()
   })
 
   test("keydown events for ArrowUp", async () => {
-    const { getByRole } = render(
+    const { user } = render(
       <Menu>
         <MenuButton>Menu</MenuButton>
         <MenuList>
@@ -263,16 +262,15 @@ describe("<Menu />", () => {
       </Menu>,
     )
 
-    const menuButton = getByRole("button", { name: "Menu" })
-
-    fireEvent.focus(menuButton)
-
-    await act(() => fireEvent.keyDown(menuButton, { key: "ArrowUp" }))
+    // focus the menu button
+    await user.keyboard("[Tab]")
+    // open the menu
+    await user.keyboard("[ArrowUp]")
     expect(screen.getByText("Edit item")).toHaveFocus()
   })
 
   test("keydown events for Enter", async () => {
-    const { getByRole } = render(
+    const { user } = render(
       <Menu>
         <MenuButton>Menu</MenuButton>
         <MenuList>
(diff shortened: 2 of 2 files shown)


.changeset/blue-monkeys-sniff.md
108yen, line 1: No `changeset` is required when changing test files.

108yen (changes requested):
`user` should not be used to identify elements. Could you consider other methods?

.changeset/blue-monkeys-sniff.md
saku-1101: @108yen 

Do you mind asking why `user` is not supposed to be used in this case??
In my very humble opinion, `user-event` is not for identifying elements, but for simulating a user's workflow, so I didn't try to get any element here. Just simulating how DOM is actually updated in the browser.😅

https://github.com/yamada-ui/yamada-ui/issues/1605#issuecomment-2119259994

saku-1101:
> `user` should not be used to identify elements. Could you consider other methods?

@108yen 

Thank you for your quick review!
Do you mind asking why user is not supposed to be used in this case??
In my very humble opinion, user-event is not for identifying elements, but for simulating a user's workflow, so I didn't try to get any element here. Just simulating how DOM is actually updated in the browser.😅

https://github.com/yamada-ui/yamada-ui/issues/1605#issuecomment-2119259994

108yen:
@saku-1101 
The code does not guarantee that the `menu` gets focus.
```tsx
    // focus the menu button
    await user.keyboard("[Tab]")
```

saku-1101:
@108yen 
That is also quite right. I'll try to sort it out.

saku-1101:
@108yen 
Hi, sorry for the response got delayed. 
The `MenuButton` element is retrieved and the focus is guaranteed to be hit with respect to the retrieved MenuButton.
In this exact state, the keyboard event is fired, which makes sense for a keyboard test on the `MenuButton.`

https://github.com/yamada-ui/yamada-ui/pull/1606/commits/0e0d46779afbdf15823b03e20d60b7bbcff26d8a

108yen (approved):
Thank you!

PR #1715, opened 2024-06-01
fix: remove render and container

illionillion:
Closes # 

## Description

Remove render and container.


quantumshiro:
What is the issues number for this?

quantumshiro (approved):

PR #1973, opened 2024-06-06
Update `@yamada-ui/tooltip` accesibility

ayanami77:
Closes #1634 

## Description

- update `@yamada-ui/tooltip` accessibility to meat the accessibility standards in the W3C docs.
## Current behavior (updates)

- `@yamada-ui/tooltip` currently does not meet the accessibility standards.

### checklist
Keyboard Interaction

- [x] Escape: Dismisses the Tooltip.

WAI-ARIA Roles, States, and Properties

- [x] The element that serves as the tooltip container has role [tooltip](https://w3c.github.io/aria/#tooltip).
- [x] The element that triggers the tooltip references the tooltip element with [aria-describedby](https://w3c.github.io/aria/#aria-describedby).

## Is this a breaking change (Yes/No):

No

PR #2006, opened 2024-06-07
Update `@yamada-ui/tooltip` accessibility

### LINKED ISSUE #1634: Update `@yamada-ui/tooltip` accessibility
### Description

his issue aims to address accessibility deficiencies in our tooltip component as per W3C guidelines.

Keyboard Interaction

- [ ] Escape: Dismisses the Tooltip.

- [ ] The element that serves as the tooltip container has role [tooltip](https://w3c.github.io/aria/#tooltip).
- [ ] The element that triggers the tooltip references the tooltip element with [aria-describedby](https://w3c.github.io/aria/#aria-describedby).

###  Problem Statement/Justification
none

###  Proposed Solution or API
No response

###  Alternatives
none 

None

###  Additional Information
No response

ayanami77:
Closes #1634 

## Description

- update `@yamada-ui/tooltip` accessibility to meat the accessibility standards in the W3C docs.
## Current behavior (updates)

- `@yamada-ui/tooltip` currently does not meet the accessibility standards.

### checklist
Keyboard Interaction

- [x] Escape: Dismisses the Tooltip.

- [x] The element that serves as the tooltip container has role [tooltip](https://w3c.github.io/aria/#tooltip).
- [x] The element that triggers the tooltip references the tooltip element with [aria-describedby](https://w3c.github.io/aria/#aria-describedby).

### memo
- In #1973, I faced a problem that `ARIA attribute element ID` does not exist initially on the DOM because it is rendered after the trigger and then appears. So, I solved this problem by enclosing `<AnimatePresence />` with a `<ui.div/>`  and adding `role`, `id` and `aria-lable` to it. 
- I also refered to smarthr-design-system's tooltip. (https://febd1f3--63d0ccabb5d2dd29825524ab.chromatic.com/iframe.html?id=data-display%EF%BC%88%E3%83%87%E3%83%BC%E3%82%BF%E8%A1%A8%E7%A4%BA%EF%BC%89-tooltip--all&viewMode=story)

No


packages/components/tooltip/src/tooltip.tsx
```diff
@@ -364,34 +379,36 @@ export const Tooltip = forwardRef<TooltipProps, "div">(
...
 
-        <AnimatePresence>
-          {isOpen ? (
-            <Portal isDisabled={!withPortal} {...portalProps}>
-              <ui.div
-                {...getPopperProps()}
-                zIndex={resolvedZIndex}
-                pointerEvents="none"
-              >
+        <ui.div id={tooltipContentId} role="tooltip" aria-label={`${label}`}>
```
108yen, line 382: It is added outside of the `Tooltip` , is that ok?

ayanami77: I cannot determine whether it is right or not, but if we see that element `<ui.div>` as also composing the `Tooltip`, it may be right.

108yen: I do not think this is correct. Can`t it be implemented at the `Portal` location?

ayanami77: You mean like this?
```tsx
 <Portal id={tooltipContentId} role="tooltip" aria-label={`${label}`} isDisabled={!withPortal} {...portalProps}>
    // content
 </Portal>
```

108yen: Yes, if possible.

ayanami77: I understand your point, but even if I implement it like that, the problem will not be resolved. It means the `<Portal />` does not exist from the beginning because of the conditional rendering, and these `id={tooltipContentId} role="tooltip" aria-label={`${label}`}` attributes does not exist either. So, in this PR, I take this approach.

image

108yen: Discuss what to do with the accessibility team.

ayanami77: Ok, Thanks!

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -364,6 +379,10 @@ export const Tooltip = forwardRef<TooltipProps, "div">(
       <>
         {trigger}
 
+        <ui.span id={tooltipContentId} hidden>
```
koralle, line 382: @ayanami77 

Sorry, I realized this after showing you the method, but you shouldn't use the `hidden` attribute solely for controlling the visibility of elements. Instead, you should use `visibility: hidden` in CSS.

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -364,6 +379,10 @@ export const Tooltip = forwardRef<TooltipProps, "div">(
       <>
         {trigger}
 
+        <ui.span id={tooltipContentId} __css={{ visibility: "hidden" }}>
```
koralle, line 382: @ayanami77 
Sorry, my mistake. `visibility: hidden` affects the layout, so the correct one is `display: none`.

koralle (approved):

108yen:
@ayanami77 
I don't have a grasp of the content, but is it no longer necessary to give labels to `portals`?

ayanami77:
@108yen 
Sorry,  what does `giving labels to portals` mean? 
Something like this?
```tsx
 <Portal id={tooltipContentId} role="tooltip" aria-label={`${label}`} isDisabled={!withPortal} {...portalProps}>
    // content
 </Portal>
```

108yen:
@ayanami77 
Yes, I am. You were originally going to do this, weren't you?
Is it no longer needed?

ayanami77:
@108yen 
Ideally, the tooltip should be what I was originally going to do. On the internet, there are a lot of examples of that. However, when it comes to following the implementation of the yamada-ui tooltip component, I think it is almost impossible. 
Of course, I chated with the member of the accessibilty team about this problem. One of them advised me to place dummy `<span />` to meet the accessibilty. I followed his advice.

108yen:
@ayanami77 
Thanks for letting me know!

108yen (approved):

PR #2121, opened 2024-06-17
Enhance `<Switch >` component accessibilities

### LINKED ISSUE #1725: Update `@yamada-ui/switch` accessibility
### Description

This issue aims to address accessibility deficiencies in our @yamada-ui/switch component as per W3C guidelines.

- [ ] The switch has role [switch](https://w3c.github.io/aria/#switch).
- [ ]  The switch has an accessible label provided by one of the following:
  - [ ] A visible label referenced by the value of [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set on the element with role switch.
  - [ ] [aria-label](https://w3c.github.io/aria/#aria-label) set on the element with role switch.
- [ ] When on, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to true.
- [ ] When off, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to false.
- [ ] If a set of switches is presented as a logical group with a visible label, either:
  - [ ] The switches are included in an element with role [group](https://w3c.github.io/aria/#group) that has the property [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set to the ID of the element containing the group label.
  - [ ] The set is contained in an HTML fieldset and the label for the set is contained in an HTML legend element.
- [ ] If the presentation includes additional descriptive static text relevant to a switch or switch group, the switch or switch group has the property [aria-describedby](https://w3c.github.io/aria/#aria-describedby) set to the ID of the element containing the description.

Improve accessibility

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

koralle:
Closes #1725

## Description

* [ ] add `switch` role
* [ ] add `aria-checked` attribute


packages/components/switch/src/switch.tsx
```diff
@@ -116,6 +116,8 @@ export const Switch = forwardRef<SwitchProps, "input">((props, ref) => {
       {...rest}
     >
       <ui.input
+        role="switch"
+        aria-checked={isChecked}
```
108yen, line 120: Shouldn't this also be in `getInputProps`?

koralle: @108yen 

If we do that, I'd like to create `useSwitch` based on `useCheckbox`, and add `role` and `aria-checked` there. What do you think?

108yen: @koralle 
Sorry. I missed that you are using `useCheckbox`.
Is it possible to put it in `useCheckbox` if only `aria-checked={isChecked}`?

koralle: @108yen 

In the `Checkbox` component, the `aria-checked` attribute can have an intermediate value (`mixed`). On the other hand, this does not exist in the `Switch` component. Therefore, I don't think it's appropriate to apply `useCheckbox`, which includes the declaration of the `aria-checked` attribute, to the `Switch` component.

> * When checked, the checkbox element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to `true`.
> * When not checked, it has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to `false`.
> * When partially checked, it has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to `mixed`.

https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/

108yen: Thanks for the reply.
This seems to be the only way to do it this time. Approved.

108yen (approved):

zonoryo17 (approved):

PR #2126, opened 2024-06-18
Updated `framer-motion`

hirotomoyamada:
## Description

Updated dependencies.

No


hirotomoyamada:
@108yen @kurakke @taroj1205 @hirayamahiroto 

Fixed tests.

packages/components/modal/src/dialog.tsx
108yen, line 1: Is `changeset` of `Modal` unnecessary?

108yen (changes requested):
Do not include multiple assertions inside `waitFor`. Please put only the first one inside `waitFor`.
If you want to make `getxxx` asynchronous, use `findxxx`.


Are you sure that `forwardRef` is imported from `react` and not `@yamada-ui/core`?

packages/components/modal/src/dialog.tsx
hirotomoyamada: This does not change the internal logic, it just resolves the TypeScript error. So there is no need for `changeset`.

hirotomoyamada:
@108yen 

This time, the `forwardRef` will be imported from React. The reason is that we want to genericize the component type.

hirotomoyamada:
@108yen 

Fixed tests.

108yen (dismissed):

kurakke (dismissed):

taroj1205 (dismissed):

kurakke (approved):

108yen (approved):

PR #2189, opened 2024-06-28
Updated documentation

hirotomoyamada:
Closes #


108yen (commented):
Does `pnpm-lock.yml` need to be included?

hirotomoyamada:
@108yen 

Yes, you must include the change. The reason is that `. /docs` `package.json` in `docs` uses `workspace:*` and you won't notice it, but it has been changed properly.

108yen (approved):

PR #2204, opened 2024-06-30
Enhance the accessibility of MonthPicker `@yamada-ui/calendar` component

### LINKED ISSUE #1733: Enhance the accessibility of `MonthPicker` (`@yamada-ui/calendar`) component
### 1. Accessibility roles

* [x] Please refer to the image.

month-picker

### 2. `aria-controls`

* [x] An element with the combobox role must have in its aria-controls attribute the ID of the corresponding popup element (the element with the dialog role).

### 3. `aria-haspopup`

* [x] You must set `aria-haspopup="dialog"` on elements with the `combobox` role.

### 4. `aria-expanded`

* [x] When an element with the `dialog` role is hidden, the element with the `combobox` role should have the `aria-expanded` attribute set to `false`.
* [x] When an element with the `dialog` role is visible, the element with the `combobox` role should have the `aria-expanded` attribute set to `true`.

~~### 5. `aria-activedescendant`~~

* [ ] ~~While a descendant element of the element with the `dialog` role is focused, the element with the `combobox` role must have the ID of the focused element as the value of its `aria-activedescendant` attribute.~~

### 6. `aria-label` / `aria-labelledby` (on the element with `combobox` role)

* [x] Please refer to the following flowchart.

```
Does the combobox have a visible label?
    ├──> No: `combobox` element has a label provided by `aria-label` . 
    └──> Yes
           └──> Is the `combobox` element an HTML element that can be labelled using the HTML `label` element (e.g., the input element)?
                  ├──> No: `combobox` element has `aria-labelledby` set to a value that refers to the labelling element. 
                  └──> Yes: `combobox` is labeled using the `label` element. 
```

### 7. `aria-modals`

* [x] You must set aria-modal="true" on elements with the dialog role.

### 8. `aria-live`

* [x] Please refer to the image.

month-picker_1

### 9. `aria-selected`

* [x] You must add `aria-selected="true"` to the cell of the currently selected date (which is implemented as a `button` element in Yamada UI).

#### When the focus is within the popup...

* [x] If possible, it would be better to clear the current value instead of accepting it when the `Space` key is pressed.

## References

* [Date Picker Combobox Example | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/)
* [MonthPicker - Yamada UI](https://yamada-ui.com/components/forms/month-picker)
* [ARIA ライブリージョン - アクセシビリティ | MDN](https://developer.mozilla.org/ja/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)

We should prioritize web accessibility to ensure our UI components are usable by everyone, including people with disabilities. This inclusivity enhances user experience and complies with legal standards, broadening Yamada UI's user base.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

cidkumagai:
Closes #1733  

## Description

Enhance the accessibility of MonthPicker `@yamada-ui/calendar` component


108yen (commented):
It affects more than just `MonthPicker`, is that a problem?

cidkumagai:
@108yen 
Thank you comment.🙇‍♂️
I've checked the scope of the impact, and it's something that other issues will address, so I don't think it's a problem.
Corrections have been made where they should have been made.

packages/components/calendar/src/use-calendar-picker.ts
```diff
@@ -425,11 +426,14 @@ export const useCalendarPicker = <T extends UseCalendarProps<any>>(
       return {
         ref: mergeRefs(inputRef, ref),
         tabIndex: !allowInput ? 0 : -1,
+        role: "combobox",
         ...props,
+        "aria-haspopup": "dialog",
+        "aria-controls": calendarId,
```
108yen, line 432: It is better to list it above `props`.

108yen (changes requested):

108yen (approved):

packages/components/calendar/src/use-calendar-picker.ts
```diff
@@ -234,6 +234,7 @@ export const useCalendarPicker = <T extends UseCalendarProps<any>>(
 
   const containerRef = useRef<HTMLDivElement>(null)
   const inputRef = useRef<HTMLInputElement>(null)
+  const calendarId = useId()
```
koralle, line 237: @cidkumagai 

It's okay to call useId() internally to generate an ID only when id is undefined.

```ts
 id ??= useId()
```

koralle: or 

```ts
const calendarId = id ?? useId()
```

koralle (approved):

hirotomoyamada (approved):

PR #2235, opened 2024-07-03
Corrected regex in `isUnit` function and added test

hirotomoyamada:
## Description

Corrected regex in `isUnit` function and added test.

No

### DIFF
--- a/packages/utils/tests/assertion.test.ts
+++ b/packages/utils/tests/assertion.test.ts
@@ -148,6 +148,7 @@ describe("Assertion", () => {
     test("Returns true for strings containing units", () => {
       expect(isUnit("100px")).toBeTruthy()
       expect(isUnit("5rem")).toBeTruthy()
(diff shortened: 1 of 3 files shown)


packages/utils/tests/assertion.test.ts
```diff
@@ -148,6 +148,7 @@ describe("Assertion", () => {
     test("Returns true for strings containing units", () => {
       expect(isUnit("100px")).toBeTruthy()
       expect(isUnit("5rem")).toBeTruthy()
+      expect(isUnit("2xl")).toBeFalsy()
     })
```
Tomoya-Matsubara, line 152: I wonder if adding test code with `toBeFalsy()` to this test may cause a contradiction to its name: 

> Returns **true** for strings containing units

Maybe creating a new test case with a name like `Returns false for strings containing invalid units` is better.

Tomoya-Matsubara (dismissed):

zonoryo17 (approved):

PR #2252, opened 2024-07-06
Enhance the accessibility of `RangeDatePicker` (`@yamada-ui/calendar`) component

### LINKED ISSUE #1920: Enhance the accessibility of `RangeDatePicker` (`@yamada-ui/calendar`) component
### 1. Accessibility roles

* [x] Please refer to the image.

yamada-ui-calendar

### 2. `aria-controls`

* [x] An element with the `combobox` role must have in its `aria-controls` attribute the ID of the corresponding popup element (the element with the `dialog` role).

### 3. `aria-haspopup`

* [x] You must set `aria-haspopup="dialog"` on elements with the `combobox` role.

### 4. `aria-expanded`

* [x] When an element with the `dialog` role is hidden, the element with the `combobox` role should have the `aria-expanded` attribute set to `false`.
* [x] When an element with the `dialog` role is visible, the element with the `combobox` role should have the `aria-expanded` attribute set to `true`.

~~### 5. `aria-activedescendant`~~

* [ ] ~~While a descendant element of the element with the `dialog` role is focused, the element with the `combobox` role must have the ID of the focused element as the value of its `aria-activedescendant` attribute~~.

### 6. `aria-label` / `aria-labelledby` (on the element with `combobox` role)

* [x] Please refer to the following flowchart.

```
Does the combobox have a visible label?
    ├──> No: `combobox` element has a label provided by `aria-label` . 
    └──> Yes
           └──> Is the `combobox` element an HTML element that can be labelled using the HTML `label` element (e.g., the input element)?
                  ├──> No: `combobox` element has `aria-labelledby` set to a value that refers to the labelling element. 
                  └──> Yes: `combobox` is labeled using the `label` element. 
```

### 7. `aria-modals`

* [x] You must set aria-modal="true" on elements with the dialog role.

### 8. `aria-live`

* [x] Please refer to the image.

yamada-ui-calendar_2

### 9. `aria-selected`

* [x] You must add `aria-selected="true"` to the cell of the currently selected date (which is implemented as a button element in Yamada UI).

### 10. `aria-multiselectable`

* [x] You must set the `aria-multiselectable="true"` attribute on elements with the `grid` role.

#### When the focus is within the popup...

* [ ] If possible, it would be better to clear the current value instead of accepting it when the `Space` key is pressed.

## References

* [Date Picker Combobox Example | APG | WAI | W3C](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-datepicker/)
* [RangeDatePicker - Yamada UI](https://yamada-ui.com/components/forms/range-date-picker)
* [ARIA ライブリージョン - アクセシビリティ | MDN](https://developer.mozilla.org/ja/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)
* [Element: ariaMultiSelectable プロパティ - Web API | MDN](https://developer.mozilla.org/ja/docs/Web/API/Element/ariaMultiSelectable)

We should prioritize web accessibility to ensure our UI components are usable by everyone, including people with disabilities. This inclusivity enhances user experience and complies with legal standards, broadening Yamada UI's user base.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

cidkumagai:
Closes #1920  

## Description

Enhance the accessibility of `RangeDatePicker` (`@yamada-ui/calendar`) component


packages/components/calendar/src/range-date-picker.tsx
```diff
@@ -139,6 +140,7 @@ export const RangeDatePicker = forwardRef<RangeDatePickerProps, "input">(
               <RangeDatePickerField
                 separator={separator}
                 value={value}
+                aria-label="RangeDatePicker"
```
mokun632, line 143: Is this `aria-label` necessary?

mokun632: If it is necessary, the wording would be better suited to everyday speech, with the expectation that it will be read out to a screen reader.

packages/components/calendar/src/range-date-picker.tsx
```diff
@@ -157,6 +159,10 @@ export const RangeDatePicker = forwardRef<RangeDatePickerProps, "input">(
 
             <Portal {...portalProps}>
               <PopoverContent
+                id={id}
```
mokun632, line 162: I was wondering if it would be better to include these areas in the `getCalendarProps` of the `useRangeDatePicker` if possible.

packages/components/calendar/src/range-date-picker.tsx
cidkumagai: Thank you for your comment.
I will certainly delete the aria-label as it doesn't seem to fit.

packages/components/calendar/src/range-date-picker.tsx
cidkumagai: I've addressed this in another issue, but it seems to need to be granted here

https://github.com/yamada-ui/yamada-ui/pull/2204

mokun632: Thanks for reference.
If i think about it, this is a popover props, not a calendar, so it shouldn't be in getCalendarProps.
Sorry.

mokun632 (approved):
:lgtm:

hirotomoyamada (approved):

PR #2267, opened 2024-07-11
fixed faild `motion` and `transitions` tests

108yen:
## Description

The test checked in the `toHaveStyle` assertion is failing, so the `waitFor` enclosed
## Current behavior (updates)

NIL
## New behavior

NIL

No


108yen:
Fixed in another PR.
#2268

PR #2557, opened 2024-08-14
fix: active year's button after click on month-picker's header

quangnmwork:
Closes  https://github.com/yamada-ui/yamada-ui/issues/2425

## Description

Active year button  after click on month-picker's header .

## Current behavior (updates)

Currently , Year picker is not active after clicking .

## New behavior

After click on year picker , the year picker button will be actived . 

Also if we don't have the previous month value ,it will be set as January . Otherwise will be set the choosen month value .  

## Additional Information

https://github.com/user-attachments/assets/98a0572c-7a59-4f4c-b36d-cc5f830c2e0b

### DIFF
--- a/packages/components/calendar/src/use-month-picker.ts
+++ b/packages/components/calendar/src/use-month-picker.ts
@@ -83,6 +83,8 @@ export const useMonthPicker = ({
     defaultValue: defaultType,
     onChange: onChangeTypeProp,
   })
+  const [isClickOnYearListAlready, setIsClickOnYearListAlready] =
+    useState<boolean>(false)
 
   const {
     inputRef,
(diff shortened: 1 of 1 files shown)


packages/components/calendar/src/use-month-picker.ts
```diff
@@ -111,11 +113,26 @@ export const useMonthPicker = ({
...
+          if (isClickOnYearListAlready) {
+            const prevMonth = value?.getMonth()
+
+            const currentValue: Date = new Date(year, prevMonth || 0)
+            const inputValue = dateToString(currentValue)
+
+            setValue(currentValue)
+            setInputValue(inputValue)
+          }
+        }
```
quangnmwork, line 129: Basiclly , we already have base value month , year when focus on Month picker Header .

So I wonder if this way is valid .

The main idea is that I will set value only after Button Year is clicked .

hirotomoyamada:
@quangnmwork 

Hi, thanks for the PR!

This #2584 was a late create, but it was a best practice, so I merged it. I'm sorry for creating this PR, but I'll close this.

Issue #2496, opened 2024-08-11
The `Menu button` children repeats mount and unmount


taroj1205:
### Which package has an issue?

@yamada-ui/react

### Package Version

1.4.9

### Description

If `<Avatar />` is in the `<MenuButton />, it re-renders every time it is clicked.

This is due to the children of `<MenuButton />` mounting and unmounting every time it is triggered.

### Steps to Reproduce

Go to https://yamada-ui.com/components/overlay/menu#usage and copy and paste the code below.

```tsx
<Menu>
  <MenuButton as={Button} rightIcon={<ChevronDown size="xl" />}>
<Avatar src="https://avatars.githubusercontent.com/u/84060430?v=4" />
  </MenuButton>

  <MenuList>
    <MenuItem>Set status</MenuItem>
    <MenuItem>Edit Profile</MenuItem>
    <MenuItem>Preferences</MenuItem>
  </MenuList>
</Menu>
```

Click on the button and see the `<Avatar />` reload.

### Browser

Chrome

### Operating System

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

### Do you know how to fix the issue?

None

### Are you willing to participate in fixing this issue and create a pull request with the fix?

None

### Additional Information

```tsx
    const Component = As ? ui(As) : Button
```

Line 61 in `packages/components/menu/src/menu-button.tsx`.

PR #2596, opened 2024-08-19
fix: improve inefficient MenuButton's render 

quangnmwork:
Closes https://github.com/yamada-ui/yamada-ui/issues/2496 

## Description

Prevent MenuButton component mounting and unmounting every time it is triggered.

No

## Additional Information

Root cause : 
The Component is rendered every time when triggered . So we need memorize it . 

So the Component is now memoized using useMemo to ensure that it is only re-created when the As prop changes. This optimization prevents unnecessary re-renders and improves performance.

Video confirmed . The avatar is now  rendered only one.

https://github.com/user-attachments/assets/bf2695a7-bce5-4ca6-8800-f9d8aa0bbc95


bearone236 (approved):

hirotomoyamada (approved):

hirotomoyamada:
Thanks!

Issue #1737, opened 2024-06-02
Update `@yamada-ui/button` accessibility


setodeve:
### Description
This issue aims to address accessibility deficiencies in our `button component` as per W3C guidelines. 
The component currently does not meet the accessibility standards outlined in the W3C documentation. 

The goal is to ensure the component is accessible to all users, including those using assistive technologies.
Ensure that all unchecked items in this checklist should be addressed.

#### General Requirements
- [x] **Button Role:**
  - [x] The button element has `role="button"`.

- [x] **Accessible Label:**
  - [x] The button has an accessible label computed from its text content, `aria-labelledby`, or `aria-label`.

- [x] **Button Description:**
  - [x] If a description of the button's function is present, the button element has `aria-describedby` set to the ID of the element containing the description.

- [x] **Disabled State:**
  - [x] When the action associated with a button is unavailable, the button has `aria-disabled="true"`.

#### Toggle Button Specific Requirements
- [x] **Toggle State:**
  - [x] The button has an `aria-pressed` attribute.
  - [x] When toggled on, `aria-pressed` is set to `true`.
  - [x] When toggled off, `aria-pressed` is set to `false`.
  - [x] The label does not change when the state changes.

#### Menu Button Specific Requirements
- [x] **Menu Button Role:**
  - [x] The button has `aria-haspopup` set to either `menu` or `true`.

#### Keyboard Interaction
- [x] **Focus Behavior:**
  - [x] When the button has focus:
    - [x] `Space` activates the button.
    - [x] `Enter` activates the button.

- [ ] **Post-Activation Focus:**
  - [x] If the button opens a dialog, focus moves inside the dialog.
  - [ ] If the button closes a dialog, focus typically returns to the button that opened the dialog.
  - [x] If the button's action does not dismiss the current context, focus remains on the button.
  - [ ] If the button's action indicates a context change, focus moves to the starting point for that action.
  - [ ] If activated with a shortcut key, focus usually remains in the context from which the shortcut key was activated.

### Problem Statement/Justification

none

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

None

### References

https://www.w3.org/WAI/ARIA/apg/patterns/button/

chida09:
>Disabled State:
When the action associated with a button is unavailable, the button has aria-disabled="true".

As few other libraries have adopted aria-disabled for buttons, we will close this issue as not requiring further action.
https://github.com/yamada-ui/yamada-ui/pull/2736

[...]

PR #2736, opened 2024-09-07
feat: add aria-disabled to button component

chida09:
ref #1737

## Description

Added aria-disabled to the button component and updated the test.

No

## Additional Information
Due to other requirements, the issue will not be closed after the merge of this PR.

### DIFF
--- a/packages/components/button/tests/button.test.tsx
+++ b/packages/components/button/tests/button.test.tsx
@@ -122,5 +122,6 @@ describe("<Button />", () => {
 
     const buttonAsDiv = getByTestId("btn")
     expect(buttonAsDiv).toHaveAttribute("disabled")
+    expect(buttonAsDiv).toHaveAttribute("aria-disabled")
   })
 })
--- a/.changeset/nine-dogs-mate.md
+++ b/.changeset/nine-dogs-mate.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/button": minor
+---
+
+Enhance button component accessibility
--- a/packages/components/button/src/button.tsx
+++ b/packages/components/button/src/button.tsx
@@ -159,6 +159,7 @@ export const Button = forwardRef<ButtonProps, "button">(
     return (
       <ui.button
         ref={mergeRefs(ref, buttonRef)}
+        aria-disabled={isDisabled ? true : undefined}
         as={as}
         className={cx("ui-button", className)}
         type={type ?? defaultType}


108yen (changes requested):
Please chack quality.

packages/components/button/tests/button.test.tsx
```diff
@@ -105,6 +105,24 @@ describe("<Button />", () => {
...
+    rerender(
+      <Button as="div" data-testid="btn">
+        Valid Button
+      </Button>,
+    )
+
+    const buttonAsDiv = getByTestId("btn")
+    expect(buttonAsDiv).not.toHaveAttribute("disabled")
+    expect(buttonAsDiv).not.toHaveAttribute("aria-disabled")
+  })
```
108yen, line 124: This test is unnecessary.

108yen (commented):
I don't see anything in the issue to add `aria-disabled`, is this change correct?
Have you discussed this with the accessibility team?

packages/components/button/tests/button.test.tsx
chida09: Why is this test code considered unnecessary?

chida09: @108yen 

>Disabled State:
 When the action associated with a button is unavailable, the button has aria-disabled="true".

Isn't this included in the requirements?
I understood that we were supposed to address the items with unchecked checkboxes.

108yen: @chida09 
I think `aria-disabled` tests are covered in one test below.
If I am mistaken, please point it out.

> Why is this test code considered unnecessary?
I think it's a good idea to explicitly test that the button is enabled.

Sorry, I misunderstood that the checked items were the ones that should be corrected!

> Isn't this included in the requirements?
> I understood that we were supposed to address the items with unchecked checkboxes.
> 
> In our team, I've been told to work on issues that seem feasible to start with.

chida09: @108yen 
>I think aria-disabled tests are covered in one test below.
If I am mistaken, please point it out.

Of course, adding a test case for the presence of aria-disabled when the button is disabled would satisfy the minimum requirement.
It seems we might have different approaches to test coverage. My intention was to create explicit test cases for both the presence of aria-disabled when the button is disabled and its absence when it is enabled.

Since this may be a matter of team policy, I will proceed as per your review, understanding that no additional test cases are needed at this time.

>Sorry, I misunderstood that the checked items were the ones that should be corrected!

I see. Thanks for checking and confirming.

108yen (dismissed):
Thanks for the explanation!
If you are implementing the “add aria-disabled only if the role is a button” feature in the button component, then I think you are right to add that test.

Approved.

hirotomoyamada:
@chida09 @setodeve 

Hi, thanks for the PR.

When you set `disabled` on a button element, the `aria-disabled` does not read. Is this really a meaningful fix?

- https://www.reddit.com/r/reactjs/comments/18hc7pz/should_you_use_ariadisabled_or_disabled_for/
- https://css-tricks.com/making-disabled-buttons-more-inclusive/

In the process of correcting this, did you look at these URL articles and others?

I feel that this revision lacks discussion and thought.

chida09:
@hirotomoyamada 
Thank you for your feedback.
I'll review the changes and submit for your review again.

chida09:
As few other libraries have adopted aria-disabled for buttons, we will close this issue as not requiring further action.

https://react-spectrum.adobe.com/react-spectrum/Button.html#disabled
https://mui.com/base-ui/react-button/
https://ant.design/components/button

PR #2760, opened 2024-09-09
Fixing the type of `SelectItemWithItems` and adding `labelProps`

### LINKED ISSUE #2759: Fixing the type of SelectItemWithItems and adding labelProps

@yamada-ui/react

### Package Version

1.5.0

### Description

Fixing the type of `SelectItemWithItems` and adding `labelProps`.

```tsx
<OptionGroup
  key={i}
  label={label ?? ""}
  {...(props as HTMLUIProps<"ul">)} // Avoid type conversion here.
>
  {items.map(({ label, value, ...props }, i) => (
    <Option key={i} value={value} {...props}>
      {label}
    </Option>
  ))}
</OptionGroup>
```

Related #2720 
Allow the style to be applied to the label as well, like this issue.

### Link to Reproduction

_No response_

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

None

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

_No response_

illionillion:
Closes #2759


packages/components/select/tests/select.test.tsx
108yen, line 1: The purpose of the test is different, so please separate them.

packages/components/select/src/use-select.ts
108yen, line 1: The `labelProps` looks like nothing is being processed, does this need to be put in here?

108yen (changes requested):

packages/components/select/src/use-select.ts
illionillion: Thanks I fixed

illionillion:
@108yen I fixed. Please review again.

packages/components/select/tests/select.test.tsx
```diff
@@ -117,6 +117,25 @@ describe("<Select />", () => {
...
+        </Select>,
+      )
+
+      const input = await screen.findByRole("combobox")
+
+      await user.click(input)
+
+      await expect(screen.findByText("Numbers")).resolves.toHaveStyle({
+        fontSize: "12px",
+      })
```
108yen, line 136: `await expect` probably doesn't make sense.
Please fix.
```tsx
const group = await screen.findByText("Numbers")
      expect(group).toHaveStyle({
        fontSize: "12px",
      })
```

packages/components/select/tests/select.test.tsx
108yen, line 1: And while you're at it, can you include `should be displayed properly when NOT present` in the `header rendering`?

108yen (changes requested):

packages/components/select/tests/select.test.tsx
illionillion: fixed

packages/components/select/tests/select.test.tsx
illionillion: Are you suggesting that we add a test to see if it works correctly even if we don't specify anything for the header property of Select?

108yen: @illionillion 
Please make this change!

```diff
  describe("header rendering", () => {
    test("should be displayed properly when present.", () => {
      const HEADER_TEXT = "Header"
      const HEADER_STYLE = { background: "green" }

      const renderHeader = () => {
        return <header style={HEADER_STYLE}>{HEADER_TEXT}</header>
      }

      render(
        <Select
          header={renderHeader()}
          placeholder="Select numbers"
          placeholderInOptions={true}
          items={[
            { label: "One", value: "" },
            { label: "Two", value: "" },
          ]}
        />,
      )

      const headerElement = screen.getByText(HEADER_TEXT)

      expect(headerElement).toBeInTheDocument()
      expect(headerElement).toHaveStyle(HEADER_STYLE)
      expect(headerElement.parentElement).toHaveClass("ui-select__header")
    })
-  })

  test("should be displayed properly when NOT present", () => {
    const { container } = render(
      <Select
        placeholder="Select numbers"
        placeholderInOptions={true}
        items={[
          { label: "One", value: "" },
          { label: "Two", value: "" },
        ]}
      />,
    )

    const headerElement = container.querySelector(".ui-select__header")
    expect(headerElement).toBeNull()
  })
+  })
```

illionillion: Thanks I fixed it.

illionillion:
Please review again. @108yen

packages/components/select/tests/select.test.tsx
108yen: @illionillion 
Wrong.

illionillion: Sorry, I fixed.
deleted duplicates.

108yen (approved):
Thank you!

packages/components/select/src/use-select.ts
```diff
@@ -792,6 +792,10 @@ export interface UseSelectOptionGroupProps extends HTMLUIProps<"ul"> {
    * The label of the option group.
    */
   label: string
+  /**
+   * Props for option group element.
+   */
+  labelProps?: HTMLUIProps<"span">
```
hirotomoyamada, line 798: Do not define props to customize components here.  Set to the `option-group.tsx` side.

packages/components/select/src/select.tsx
```diff
@@ -34,7 +34,7 @@ interface SelectItemWithValue extends OptionProps {
 
 interface SelectItemWithItems extends OptionGroupProps {
   items?: SelectItemWithValue[]
-}
+} & UseSelectOptionGroupProps
```
hirotomoyamada, line 37: This fix is no longer needed in #2803. Please remove this modification.

.changeset/sweet-shoes-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@yamada-ui/select": patch
+---
+
+Fixing the type of SelectItemWithItems and adding labelProps.
```
hirotomoyamada, line 5: Please correct this description as well.

https://github.com/yamada-ui/yamada-ui/pull/2760/files#r1762265693

packages/components/select/src/use-select.ts
illionillion: fixed

packages/components/select/src/select.tsx
illionillion: fixed

.changeset/sweet-shoes-lie.md
illionillion: fixed

hirotomoyamada (approved):

Issue #2796, opened 2024-09-12
Colors used in `DropzoneAccept` in `Dropzone` are wrong.


hirotomoyamada:
### Provide the exact quote of the error or issue

Colors used in `DropzoneAccept` in `Dropzone` are wrong.

The correct name is `"success"`.

```diff
<Dropzone accept={IMAGE_ACCEPT_TYPE} maxSize={3 * 1024 ** 2}>
  <HStack color={["blackAlpha.500", "whiteAlpha.500"]}>
    <DropzoneAccept>
-     <Upload fontSize="6xl" color="primary" />
+     <Upload fontSize="6xl" color="success" />
    </DropzoneAccept>

    <DropzoneReject>
      <X fontSize="6xl" color="danger" />
    </DropzoneReject>

    <DropzoneIdle>
      <ImageIcon fontSize="6xl" />
    </DropzoneIdle>

    <VStack gap="2xs">
      <Text fontSize="xl">Drag images here or click to select files</Text>
      <Text fontSize="sm">
        Attach as many files as you like, each file should not exceed 3mb
      </Text>
    </VStack>
  </HStack>
</Dropzone>
```

### Does the docs page already exist? Please link to it.

https://yamada-ui.com/components/forms/dropzone#display-according-to-status

### Are you willing to participate in fixing this issue and create a pull request with the fix?

None

KekmaTime:
@hirotomoyamada i would like to work on this issue

hirotomoyamada:
@KekmaTime 
Hi, than

PR #2797, opened 2024-09-13
(fix) : changed the color used in `DropzoneAccept` and `Dropzone` to …

KekmaTime:
Closes https://github.com/yamada-ui/yamada-ui/issues/2796

## Description
Colors used in `DropzoneAccept` in `Dropzone` are wrong.

The correct name is `success`.

## Current behavior (updates)

It is currently using `primary` which is wrong.

## New behavior

I've changed it from `primary` to `success`


hirotomoyamada (approved):

hirotomoyamada:
Thanks!

Issue #1959, opened 2024-06-05
Update `@yamada-ui/select` accessibility


nakayan5:
### Description

This issue aims to address accessibility deficiencies in our select component as per W3C guidelines.
The component currently does not meet the accessibility standards outlined in the W3C documentation.

The goal is to ensure the component is accessible to all users, including those using assistive technologies.
Ensure that all unchecked items in this checklist should be addressed.

Keyboard interaction: none

WAI-ARIA Roles, States, and Properties
- [x] An element that contains or owns all the listbox options has role [listbox](https://w3c.github.io/aria/#listbox).
- [x] Each option in the listbox has role [option](https://w3c.github.io/aria/#option) and is contained in or owned by either:
  - [x] The element with role listbox.
  - [x] An element with role [group](https://w3c.github.io/aria/#group) that is contained in or owned by the element with role listbox.
- [x] Options contained in a group are referred to as grouped options and form what is called an option group. If a listbox contains grouped options, then:
  - [x] All option groups contain at least one option.
  - [x] Each option group has an accessible name provided via [aria-label](https://w3c.github.io/aria/#aria-label) or [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby).
- [ ] If the element with role listbox is not part of another widget, such as a combobox, then it has either a visible label referenced by [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) or a value specified for [aria-label](https://w3c.github.io/aria/#aria-label).
- [x] If the listbox supports selection of more than one option, the element with role listbox has [aria-multiselectable](https://w3c.github.io/aria/#aria-multiselectable) set to true. Otherwise, aria-multiselectable is either set to false or the default value of false is implied.
- [x] The selection state of each selectable option is indicated with either [aria-selected](https://w3c.github.io/aria/#aria-selected) or [aria-checked](https://w3c.github.io/aria/#aria-checked):
  - [x] If the selection state is indicated with aria-selected, then aria-checked is not specified for any options. Alternatively, if the selection state is indicated with aria-checked, then aria-selected is not specified for any options. See notes below regarding considerations for which property to use and for details of the unusual conditions that might allow for both properties in the same listbox.
[...]

PR #2800, opened 2024-09-15
feat: useSelectList to support multi-select functionality

uraaaa24:
Related https://github.com/yamada-ui/yamada-ui/issues/1959

## Description

This update ensures that the aria-multiselectable attribute is explicitly set to true when the listbox supports multiple selections, and set to false when single selection is supported, improving accessibility.
The 1959 issue states.
>  If the listbox supports selection of more than one option, the element with role listbox has aria-multiselectable set to true. Otherwise, aria-multiselectable is either set to false or the default value of false is implied.

## Current behavior (updates)

The aria-multiselectable attribute is not visible even when multiple selections are supported.

## New behavior

The aria-multiselectable attribute is shown even if multiple selection is supported.
Not specifically displayed for single selections.

No


kaehehehe (approved):

neginasu (approved):

hirotomoyamada:
@uraaaa24 

Hi, thanks for the PR.

Is it safe to close the issue #1959?

uraaaa24:
@hirotomoyamada 

I expect to be able to close the issue once this pull request is captured, but I would like to have @setodeve  confirm that the issue is closed as well.
Please only merge this pull request once!

hirotomoyamada (approved):

PR #2926, opened 2024-09-23
Deprecated `isClick`, introduced `isClickable`

hirotomoyamada:
## Description

Deprecated `isClick`, introduced `isClickable`.

No

### DIFF
--- a/packages/components/input/src/input-element.tsx
+++ b/packages/components/input/src/input-element.tsx
@@ -8,8 +8,16 @@ interface InputElementOptions {
    * If `true`, the element clickable.
    *
    * @default false
+   *
+   * @deprecated Use `isClickable` instead.
    */
   isClick?: boolean
+  /**
+   * If `true`, the element clickable.
+   *
+   * @default false
+   */
+  isClickable?: boolean
   /**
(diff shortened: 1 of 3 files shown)


packages/components/input/src/input-element.tsx
```diff
@@ -21,9 +29,14 @@ interface InputElementOptions {
 export interface InputElementProps extends HTMLUIProps, InputElementOptions {}
 
 const InputElement = forwardRef<InputElementProps, "div">(
-  ({ className, isClick = false, placement = "left", ...rest }, ref) => {
+  (
+    { className, isClick = false, isClickable, placement = "left", ...rest },
```
zonoryo17, line 33: badge

Is the default argument for `isClickable` here not necessary?

hirotomoyamada: Hi, it is not necessary to set a default value here. If it is set, `isClick` will always be ignored.

zonoryo17: I see. I understand.

zonoryo17 (approved):
LGTM

PR #2955, opened 2024-09-24
fix : add displayName and __ui__ to select component

### LINKED ISSUE #2901: `displayName` is not set for the component of "@yamada-ui/select"
package: @yamada-ui/select

### Description
The component "@yamada-ui/select" uses `forwardRef`, but `displayName` is not set, making it difficult to see the display in [React Developer Tools](https://github.com/facebook/react/tree/main/packages/react-devtools).

### Additional Information

For example, like this.

```tsx
  ...
})

```

kiran0022:
Closes #2901

## Description

Set displayName for the @yamada-ui/select component to make debugging easier.
This change improves the ability to identify the component name in development tools.

## Current behavior (updates)

Currently, displayName is not set for the @yamada-ui/select component, making it difficult to identify the component in debugging tools like React Developer Tools.

## New behavior

With displayName set, the @yamada-ui/select component now displays its name correctly in debugging tools like React Developer Tools, making it easier for developers to identify and debug the component.

No

## Additional Information

None


illionillion (changes requested):
@kiran0022 

Thank for PR!!

Please all components of files in the same directory  other than select.tsx should be modified in the same way.

kiran0022:
@illionillion 

Thanks for the feedback! 
I've updated the related select components accordingly. Let me know if you need anything else.

.changeset/four-readers-doubt.md
illionillion, line 1: Multiple changesets are not needed for a single package change, and this is not a `major` change.
Please delete this.

packages/components/select/src/select.tsx
illionillion, line 1: The `SelectField` should be modified in the same way.

packages/components/select/src/select-icon.tsx
illionillion, line 1: The `SelectClearIcon` should be modified in the same way.

illionillion: import `FC` from `@yamada-ui/core`

packages/components/select/src/option.tsx
illionillion, line 1: Fix here too with all components.

packages/components/select/src/multi-select.tsx
illionillion, line 1: Fix here too with all components.

illionillion (changes requested):
@kiran0022 

Thanks for the correction.

Please apply the `displayName` and `__ui__` fixes to all components.

For normal `FC` components that do not use `forwardRef`, import the `FC` from `@yamada-ui/core` and modify it.

If you are told that there is no `FC` type in `@yamada-ui/core`, please sync the latest contents to the fork repository and merge it into your branch.

kiran0022:
@illionillion

Thanks for your guidance! 
I've addressed the change requests and updated the PR. Let me know if there's anything else you need!

hirotomoyamada:
Thanks!

PR #2959, opened 2024-09-25
refactor: update alpha-slider accessibility

### LINKED ISSUE #2084: Update `AlphaSlider`  accessibility
### Description

related https://github.com/yamada-ui/yamada-ui/issues/1505
related https://github.com/yamada-ui/yamada-ui/issues/1727

This issue aims to address accessibility deficiencies in our slider component as per W3C guidelines.

- [ ] If the value of aria-valuenow is not user-friendly, e.g., the day of the week is represented by a number, the [aria-valuetext](https://w3c.github.io/aria/#aria-valuetext) property is set to a string that makes the slider value understandable, e.g., "Monday".

improve accessibility

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

kaz-uen:
Closes #2084 

## Description

Add aria-valuetext to AlphaSlider to eliminate accessibility issues.
This response adds an aria-valuetext attribute to AplhaSlider to provide the current value to the user.

### DIFF
--- a/packages/components/color-picker/src/use-color-slider.ts
+++ b/packages/components/color-picker/src/use-color-slider.ts
@@ -103,6 +103,7 @@ export const useColorSlider = ({
     onChangeEnd: onChangeEndProp,
     thumbColor,
     style: styleProp,
+    "aria-valuetext": ariaValueText,
     ...rest
   } = useFormControlProps(props)
   const [
@@ -355,6 +356,7 @@ export const useColorSlider = ({
         "aria-valuenow": value,
         "aria-valuemin": min,
         "aria-valuemax": max,
+        "aria-valuetext": ariaValueText ?? value.toString(),
         "data-active": dataAttr(isDragging && focusThumbOnChange),
         onKeyDown: handlerAll(props.onKeyDown, onKeyDown),
         onFocus: handlerAll(props.onFocus, onFocusProp),
@@ -377,6 +379,7 @@ export const useColorSlider = ({
       onKeyDown,
       onFocusProp,
       onBlurProp,
+      ariaValueText,
     ],
   )
 
--- a/.changeset/blue-kids-worry.md
+++ b/.changeset/blue-kids-worry.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/color-picker": minor
+---
+
+add aria-valuetext property to AlphaSlider


packages/components/color-picker/src/use-color-slider.ts
```diff
@@ -355,6 +356,7 @@ export const useColorSlider = ({
         "aria-valuenow": value,
         "aria-valuemin": min,
         "aria-valuemax": max,
+        "aria-valuetext": ariaValueText ?? value.toString(),
```
setodeve, line 359: It seems that `ariaValueText` has the following characteristics, 
so it might be better to add more user-friendly wording.

https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaValueText
>The ariaValueText property of the [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) interface reflects the value of the [aria-valuetext](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuetext) attribute, which defines the human-readable text alternative of aria-valuenow for a range widget.

setodeve (changes requested):

packages/components/color-picker/src/use-color-slider.ts
setodeve: Saturation slider might be helpful for reference.

https://github.com/yamada-ui/yamada-ui/blob/2512e87fc92204b7e3d59b10cbe1ac07ddf2dc2a/packages/components/color-picker/src/use-saturation-slider.ts#L372

kaz-uen: @setodeve 
Thank you for confirming.
I will add user friendly wording.

kaz-uen: @setodeve 
Since it seems to be common for alpha values to indicate transparency from 0% to 100% using “%” as the unit, how about adding a % unit to the value by multiplying value by 100?

https://developer.mozilla.org/ja/docs/Web/CSS/opacity#alpha-value
https://ja.wikipedia.org/wiki/%E3%82%A2%E3%83%AB%E3%83%95%E3%82%A1%E5%80%A4_(%E3%83%87%E3%82%B8%E3%82%BF%E3%83%AB%E7%94%BB%E5%83%8F)

setodeve: “aria-valuetext = ‘50%’” is a good pattern!

hirotomoyamada:
No update, so we are closing.

PR #2979, opened 2024-09-27
feat: set display name and UI for markdown

### LINKED ISSUE #2881: `displayName` is not set for the component of "@yamada-ui/markdown"
package: @yamada-ui/markdown

### Description
The component "@yamada-ui/markdown" uses `forwardRef`, but `displayName` is not set, making it difficult to see the display in [React Developer Tools](https://github.com/facebook/react/tree/main/packages/react-devtools).

### Additional Information

For example, like this.

```tsx
  ...
})

```

0xtkey256:
Closes Closes #2881

## Description

set for the component of `@yamada-ui/markdown`


packages/components/markdown/src/markdown.tsx
cidkumagai, line 1: @0xtkey256 
I think the Code component also needs to be modified.

cidkumagai (changes requested):

packages/components/markdown/src/markdown.tsx
0xtkey256: @cidkumagai 
Thanks, I've set the `displayname` and `__ui__`, and also changed `FC` import from `@yamada-ui/core`. Please review the change.

hirotomoyamada (approved):

PR #2992, opened 2024-09-28
Update prop from `size` to `boxSize`

hirotomoyamada:
## Description

Update prop from `size` to `boxSize`.

No


zonoryo17 (approved):
LGTM

Issue #1728, opened 2024-06-02
Update `@yamada-ui/carousel` accessibility


setodeve:
### Description
This issue aims to address accessibility deficiencies in our `carousel component` as per W3C guidelines. 
The component currently does not meet the accessibility standards outlined in the W3C documentation. 

The goal is to ensure the component is accessible to all users, including those using assistive technologies.
Ensure that all unchecked items in this checklist should be addressed.

#### General Requirements
- **Visible Labels:**
  - [ ] Carousel container has an accessible label using `aria-labelledby` or `aria-label`.
  - [x] Label does not include the word "carousel".

- **Roles and Descriptions:**
  - [ ] Carousel container has `role="region"` or `role="group"`.
  - [ ] `aria-roledescription="carousel"` is set on the carousel container.
  - [ ] Each slide container has `role="group"` with `aria-roledescription="slide"`.

#### Rotation Control
- **Rotation Control Button:**
  - [x] Exists as a native button element or implements the button pattern.
  - [ ] Accessible label provided by `aria-label` or inner text.
  - [ ] Label changes based on action (e.g., "Stop slide rotation" or "Start slide rotation").
  - [x] Button does not use `aria-pressed`.

#### Slide Navigation
- **Next and Previous Slide Controls:**
  - [x] Exist as native button elements or implement the button pattern.
  - [x] Styled as arrows.
  - [ ] Do not move focus when activated.

- **Slide Picker Controls (if present):**
  - [x] Group of elements, often styled as small dots or tabs.
  - [x] Each control is a native button element or implements the button pattern.
  - [ ] Accessible name indicates the slide it will display.
  - [ ] Current slide control has `aria-disabled="true"`.

#### Keyboard Interaction
- **Tab and Shift + Tab Navigation:**
  - [x] Moves focus through carousel interactive elements.

- **Automatic Slide Rotation:**
  - [ ] Stops when any carousel element receives keyboard focus.
  - [ ] Does not resume unless the user activates the rotation control.
  - [x] Stops when the mouse hovers over the carousel.

#### WAI-ARIA Implementation
- **Basic Carousel Elements:**
  - [ ] Carousel container includes all components, with `role="region"` or `role="group"`.
  - [ ] Rotation, next slide, and previous slide controls are native button elements or implement the button pattern.
  - [ ] Each slide container has `role="group"` with `aria-roledescription="slide"`.
  - [ ] Each slide has an accessible name using `aria-labelledby` or `aria-label`.
[...]

PR #3036, opened 2024-10-03
add roles and descriptions to carousel component

uraaaa24:
Closes https://github.com/yamada-ui/yamada-ui/issues/1728

## Description

add a11y to `carousel`

https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-1-prev-next/
https://www.w3.org/WAI/ARIA/apg/patterns/carousel/

No


kaehehehe (approved):

hirotomoyamada (changes requested):
Set `aria-label`, `aria-live` and `aria-controls`.

hirotomoyamada:
@uraaaa24 

Please check quolity test and lint.

uraaaa24:
@hirotomoyamada 
Sorry for the late confirmation.
I've fixed it, please check back.

hirotomoyamada (changes requested):
The `id` must be unique.This will be violated if there are multiple `Carousels` on the screen.

packages/components/carousel/src/use-carousel.ts
```diff
@@ -494,7 +506,9 @@ export const useCarouselIndicators = () => {
       const isSelected = index === selectedIndex
 
       return {
```
hirotomoyamada, line 508: Is `aria-controls` not required?

packages/components/carousel/src/use-carousel.ts
```diff
@@ -366,7 +367,12 @@ export const useCarousel = ({
     (props = {}, ref = null) => ({
       ...containerProps,
       ...props,
+      id: "carousel-items",
```
hirotomoyamada, line 370: This should be set to `getSlidesProps`.

packages/components/carousel/src/use-carousel.ts
```diff
@@ -366,7 +367,12 @@ export const useCarousel = ({
...
       ref,
+      "aria-label": "Highlighted sections",
+      "aria-live": autoplay ? "off" : "polite",
```
hirotomoyamada, line 373: This should be set to `getSlidesProps`.

packages/components/carousel/src/use-carousel.ts
uraaaa24: I had `aria-controls` set by `getControlProps`.

```ts
  const getControlProps: PropGetter<"button"> = useCallback(
    (props = {}, ref = null) => ({
      ...props,
      ref,
      "aria-controls": "carousel-items",
      disabled,
      onClick: handlerAll(props.onClick, onClick),
    }),
    [disabled, onClick],
  )
```

Does this mean that this is also a mistake?

uraaaa24: Refer to the end of the table below.

https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-1-prev-next/#rps_label

hirotomoyamada: No, I think we need to set it on the control button as well and have `aria-controls` on the indicator as well. I thought so.

uraaaa24: Thanks for letting me know! I will respond that way!

uraaaa24:
@hirotomoyamada 
I have finished correcting the points you pointed out!
Please check it!

hirotomoyamada (dismissed):

hirotomoyamada (approved):

PR #3076, opened 2024-10-08
Updated hover and disabled styles

hirotomoyamada:
## Description

Updated hover and disabled styles.

No


odendayoko (approved):
@hirotomoyamada 
LGTM🧡
I see that we have to specify `_disabled` after `_hover` for it to work well...👀

PR #2740, opened 2024-09-08
[WIP] Implementation of the Swap component

kaehehehe:
Closes #2692

## Description

I'm submitting this PR to update Mr. Yamada. The component implementation is nearly complete, but I still need to start on the test code.

https://github.com/user-attachments/assets/0b8c4db7-e6d6-4002-bedd-e12dcb38d66f

PR #3084, opened 2024-10-10
Implementation of the Swap component

kaehehehe:
Closes #2740

## Description

I developed a new animation component using Motion.
Initially, we planned to create the Swap component, but we decided to break it down and develop three separate components: Airy, Rotate, and Flip.


kaehehehe:
fixed conflict! @hirotomoyamada

kaehehehe:
I will create a new PR.

PR #3139, opened 2024-10-16
Removed `"mono"` from SEMANTIC_COLOR_SCHEMES

hirotomoyamada:
## Description

Removed `"mono"` from SEMANTIC_COLOR_SCHEMES.

The `"mono"` value is not needed because it is not yet implemented.

No


zonoryo17 (approved):
LGTM

PR #3147, opened 2024-10-16
fix(tabs): added necessary aria attributes to Tab and TabPanel component

### LINKED ISSUE #1636: Update `@yamada-ui/tabs` accessibility
### Description

his issue aims to address accessibility deficiencies in our tabs component as per W3C guidelines.

- [x] tab
  - [x] When the tab list contains the focus, moves focus to the next element in the page tab sequence outside the tablist, which is the tabpanel unless the first element containing meaningful content inside the tabpanel is focusable.

- [x] When focus is on a tab in a tablist with either horizontal or vertical orientation:
  - [x] Space or Enter: Activates the tab if it was not activated automatically on focus.

- [x] If the tab list has a visible label, the element with role tablist has [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set to a value that refers to the labelling element. Otherwise, the tablist element has a label provided by [aria-label](https://w3c.github.io/aria/#aria-label).
- [ ] Each element with role tab has the property [aria-controls](https://w3c.github.io/aria/#aria-controls) referring to its associated tabpanel element.
- [x] The active tab element has the state [aria-selected](https://w3c.github.io/aria/#aria-selected) set to true and all other tab elements have it set to false.
- [ ] Each element with role tabpanel has the property [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) referring to its associated tab element.
- [x] If the tablist element is vertically oriented, it has the property [aria-orientation](https://w3c.github.io/aria/#aria-orientation) set to vertical. The default value of aria-orientation for a tablist element is horizontal.

none

### Proposed Solution or API

_No response_

### Alternatives

none

None

### Additional Information

_No response_

taiseiiiii:
Closes #1636

## Description

Added necessary aria attributes to Tab and TabPanel component

## Current behavior (updates)

Regarding Tab component, it doesn't have `id` and `aria-controls`
As for TabPanel, it doesn't have `id` and `aria-labelledby`

## New behavior

The Tab and TabPanel are now associated with each other.

No


packages/components/tabs/src/tabs.tsx
108yen, line 1: Is it not possible to pass only `uuid` to props?
We can generate the respective ids at the destination we pass them to.
The name `uuid` is rarely used. So please rename `id`.

108yen (changes requested):
Closes issue link is broken.

packages/components/tabs/src/tabs.tsx
taiseiiiii: @108yen 

> Closes issue link is broken.

Sorry, I have overlooked. I just correct it.

> Is it not possible to pass only uuid to props?

It implies only uuid should be defined in TabsContext?

> The name uuid is rarely used. So please rename id.

I consider `baseId` more preferable according to your feedback
Is it proper?

108yen: @taiseiiiii 
> It implies only uuid should be defined in TabsContext?

Yes. If it is possible.
> I consider baseId more preferable according to your feedback
> Is it proper?

That might still be a good idea.

taiseiiiii: @108yen 

> Yes. If it is possible.

Sorry, I still don't understand.
If you have any idea of this implementation, would it be possible to share a few example?
I think each props(tabIds, tabPanelIds) should be defined in TabsContext. 
It's because the Tab component and the TabPanel component need to reference each other's ids.

108yen: @taiseiiiiiii 
Why don't we do the same on the tabpanel side?

taiseiiiii: @108yen 
Sorry, I may have been mistaken.
In the beginning, I think this component should set the id from user Input.
Because, in the document of Tabs, ARIA Roles and Attributes section includes following statement.

https://yamada-ui.com/components/disclosure/tabs/accessibility

> The id used to associate with role="tabpanel".

That's so I implemented current source code.

Does that mean that such a requirement is not necessary in the first place?
```
<Tabs>
      <Tab id="tab-1">孫悟空</Tab>
      <Tab>ベジータ</Tab>
      <Tab>フリーザ</Tab>

      <TabPanel id="tabpanel-1">
        <Text>クリリンのことか……クリリンのことかーーーっ！！！！！</Text>
      </TabPanel>

~~~~~~~~
</Tabs>
```

108yen: @taiseiiiii 
If user passes id props, user must manage ids. We should only consider when user doesn't pass ids.
The way to manage IDs on the user side should be written in the documentation.

taiseiiiii: @108yen 
I finally understand!!!
I apologize for checking so many times.

I've just corrected according to your feedback.
Would it be possible to confirm again?

108yen:
@taiseiiiii 
Isn't it a problem that the tab id and the tab panel id are the same?

taiseiiiii:
@108yen 
Does the following approach seem suitable?
Add the suffix to the each id

`tab.tsx`
```
    const tabId = `${baseId}-${index}-tab`
    const tabpanelId = `${baseId}-${index}-tabpanel`

   return (
      <ui.button
        id={tabId}
        className={cx("ui-tabs__tab", className)}
        role="tab"
        __css={css}
        {...props}
        {...rest}
        type="button"
        aria-controls={tabpanelId}
```

`tab-panel.tsx`
```
    const tabpanelId = `${baseId}-${index}-tabpanel`
    const tabId = `${baseId}-${index}-tab`

    return (
      <ui.div
        id={tabpanelId}
        ref={ref}
        className={cx("ui-tabs__panel", className)}
        role="tabpanel"
        __css={css}
        {...rest}
        aria-labelledby={tabId}
```

108yen:
@taiseiiiii 
Yes

taiseiiiii:
@108yen 
I've just addressed your feedback.
Would it be possible to confirm modifications?

packages/components/tabs/src/tab-panel.tsx
```diff
@@ -27,11 +35,13 @@ export const TabPanel = forwardRef<TabPanelProps, "div">(
 
     return (
       <ui.div
+        id={tabpanelId}
         ref={ref}
         className={cx("ui-tabs__panel", className)}
         role="tabpanel"
         __css={css}
         {...rest}
+        aria-labelledby={tabId}
```
108yen, line 44: User can't change `aria-labelledby`.

packages/components/tabs/src/tab.tsx
```diff
@@ -69,12 +73,14 @@ export const Tab = forwardRef<TabProps, "button">(
...
     return (
       <ui.button
+        id={tabId}
         className={cx("ui-tabs__tab", className)}
         role="tab"
         __css={css}
         {...props}
         {...rest}
         type="button"
+        aria-controls={tabpanelId}
```
108yen, line 83: User can't change `aria-controls`.

108yen (changes requested):

taiseiiiii:
@108yen 
Sorry. I've just corrected.
Please review again.

108yen (dismissed):

108yen:
@taiseiiiii 
We were wrong.
We need to adjust `aria-controls` and `aria-labelledby` when users pass ids.

taiseiiiii:
@108yen 
It's means, should I modify according to my [initial commit](https://github.com/yamada-ui/yamada-ui/pull/3147/commits/540b69b543ec563c6d674cb16e6a977f86388a5c)?

In addition, I gonna modify it that add the suffix to each id in order to avoid duplicate tabId and tabPanelId.
Is this thought correct?

108yen:
@taiseiiiii 

When user passes id to `Tab`, we need to create `aria-labelledby` of `TabPanel` based on id passed by user.
Id of `TabPanel` should be the same.

I think it is necessary to suck the id up to the `Tabs` using `ref` or something and process it.

taiseiiiii:
@108yen 
I've just modified in order to handle each attribute according to user input.
Please review again.

taiseiiiii:
@taiseiiiii 
It works properly in my local environment.
Is it latest?

`tabs.stories.tsx`

```
export const basic: Story = () => {
  return (
    <Tabs>
      <Tab id="test">孫悟空</Tab>
      <Tab>ベジータ</Tab>
      <Tab>フリーザ</Tab>

      <TabPanel>
        <Text>クリリンのことか……クリリンのことかーーーっ！！！！！</Text>
      </TabPanel>
      <TabPanel>
        <Text>へっ！きたねぇ花火だ</Text>
      </TabPanel>
      <TabPanel>
        <Text>
          私の戦闘力は530000です。ですがもちろんフルパワーであなたと戦う気はありませんからご心配なく……
        </Text>
      </TabPanel>
    </Tabs>
  )
}
```

result

スクリーンショット 2024-10-21 19 33 38

I consider it proper, because "test" sets to aria-labelledby in first tabpanel component.

108yen (dismissed):

hirotomoyamada (changes requested):
If you set `TabList` or `TabPanels` to `children`, this logic is not supported.

It needs to be fundamentally rebuilt.

taiseiiiii:
@hirotomoyamada 
Thank you for your review.

I could only come up with the following way.
It seems to work properly.

If there are any other suitable ways to implement this, could you give me some tips?

Before
```
const tabPanelIds = cloneTabPanels.map(
  (panel, index) => panel.props.id ?? `${uuid}-${index}-tabpanel`,
)

const tabIds = cloneTabs.map(
  (tab, index) => tab.props.id ?? `${uuid}-${index}-tab`,
)
```

After
```
const tabPanelIds = (customTabPanels ? pickChildren(getValidChildren(customTabPanels.props.children), TabPanel) : cloneTabPanels).map(
  (panel, index) => panel.props.id ?? `${uuid}-${index}-tabpanel`,
)

const tabIds = (customTabList ? pickChildren(getValidChildren(customTabList.props.children), Tab) : cloneTabs).map(
  (tab, index) => tab.props.id ?? `${uuid}-${index}-tab`,
)
```

hirotomoyamada:
@taiseiiiii 

That is a bad approach from a performance standpoint.

To solve this problem, id must be transferred from the child to the parent.

hirotomoyamada (approved):

PR #3150, opened 2024-10-17
Implementation of the Swap component

kaehehehe:
Closes https://github.com/yamada-ui/yamada-ui/pull/2740

## Description

I developed a new animation component using Motion.
Initially, we planned to create the Swap component, but we decided to break it down and develop three separate components: Airy, Rotate, and Flip.


koralle:
@kaehehehe 

Great PR! 

However, there is one thing I would like to confirm. 

Since the `<button>` element is rendered in the browser, the `Name` attribute is essential for accessibility (e.g. by using `aria-label`, `aria-labelledby`, and more...). 
Should we require users to always pass it, or should we set default values for these components internally? 
Which specification will you adopt?

kaehehehe:
@koralle 
These components provide animations and can be used in various ways depending on the user's preferences. Therefore, I believe it is better to require users to specify the values rather than setting default values.

packages/components/transitions/package.json
```diff
@@ -54,11 +54,13 @@
...
     "@yamada-ui/use-value": "workspace:*",
-    "@yamada-ui/utils": "workspace:*"
+    "@yamada-ui/utils": "workspace:*",
+    "@yamada-ui/use-controllable-state": "workspace:*"
   },
   "devDependencies": {
     "clean-package": "2.2.0",
-    "react": "^18.3.1"
+    "react": "^18.3.1",
+    "@yamada-ui/lucide": "workspace:*"
```
hirotomoyamada, line 63: Remove `"@yamada-ui/lucide"`.

packages/components/transitions/tests/rotate.test.tsx
```diff
@@ -0,0 +1,63 @@
+import type { AiryIdent } from "../src"
+import { Apple, Cherry } from "@yamada-ui/lucide"
```
hirotomoyamada, line 2: Use `TestIcon` instead.

packages/components/transitions/tests/flip.test.tsx
```diff
@@ -0,0 +1,77 @@
+import type { AiryIdent } from "../src"
+import { Apple, Cherry } from "@yamada-ui/lucide"
```
hirotomoyamada, line 2: Use `TestIcon` instead.

packages/components/transitions/package.json
kaehehehe: [d24d14d](https://github.com/yamada-ui/yamada-ui/pull/3150/commits/d24d14d22e4b4ca42f15c41e5c541b9d8a31f3db)

packages/components/transitions/tests/flip.test.tsx
kaehehehe: [b6c6d30](https://github.com/yamada-ui/yamada-ui/pull/3150/commits/b6c6d30836d6d99f4c9458f0fd62fc3372076abf)

packages/components/transitions/tests/rotate.test.tsx
kaehehehe: [b6c6d30](https://github.com/yamada-ui/yamada-ui/pull/3150/commits/b6c6d30836d6d99f4c9458f0fd62fc3372076abf)

hirotomoyamada:
@kaehehehe 

Please summarize your commitments.

hirotomoyamada (approved):

Issue #2059, opened 2024-06-11
Enhance the accessibility of `MultiSelect` (@yamada-ui/select) component


koralle:
### 1. Accessibility roles

![Image](https://github.com/yamada-ui/yamada-ui/assets/33865215/2de94fc7-211d-4b20-9a1d-58bd2b5ba246)

### 2. `aria-controls`

* [x] An element with the `combobox` role must have in its `aria-controls` attribute the ID of the corresponding popup element (the element with the `listbox` role).

### 2. `aria-activedescendant`

* [x] When any element with the `option` role within an element with the `listbox` role receives focus, you need to set the ID of that element to the `aria-activedescendant` attribute of the element with the `combobox` role.

### 3. `aria-selected`

* [x] Set `aria-selected="true"` on the selected elements with the `option` role. For the elements with the option role that are not selected but can be selected, set `aria-selected="false"`. If an element cannot be selected, do not set the `aria-selected` attribute at all.

### 4. `aria-multiselectable`

* [x] You need to set `aria-multiselectable="true"` on the element with the `listbox` role to inform assistive technologies that multiple selection is possible.

## Keyboard Interaction

* [ ] When `Alt` + `Up Arrow` is pressed while the popup is displayed...
  * [ ] If an element within the popup is focused, return the focus to the element with the `combobox` role.
  * [ ] Close the popup.

## References

* [ ] [MultiSelect - Yamada UI](https://yamada-ui.com/components/forms/multi-select)

### Problem Statement/Justification

We should prioritize web accessibility to ensure our UI components are usable by everyone, including people with disabilities. This inclusivity enhances user experience and complies with legal standards, broadening Yamada UI's user base.

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

None

zonoryo17:
I'll work on this.

zonoryo17:
I had assigned it, but I will remove it once as I have been holding it too long.

chida09:
>  An element with the combobox role must have in its aria-controls attribute the ID of the corresponding popup element (the element with the listbox role).

Already supported.
<img width="1323" alt="スクリーンショット 2024-09-22 8 24 22" src="https://github.com/user-attachments/assets/885d8731-4847-4659-9da5-9c92686e9fae">

chida09:
>3. aria-selected
[...]

PR #3238, opened 2024-10-26
improve keyboard navigation and accessibility

uraaaa24:
Related https://github.com/yamada-ui/yamada-ui/issues/2059

## Description

It is written in `multiSelect` issue, but actually relates to `select` and `multiSelect`.

- Alt + ArrowUp
If the popup has focus, returns focus to the combobox, otherwise it closes the popup.
- Alt + ArrowDown
If the popup is available but not displayed, displays the popup without moving focus.

Other accessibility support will be provided in a separate PR.

No

### DIFF
--- a/.changeset/rich-trains-run.md
+++ b/.changeset/rich-trains-run.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/select": minor
+---
+
+improve keyboard navigation and accessibility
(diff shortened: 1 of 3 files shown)


hirotomoyamada (commented):
Why did you implement this keyboard operation? Where is the source?

uraaaa24:
@hirotomoyamada 
The following was used as a reference
I also implemented the issue as described in the issue.

https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#keyboardinteraction

uraaaa24:
@hirotomoyamada 
I have made the correction, please check back when you have a chance!

PR #3242, opened 2024-10-26
feat(menu): add `displayName` to components related with `Menu`

### LINKED ISSUE #2882: `displayName` is not set for the component of "@yamada-ui/menu"
package: @yamada-ui/menu

### Description
The component "@yamada-ui/menu" uses `forwardRef`, but `displayName` is not set, making it difficult to see the display in [React Developer Tools](https://github.com/facebook/react/tree/main/packages/react-devtools).

### Additional Information

For example, like this.

```tsx
  ...
})

```

koralle:
Closes #2882

### DIFF
--- a/packages/components/menu/src/context-menu.tsx
+++ b/packages/components/menu/src/context-menu.tsx
@@ -1,5 +1,4 @@
-import type { ThemeProps } from "@yamada-ui/core"
-import type { FC } from "react"
+import type { FC, ThemeProps } from "@yamada-ui/core"
 import type { MenuProps } from "./menu"
 import { omitThemeProps, useComponentMultiStyle } from "@yamada-ui/core"
 import { Menu } from "./menu"
@@ -24,3 +23,6 @@ export const ContextMenu: FC<ContextMenuProps> = (props) => {
     </ContextMenuProvider>
   )
 }
+
+ContextMenu.displayName = "ContextMenu"
+ContextMenu.__ui__ = "ContextMenu"
--- a/packages/components/menu/src/context-menu.tsx
+++ b/packages/components/menu/src/context-menu.tsx
@@ -24,5 +24,4 @@ export const ContextMenu: FC<ContextMenuProps> = (props) => {
   )
 }
 
-ContextMenu.displayName = "ContextMenu"
 ContextMenu.__ui__ = "ContextMenu"
--- a/packages/components/menu/src/menu-button.tsx
+++ b/packages/components/menu/src/menu-button.tsx
@@ -95,6 +95,9 @@ export const MenuButton = forwardRef<MenuButtonProps, "button">(
   },
 )
 
+MenuButton.displayName = "MenuButton"
+MenuButton.__ui__ = "MenuButton"
+
 const Button = forwardRef<MenuButtonProps, "button">((rest, ref) => {
   const { styles } = useMenu()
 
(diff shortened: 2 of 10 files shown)


packages/components/menu/src/context-menu.tsx
```diff
@@ -24,3 +23,6 @@ export const ContextMenu: FC<ContextMenuProps> = (props) => {
     </ContextMenuProvider>
   )
 }
+
+ContextMenu.displayName = "ContextMenu"
+ContextMenu.__ui__ = "ContextMenu"
```
kurakke, line 28: There’s no need to specify displayName for components that don't use forwardRef.

packages/components/menu/src/menu-item.tsx
```diff
@@ -345,3 +356,6 @@ const CheckIcon: FC = () => (
     />
   </svg>
 )
+
+CheckIcon.displayName = "CheckIcon"
+CheckIcon.__ui__ = "CheckIcon"
```
kurakke, line 361: There’s no need to specify displayName for components that don't use forwardRef.

packages/components/menu/src/menu-button.tsx
```diff
@@ -95,6 +95,9 @@ export const MenuButton = forwardRef<MenuButtonProps, "button">(
   },
 )
 
+MenuButton.displayName = "MenuButton"
+MenuButton.__ui__ = "MenuButton"
+
 const Button = forwardRef<MenuButtonProps, "button">((rest, ref) => {
   const { styles } = useMenu()
```
kurakke, line 103: It looks like Button is missing displayName and `__ui__`.
Since it uses forwardRef, Button.displayName = "Button" and Button.__ui__ = "Button" should be added to keep it consistent with other components.

packages/components/menu/src/menu.tsx
```diff
@@ -205,3 +204,6 @@ export const Menu: FC<MenuProps> = (props) => {
     </DescendantsContextProvider>
   )
 }
+
+Menu.displayName = "Menu"
+Menu.__ui__ = "Menu"
```
kurakke, line 209: There’s no need to specify displayName for components that don't use forwardRef.

packages/components/menu/src/context-menu.tsx
koralle: @hirotomoyamada 
> There’s no need to specify displayName for components that don't use forwardRef.

Is this right?

hirotomoyamada: @koralle 

Yes, correct.

However, there are unnecessary `displayName` everywhere in the Yamada UI, so I would like to revamp it somewhere.

packages/components/menu/src/menu-button.tsx
koralle: I heard from @hirotomoyamada that they are unnecessary in this case.

kurakke: Sorry, That's OK.

packages/components/menu/src/menu-item.tsx
koralle: @kurakke 
This was also the case with `Button`. Both `displayName` and `__ui__` were unnecessary cases, so I removed them.

kurakke: OK.

kurakke (approved):

hirotomoyamada (approved):

PR #3291, opened 2024-10-30
Udated insight

108yen:
## Description
Change initial period. 
Added this month and last month period options. 

No


docs/pages/insights/insights-header.tsx
```diff
@@ -281,7 +301,32 @@ const PeriodSelect: FC<PeriodSelectProps> = memo(() => {
...
+                bg: ["blackAlpha.200", "whiteAlpha.200"],
+              }}
+              onClick={() => onSuggestChange(value)}
+            >
+              {t(`insights.period.${value}`)}
+            </Center>
+          )
+        })}
+
+        <Divider />
```
hirotomoyamada, line 327: The color of the border is different from the others.

docs/pages/insights/insights-header.tsx
```diff
@@ -233,16 +239,30 @@ const PeriodSelect: FC<PeriodSelectProps> = memo(() => {
...
-        dayjs()
-          .tz()
-          .subtract(parseInt(count ?? ""), unit as ManipulateType)
-          .toDate(),
-        dayjs().tz().subtract(1, "d").toDate(),
-      ]
+    (type: InsightMonthPeriodSuggest | InsightRangePeriodSuggest) => {
+      let value: [Date, Date]
+
+      if (type == "this-month") {
```
hirotomoyamada, line 245: If it was the 1st of the month, the result is the same as `"last-month"`.

This is because the insight page is one day behind.
So if it is the 1st of the month, you need to hide or disable `"this-month"`.

docs/pages/insights/insights-header.tsx
108yen: @hirotomoyamada 
Which border color?

docs/pages/insights/insights-header.tsx
108yen: Fixed.

docs/pages/insights/insights-header.tsx
hirotomoyamada: It's `["blackAlpha.200", "whiteAlpha.100"]`.

108yen: @hirotomoyamada 
Fixed.

hirotomoyamada (approved):

PR #3319, opened 2024-11-01
test(rating): test coverage for custom color

### LINKED ISSUE #2358: Enhance Test Coverage for `@yamada-ui/rating`
package: @yamada-ui/rating

### Description
`@yamada-ui/rating` has a test coverage of less than 95%. We will enhance it to above 95%.

### Target files
- [packages/components/rating/src/rating-item.tsx](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages/components/rating/src/rating-item.tsx)
  - [L34](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages/components/rating/src/rating-item.tsx#L34)

Testing is essential to ensure quality when operating a package.

### Additional Information

nathanlao:
Closes #2358 

## Description

@yamada-ui/rating has a test coverage of less than 95%. We will enhance it to above 95%.

## Current behavior (updates)

No test cases for custom color of `<Rating />`

## New behavior

New test covers target files:

- [packages/components/rating/src/rating-item.tsx](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages/components/rating/src/rating-item.tsx)
	- [L34](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages/components/rating/src/rating-item.tsx#L34)

No

## Additional Information
When I test the color using `getComputedStyle`, it always retrieves an `undefined` color attribute. It seems that JSDOM does not process CSS custom properties, like `var(--ui-colors-yellow-500)`, so I had to test the CSS rules directly to see if the css rule exists and the color attributes are applying correctly.

### DIFF
--- a/packages/components/rating/tests/rating.test.tsx
+++ b/packages/components/rating/tests/rating.test.tsx
@@ -126,4 +126,37 @@ describe("<Rating />", () => {
     await user.click(items[4]!)
     expect(items[3]).not.toHaveAttribute("data-focus")
   })
+
+  test("should use custom color correctly", () => {
+    const getColor = (value: number) => {
+      switch (value) {
+        case 1:
+          return "red.500"
+        case 2:
+          return "orange.500"
+        case 3:
+          return "yellow.500"
+        case 4:
+          return "green.500"
+        case 5:
+          return "blue.500"
+        default:
+          return undefined
+      }
+    }
+    const { container } = render(<Rating color={getColor} />)
+    const items = container.querySelectorAll(".ui-rating__item")
+
+    const expectedColors = [
+      "red.500",
+      "orange.500",
+      "yellow.500",
+      "green.500",
+      "blue.500",
+    ]
+    expectedColors.forEach((expectedColor, index) => {
+      const item = items[index]
+      expect(item).toHaveStyle(`color: ${expectedColor}`)
+    })
+  })
 })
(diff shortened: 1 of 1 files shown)


packages/components/rating/tests/rating.test.tsx
```diff
@@ -126,4 +126,44 @@ describe("<Rating />", () => {
...
+        case 3:
+          return "yellow.500"
+        case 4:
+          return "green.500"
+        case 5:
+          return "blue.500"
+        default:
+          return undefined
+      }
+    }
```
kaehehehe, line 146: How about writing the code like this? (This is just my opinion :))

```tsx
  test.each([1, 2, 3, 4, 5])(
    "should use color for defaultValue %i",
    (defaultValue) => {
      const colors: Record<number, string> = {
        1: "red.500",
        2: "orange.500",
        3: "yellow.500",
        4: "green.500",
        5: "blue.500",
      }

      const getColor = (value: number): string | undefined => colors[value]

      // Below is the existing code.

      const { container } = render(
        <Rating color={getColor} defaultValue={defaultValue} />,
      )
  )
```

nathanlao: Thank you for the idea! However, when I used `test.each`, it failed on tests since different `defaultValue` is applied across the tests. Each test renders the component with a different `defaultValue`, it affects how the `data-filled` attributes and CSS classes. 

But I like the object type to handle color values and applied it on the tests 🙌

nathanlao: And using `Record<number, string>` didn't pass the ESlint, so I used the index signature.

kaehehehe (approved):

hirotomoyamada (approved):

PR #3333, opened 2024-11-04
fix: fixed so that `variant` is reflected even when `as={IconButton}`

### LINKED ISSUE #3332: `variant` does not work with `<FileButton as={IconButton}>`

@yamada-ui/react

### Package Version

1.6.1

### Description

`variant` does not work for `<FileButton as={IconButton}>` while variant works for `FileButton` or `IconButton` alone.

```tsx
<VStack alignItems="flex-start">
  <FileButton as={IconButton} icon={<Plus />} variant="ghost" />
  <IconButton icon={<Plus />} variant="ghost" />
</VStack>
```

image

### Link to Reproduction

_No response_

### Steps to Reproduce

2. Click on '...'
3. Scroll down to '...'

### Browser

None

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

None

None

### Additional Information

_No response_

illionillion:
Closes #3332 

## Description

```tsx
<VStack alignItems="flex-start">
  <FileButton as={IconButton} icon={<Plus />} variant="outline" />
  <IconButton icon={<Plus />} variant="outline" />
</VStack>
```

image


packages/components/file-button/src/file-button.tsx
illionillion, line 1: I don't think this is a very good method and would like to know if there are other suggestions for improvement.

Shunii85: I don't have any ideas, but I think it may be caused by the `omitThemeProps` function.

https://github.com/yamada-ui/yamada-ui/blob/3e790149b6dae6d2ffb0f15baab15536f64d5a4c/packages/core/src/theme.ts#L254-L259

The `variant` is omitted here.

koralle: @illionillion @Shunii85 
Indeed, it is correct that the `variant` property is omitted by `omitThemeProps`.
Looking at the other components, the `Loading` took out the `variant` prop as follows.

```tsx
// packages/components/loading/src/loading.tsx
const [
  { color, ...styles },
  { colorScheme, size = "1em", variant = "oval", ...mergedProps },
] = useComponentStyle("Loading", props)

```

illionillion: @koralle @Shunii85 
I rewroted it like this.

```tsx
const [styles, { variant, ...mergedProps }] = useComponentStyle(
    "FileButton",
    props,
  )
```

Shunii85 (dismissed):

hirotomoyamada (approved):

hirotomoyamada (approved):

hirotomoyamada:
@illionillion @Shunii85 

The fixes you have made have not improved to `size` and `colorScheme`.

Issue #3302, opened 2024-10-31
Rename `props`: `is*` eliminates all `is` prefix in the `Modal` component


nathanlao:
package: @yamada-ui/modal

### Description

Change and unify the names of props. 
- Removing all `is` prefix from `is*`

### Problem Statement/Justification

- For example, `isOpen` becomes `open` and `isDisabled` becomes `disabled`
- Unify `orientation` and `direction` to either one
- props to be changed are marked `@deprecated` in v1

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

None

MrRobot-fm:
Hi @nathanlao , can i take this task?

nathanlao:
Hi @MrRobot-fm, thank you for the comment.

I have assigned you and please create a working branch from v2.0.

Good luck!

MrRobot-fm:
@nathanlao 
I have to change only the props with the prefix `is` that was marked `@deprecated`?

nathanlao:
@MrRobot-fm Yes and I believe it includes the docs and demos as well if there are any.

PR #3381, opened 2024-11-08
feat(components): rename props from isOpen to open

nathanlao:
Related #3302

## Description

This pull request is to remove the `is` prefix for `isOpen` in the props

## Current behavior (updates)

Props in components were using `isOpen`

## New behavior

Props in components are now using `open`

No

## Additional Information
Since #3302 is quite large, and this PR is only to rename `isOpen`

### DIFF
--- a/packages/components/modal/src/drawer-content.tsx
+++ b/packages/components/modal/src/drawer-content.tsx
@@ -1,3 +1,5 @@
+/* eslint-disable perfectionist/sort-union-types */
+/* eslint-disable perfectionist/sort-jsx-props */
 import type { CSSUIObject, ThemeProps } from "@yamada-ui/core"
 import type { MotionPanInfo } from "@yamada-ui/motion"
 import type { Merge } from "@yamada-ui/utils"
@@ -14,7 +16,7 @@ import { useDrawer, useModal } from "./modal-context"
 
 export interface DrawerContentProps
   extends Merge<
-    Omit<DrawerProps, "isOpen" | "placement" | keyof ThemeProps>,
+    Omit<DrawerProps, "open" | "isOpen" | "placement" | keyof ThemeProps>,
     Required<
       Pick<
         DrawerProps,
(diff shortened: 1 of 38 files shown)


kaehehehe (commented):
Is it OK to merge into the main branch? Mayby this PR should be merged into 2.0 branch. @nathanlao

nathanlao:
> Is it OK to merge into the main branch? Mayby this PR should be merged into 2.0 branch. @nathanlao 

I think you are right! Just changed the target branch to 2.0

packages/components/tooltip/src/tooltip.tsx
```diff
@@ -181,7 +189,7 @@ export const Tooltip = motionForwardRef<TooltipProps, "div">(
       "Tooltip",
       props,
     )
-    const {
+    let {
```
kaehehehe, line 192: Why did you change from const to let? Is there a reason for that?

packages/components/modal/src/drawer-content.tsx
```diff
@@ -46,7 +48,7 @@ export const DrawerContent = motionForwardRef<DrawerContentProps, "div">(
     },
     ref,
   ) => {
-    const { describedbyId, duration, isOpen, labelledbyId, onClose } =
+    let { describedbyId, duration, isOpen, labelledbyId, open, onClose } =
```
kaehehehe, line 51: Why did you change from const to let? Is there a reason for that?

kaehehehe (commented):
"I think it's best to create a new branch from 2.0 and commit. Do we need a changeset for this update?

packages/components/tooltip/src/tooltip.tsx
nathanlao: Oops, sorry, I was testing around locally and forgot to change it back, I have changed it back :)

packages/components/modal/src/drawer-content.tsx
nathanlao: Good catch! I have changed it back

nathanlao:
> I think it's best to create a new branch from 2.0 and commit. Do we need a changeset for this update?

Sorry! I think @hirotomoyamada mentioned that for `@deprecated` issues, we should branch out from main.

hirotomoyamada:
@nathanlao 

Hi, there are a lot of changes in the source code and it is very difficult to review.

Can you close this PR and divide the PR into smaller pieces, per package?

PR #3527, opened 2024-11-17
`Tabs`: remove `is` prefix in props

nathanlao:
Closes #3498 

## Description

Removing all `is` prefix from `is*`

## Current behavior (updates)

Before: isFitted, isLazy, isManual

## New behavior

After: fitted, lazy, manual

## Is this a breaking change (Yes/No):

No

PR #3538, opened 2024-11-17
`segmented-control`: remove is prefix in props

### LINKED ISSUE #3494: Rename `props`: `is*` eliminates all is prefix in the `SegmentedControl` package
### Description

Change and unify the names of props.
Removing all is prefix from is*.

For example, `isOpen` becomes `open` and `isDisabled` becomes `disabled`.

### Proposed Solution or API

props to be changed are marked `@deprecated` in v1.

### Alternatives

_No response_

None

### Additional Information

_No response_

nathanlao:
Closes #3494

## Description

Removing all is prefix from is*.

## Current behavior (updates)

Before: isDisabled, isReadOnly

## New behavior

After: disabled, readOnly

No


packages/components/segmented-control/src/segmented-control.tsx
```diff
@@ -412,6 +435,9 @@ export const SegmentedControlButton = forwardRef<
     },
     ref,
   ) => {
+    disabled ??= isDisabled
```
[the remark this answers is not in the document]
nathanlao: @odendayoko 
Thank you for pointing it out! I have renamed the variales as well.

However, when I applied the changes you suggested for `<SegmentedControlButton />`, the following test case consistently failed. I also tested the `isDisabled` prop on the main branch, and the test case still failed. Could there be an issue with the test case itself?

```js
const { index, register } = useDescendant({
    disabled: disabled || readOnly,
})
```

https://github.com/yamada-ui/yamada-ui/blob/7201179a2bfd172e635998f8bc2e84f2fd06d78c/packages/components/segmented-control/tests/segmented-control.test.tsx#L112-L128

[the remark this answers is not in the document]
odendayoko: @nathanlao 
Sorry... It certainly fails the test😢

When I wrapped line 128 with waitFor like this: `waitFor(() => expect(thirdButton).toHaveAttribute("data-focus"))`, the test passed.
I believe that elements with the `disabled` attribute cannot receive focus even when using `fireEvent.focus()`.
Therefore, a test case that more closely resembles user interaction might look like this:
```ts
test("focus moves to the next element when the focused element is disabled", () => {
    render(
      <SegmentedControl>
        <SegmentedControlButton value="one">One</SegmentedControlButton>
        <SegmentedControlButton disabled value="two">
          Two
        </SegmentedControlButton>
        <SegmentedControlButton value="three">Three</SegmentedControlButton>
      </SegmentedControl>,
    )

    const firstButton = screen.getByRole("radio", { name: /one/i })
    const thirdButton = screen.getByRole("radio", { name: /three/i })

    fireEvent.focus(firstButton)
    fireEvent.keyDown(firstButton, { key: "ArrowRight" })

    waitFor(() => expect(thirdButton).toHaveAttribute("data-focus"))
  })
```

odendayoko:
@nathanlao @hirotomoyamada 
Renaming from `useIsMounted` to `useMounted` is causing the quality test to fail 😢.

In the context of the task "Removing all is prefix from is*", should we also rename hooks or internal implementations such as `isFocusVisible`, `isMounted`, or `isSelected` that are used only internally?

I initially thought that both public props and internal variable/function names should have the `is` prefix removed. Is it correct that for this task, we only need to focus on public-facing props and don't need to rename internal usages like these?

```ts
// before
 const [isFocusVisible, setIsFocusVisible] = useState<boolean>(false)

// after
const [focusVisible, setFocusVisible] = useState<boolean>(false)
```

hirotomoyamada:
@odendayoko @nathanlao 

Hi, In the future, we would like to remove `is` from all variables, component names, and hook names.

packages/components/segmented-control/src/segmented-control.tsx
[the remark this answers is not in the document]
nathanlao: Thank you so much! I applied your suggested changes!

nathanlao:
@odendayoko 
I have renamed the related hooks and all variables in the `segmented-control` package based on Mr.Yamada's requirement. Since `useMounted` is used in `skeleton` package as well, I have also renamed it otherwise it fails on the quality test. 

Please re-review! Thank you!

packages/components/segmented-control/src/segmented-control.tsx
```diff
@@ -421,7 +447,7 @@ export const SegmentedControlButton = forwardRef<
     } = useSegmentedControl()
 
     const { index, register } = useDescendant({
-      disabled: isDisabled || isReadOnly,
+      disabled: disabled || readOnly,
     })
```
odendayoko, line 451: @nathanlao @hirotomoyamada 
I think @nathanlao's modifications were correct, but after looking at the [Tab PR](https://github.com/yamada-ui/yamada-ui/pull/3527/files#diff-797ad61210a8f774540381a3d271654ebf269e80aaad835b3afe97d6af2b4957), I'm wondering if I might have been wrong in my review.

In the `Tab`modification, `isDisabled` and `isFocusable` are not destructured, and only the renamed `disabled` and `focusable` are passed as arguments to `useTabDescendant`.
I was concerned that when `isDisabled` or `isFocusable` props are passed to `Tab`, these options wouldn't be properly registered as `disabled` options in the `register` argument. However, this PR has been approved.

In light of this, the correct revision might be the following:
```ts
export const SegmentedControlButton = forwardRef<
  SegmentedControlButtonProps,
  "input"
>(
  (
    {
      className,
      children,
      disabled,
      readOnly,
      value,
      motionProps,
      onChange,
      ...rest
    },
    ref,
  ) => {
    const [, mounted] = useMounted({ rerender: true })
    const {
      styles,
      value: selectedValue,
      getInputProps,
      getLabelProps,
    } = useSegmentedControl()

    const { index, register } = useDescendant({
      disabled: disabled || readOnly,
    })
```
I considered it but couldn't figure it out, so please let me know which is correct🙇‍♀️

odendayoko (changes requested):

packages/components/segmented-control/tests/segmented-control.test.tsx
```diff
@@ -113,7 +113,7 @@ describe("<SegmentedControl />", () => {
     render(
       <SegmentedControl>
         <SegmentedControlButton value="one">One</SegmentedControlButton>
-        <SegmentedControlButton disabled value="two">
+        <SegmentedControlButton isDisabled value="two">
```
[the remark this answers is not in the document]
nathanlao: Oops! I have changed it back

packages/components/segmented-control/src/segmented-control.tsx
nathanlao: @hirotomoyamada 
I would also want confirmation from Mr. Yamada as well.

Shunii85: @nathanlao @odendayoko 
I think @nathanlao 's modifiers are correct and `Tab`'s modifiers are not. With `isDisabled` props, I could not confirm the behavior of skipping disabled tabs and moving on to the next tab in the documentation (with `disabled` props ok)

The Tab component seemed to be implemented correctly on the face of it, but I think that is due to use-clickable.

Shunii85: I've created an Issue.(#3797 )
If I'm wrong, tell me please!

hirotomoyamada: @nathanlao 

You fit the bill. Congratulations.

packages/components/segmented-control/tests/segmented-control.test.tsx
```diff
@@ -58,7 +58,7 @@ describe("<SegmentedControl />", () => {
 
   test("should disable segmented control", () => {
     render(
-      <SegmentedControl data-testid="SegmentedControl" isDisabled>
+      <SegmentedControl data-testid="SegmentedControl" disabled>
```
odendayoko, line 61: @nathanlao 
We should ensure that the props with `@depricated` are also working, so please do not change the props in the test code.

packages/components/segmented-control/tests/segmented-control.test.tsx
```diff
@@ -99,7 +99,7 @@ describe("<SegmentedControl />", () => {
 
   test("SegmentedControl with non-SegmentedControlButton children renders correctly", async () => {
     const { container } = render(
-      <SegmentedControl data-testid="SegmentedControl" isDisabled>
+      <SegmentedControl data-testid="SegmentedControl" disabled>
```
odendayoko, line 102: @nathanlao 
We should ensure that the props with `@depricated` are also working, so please do not change the props in the test code.

odendayoko (changes requested):
@nathanlao 
Sorry for the delay in re-reviewing.🙇‍♀️

Due to a number of bugs from PR after the `@depricated` was given, it is now the policy not to modify the test code and Storybook.

Can you please revert back to the original?

packages/components/segmented-control/tests/segmented-control.test.tsx
nathanlao: Thank you! I will keep that in mind!

nathanlao:
@odendayoko 

Thank you!!

I have reverted the changes related to Storybook and tests. Additionally, it seems the Yamada team prefers not to change the hook name, so I reverted the updates to the `useIsMounted` hook as well. Please let me know if this is unnecessary.

odendayoko:
@nathanlao 
Thanks!

>Additionally, it seems the Yamada team prefers not to change the hook name, so I reverted the updates to the useIsMounted hook as well. 

I was aware of removing `is` from hooks names as well.

However, please check with @hirotomoyamada as the policy may have changed as to whether the naming of hooks should be changed.

nathanlao:
Hi @hirotomoyamada 

> However, please check with @hirotomoyamada as the policy may have changed as to whether the naming of hooks should be changed.

As @odendayoko mentioned, I reverted the changes of `useIsMounted` hook here [`83323d7` (#3538)](https://github.com/yamada-ui/yamada-ui/pull/3538/commits/83323d7190040b5b1d18b1ea04e1e74beda01fde), but I would like to double confirm with you whether we still need to change the hook name?

hirotomoyamada:
@nathanlao 

For now, I would like to see `useIsMounted` renamed to `useMounted` as a hook and not to be taken as a function.

nathanlao:
@odendayoko @hirotomoyamada 

Storybook changes have been reverted and `useIsMounted` has renamed as `useMounted`, test changes have been reverted except for https://github.com/yamada-ui/yamada-ui/pull/3538/files#diff-29557407e6d95d0c95ece569d35d6650d1f0da121dfbb8da4047d272cabd916a , added `waitFor` to handle focus in a disabled element.

Thank you, please check!

odendayoko (approved):
@nathanlao 
LGTM🧡
Sorry for the poor review process🙇‍♀️Thanks for sticking with it!

packages/utils/tests/react.test.tsx
hirotomoyamada, line 1: Do not modify this file.

packages/utils/src/react.tsx
hirotomoyamada, line 1: Do not remove `useIsMounted`.

Please add `useIsMounted`, `UseIsMountedProps` and `UseIsMountedReturn`.

```
/**
 * @deprecated Use `UseMountedProps` instead.
 */
export type UseIsMountedProps = UseMountedProps

/**
 * @deprecated Use `useMounted` instead.
 */
export const useIsMounted = useMounted

/**
 * @deprecated Use `UseMountedReturn` instead.
 */
export type UseIsMountedReturn. = UseMountedReturn
```

packages/utils/tests/react.test.tsx
nathanlao: Thank you! Fixed! Please re-review. @hirotomoyamada

packages/utils/src/react.tsx
nathanlao: Thank you! Fixed! Please re-review. @hirotomoyamada

hirotomoyamada:
@nathanlao 

There are still four unresolved reviews.

nathanlao:
> @nathanlao
> 
> There are still four unresolved reviews.

@hirotomoyamada 
Sorry I don't see any unresolved comments, am I missing anything 😓

hirotomoyamada:
@nathanlao 

- Please change `disabledState` to `trulyDisabled`.
- Please change `readOnlyState` to `trulyReadOnly`.

These are!

nathanlao:
> @nathanlao
> 
> * Please change `disabledState` to `trulyDisabled`.
> * Please change `readOnlyState` to `trulyReadOnly`.
> 
> These are!

@hirotomoyamada 
Oops. I don’t see them at all.
Thanks! Got them fixed!

hirotomoyamada (approved):

PR #3589, opened 2024-11-23
Updated props and Popper integration

hirotomoyamada:
## Description

Updated props and Popper integration.

No


zonoryo17 (approved):
LGTM

PR #3591, opened 2024-11-24
Added generic type to component exports

hirotomoyamada:
## Description

Added generic type to component exports.

No


odendayoko (approved):
@hirotomoyamada 
I have confirmed that the template is generated as expected.
LGTM🧡

PR #3635, opened 2024-11-26
Reordered story components

hirotomoyamada:
## Description

Reordered story components.

No


odendayoko:
@hirotomoyamada 
Why did you need to change the order?Is it to align the order with the document site?

hirotomoyamada:
@odendayoko 

> Why did you need to change the order?Is it to align the order with the document site?

Basically, `size`, `variant` and `colorScheme` are written first. So this is the wrong order.

odendayoko (approved):
@hirotomoyamada 
> Basically, size, variant and colorScheme are written first. So this is the wrong order.

Thanks!! LGTM🧡

PR #3686, opened 2024-11-30
feat: rename is prefix in variables for select

chida09:
Closes https://github.com/yamada-ui/yamada-ui/issues/3619

### DIFF
--- a/packages/components/select/src/multi-select.tsx
+++ b/packages/components/select/src/multi-select.tsx
@@ -31,6 +31,12 @@ import {
 
 interface MultiSelectOptions {
   /**
+   * If `true`, display the multi select clear icon.
+   *
+   * @default true
+   */
+  clearable?: boolean
+  /**
    * The custom display value to use.
    */
   component?: FC<{
@@ -59,6 +65,8 @@ interface MultiSelectOptions {
    * If `true`, display the multi select clear icon.
    *
    * @default true
+   *
+   * @deprecated Use `clearable` instead.
    */
   isClearable?: boolean
   /**
@@ -97,7 +105,7 @@ interface MultiSelectOptions {
 
 export interface MultiSelectProps
   extends ThemeProps<"MultiSelect">,
-    Omit<UseSelectProps<string[]>, "isEmpty" | "placeholderInOptions">,
+    Omit<UseSelectProps<string[]>, "placeholderInOptions">,
     MultiSelectOptions {}
 
 /**
@@ -109,6 +117,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
   const [styles, mergedProps] = useComponentMultiStyle("MultiSelect", props)
   let {
     className,
+    clearable,
     closeOnSelect = false,
     color,
     component,
@@ -133,7 +142,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
   const {
     children,
     descendants,
-    isEmpty,
+    empty,
     placeholder,
     value,
     formControlProps,
@@ -150,6 +159,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
     placeholderInOptions: false,
   })
 
+  clearable ??= isClearable
   h ??= height
   minH ??= minHeight
 
@@ -183,7 +193,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
                 />
               </PopoverTrigger>
 
-              {isClearable && value.length ? (
+              {clearable && value.length ? (
                 <SelectClearIcon
                   {...clearIconProps}
                   onClick={handlerAll(clearIconProps?.onClick, onClear)}
@@ -194,7 +204,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
               )}
             </ui.div>
 
-            {!isEmpty ? (
+            {!empty ? (
               <Portal {...portalProps}>
                 <SelectList
                   footer={runIfFunc(footer, { value, onClose })}
--- a/packages/components/select/src/use-select-option.ts
+++ b/packages/components/select/src/use-select-option.ts
@@ -11,7 +11,7 @@ import {
   isUndefined,
   mergeRefs,
 } from "@yamada-ui/utils"
-import { useCallback, useId, useRef } from "react"
+import { useCallback, useId, useMemo, useRef } from "react"
 import { useSelectContext, useSelectDescendant } from "./use-select"
 
 const isTargetOption = (target: EventTarget | null): boolean =>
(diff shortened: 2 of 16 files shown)


packages/components/select/src/multi-select.tsx
```diff
@@ -109,6 +117,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
   const [styles, mergedProps] = useComponentMultiStyle("MultiSelect", props)
   let {
     className,
+    clearable = true,
```
kaehehehe, line 120:
```suggestion
    clearable
```

packages/components/select/src/multi-select.tsx
```diff
@@ -150,6 +159,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
     placeholderInOptions: false,
   })
 
+  const clearableProps = clearable || isClearable
```
kaehehehe, line 162: clearable ??= isClearable

packages/components/select/src/multi-select.tsx
```diff
@@ -183,7 +193,7 @@ export const MultiSelect = forwardRef<MultiSelectProps, "div">((props, ref) => {
                 />
               </PopoverTrigger>
 
-              {isClearable && value.length ? (
+              {clearableProps && value.length ? (
```
kaehehehe, line 196:
```suggestion
              {clearable && value.length ? (
```

packages/components/select/src/use-select-option.ts
```diff
@@ -195,8 +209,8 @@ export const useSelectOption = (props: UseSelectOptionProps) => {
   return {
     children,
     customIcon,
-    isFocused,
-    isSelected,
+    isFocused: focused,
+    isSelected: selected,
```
kaehehehe, line 213:
```suggestion
/*
* @deprecated Use `focused` instead.
*/
    isFocused: focused,
 /*
* @deprecated Use `selected` instead.
*/
    isSelected: selected,
    selected,
    focused
```

packages/components/select/src/use-select.tsx
```diff
@@ -731,8 +723,8 @@ export const useSelect = <T extends MaybeValue = string>(
     descendants,
     fieldRef,
     focusedIndex,
-    isEmpty,
-    isOpen,
+    isEmpty: empty,
+    isOpen: open,
```
kaehehehe, line 727:
```suggestion
    isEmpty: empty,
    isOpen: open,
    empty,
    open
```

kaehehehe (commented):
I think it would be better to align the code style with the other components!

packages/components/select/src/multi-select.tsx
chida09: You suggest removing the default value of `clearable`, but I set this to match `isClearable = true`. 
Should I remove the default value?

https://github.com/yamada-ui/yamada-ui/pull/3686/files#diff-4fd53e910fe635301af83fbb115c3187ab09e3f8ba4563622610118bec6e38ffR129

chida09: @kaehehehe 
You suggest removing the default value of `clearable`, but I set this to match `isClearable = true`. 
Should I remove the default value?

https://github.com/yamada-ui/yamada-ui/blob/main/packages/components/select/src/multi-select.tsx#L120

kaehehehe: @chida09 
It could be beneficial to set `clearable = true` when removing deprecated props in `version 2.0`.

chida09: @kaehehehe 
Since removing default values would change the behavior, wouldn't it be better not to remove them starting from the deprecation stage?

kaehehehe: I think it would be better to set clearable = true when working with version 2.0! @chida09

chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/2689aac87c5703595a36189df9bcd8feb1d04cb7?w=1#diff-4fd53e910fe635301af83fbb115c3187ab09e3f8ba4563622610118bec6e38ffL120

packages/components/select/src/multi-select.tsx
chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/2689aac87c5703595a36189df9bcd8feb1d04cb7?w=1#diff-4fd53e910fe635301af83fbb115c3187ab09e3f8ba4563622610118bec6e38ffL162

packages/components/select/src/multi-select.tsx
chida09: Fxied.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/2689aac87c5703595a36189df9bcd8feb1d04cb7?w=1#diff-4fd53e910fe635301af83fbb115c3187ab09e3f8ba4563622610118bec6e38ffL196

packages/components/select/src/use-select-option.ts
chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/2689aac87c5703595a36189df9bcd8feb1d04cb7?w=1#diff-40f7556e8f6b8aabc917fd5ba1375a29d5c9ced1f8277d1cbaf9eb7ef85f4919R212

packages/components/select/src/use-select.tsx
chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/2689aac87c5703595a36189df9bcd8feb1d04cb7?w=1#diff-cc2fc4aa6b03bb10b2e0bf5708b590b15cb7284ceef3226156f72265fe9f3be2R724

kaehehehe (approved):
LGTM

packages/components/select/src/use-select.tsx
hirotomoyamada, line 1: The `useSelect` is a hook used internally. So, don't bother to return both `isEmpty` and `empty`.

Other files are similar, so please modify them in the same way.

Also, there is no need to rename the returned variable.

kaehehehe: @hirotomoyamada 
Does this mean that there is no need for modification in cases like the one below? However, the '`is`' prefix has already been removed in another component, and it has been merged.


```tsx
  return {
    children,
    customIcon,
    focused,
    /*
     * @deprecated Use `focused` instead.
     */
    isFocused: focused,
    /*
     * @deprecated Use `selected` instead.
     */
    isSelected: selected,
    selected,
    getOptionProps,
  }
```

chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/3686/commits/ac86adccad3f897824457f16aabb241d51614d9f?w=1

chida09: @hirotomoyamada 
By the way, focused in useSelectOption appears to be unused. Is it necessary to return it?

https://github.com/yamada-ui/yamada-ui/blob/feat/rename-variables-in-select/packages/components/select/src/use-select-option.ts#L212

https://github.com/yamada-ui/yamada-ui/blob/feat/rename-variables-in-select/packages/components/select/src/option.tsx#L21"

hirotomoyamada (dismissed):

hirotomoyamada (approved):

PR #3807, opened 2024-12-02
feat(use-disclosure): remove is prefix from props

### LINKED ISSUE #3804: Rename props: is* eliminates all is prefix in the `useDisclosure`
### Description

Change and unify the names of props.
Removing all is prefix from is*

For example, `isOpen` becomes `open` and `isDisabled` becomes `disabled`.

### Proposed Solution or API

props to be changed are marked `@deprecated` in v1

### Alternatives

_No response_

None

### Additional Information

_No response_

kaehehehe:
Closes #3804

## Description

Removing all is prefix from is*.

No

### DIFF
--- a/packages/hooks/use-disclosure/src/index.ts
+++ b/packages/hooks/use-disclosure/src/index.ts
@@ -6,7 +6,11 @@ export interface UseDisclosureProps<
   M extends (...args: any[]) => Promise<void> | void = () => void,
 > {
   defaultIsOpen?: boolean
+  /**
+   * @deprecated Use `open` instead.
+   */
   isOpen?: boolean
+  open?: boolean
   timing?: "after" | "before"
   onClose?: M
   onOpen?: Y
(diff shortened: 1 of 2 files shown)


chida09 (approved):
🙆‍♂️

hirotomoyamada (approved):

packages/hooks/use-disclosure/src/index.ts
```diff
@@ -92,12 +107,16 @@ export const usePromiseDisclosure = <
...
+  let {
     isOpen,
+    open,
     onClose: onInternalClose,
     onOpen: onInternalOpen,
     onToggle,
   } = useDisclosure<Y, M>(rest)
+
+  open = isOpen
+
```
nathanlao, line 119: @kaehehehe 
Sorry! Just checking around. Since the PR has merged, this might be random, but may I ask if this is change was intentional? Shouldn we use `??=` operator instead?

kaehehehe: @nathanlao 
The `isOpen` can be `undefined`, so I can't use the `??=` operator. Therefore, I always assign the value of `isOpen` to the `open` variable.

PR #4182, opened 2024-12-21
fix: replaced icon in `Autocomplete`

### LINKED ISSUE #3534: Replace `AutocompleteIcon` with `LucideIcon`
### Description

Replace `AutocompleteIcon` with `LucideIcon`

Replace the icons from `@yamada-ui/icon` with those from `@yamada-ui/lucide`.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

illionillion:
Closes #3534  

## Description

Replaced by `LucideIcon`

image image image

### DIFF
--- a/packages/react/src/components/autocomplete/autocomplete-option.tsx
+++ b/packages/react/src/components/autocomplete/autocomplete-option.tsx
@@ -3,6 +3,7 @@ import type { CSSUIObject, FC } from "../../core"
 import type { UseAutocompleteOptionProps } from "./use-autocomplete-option"
 import { ui } from "../../core"
 import { cx } from "../../utils"
+import { CheckIcon } from "../icon/icons"
 import { useAutocompleteContext } from "./autocomplete-context"
 import { AutocompleteItemIcon } from "./autocomplete-icon"
 import { useAutocompleteOption } from "./use-autocomplete-option"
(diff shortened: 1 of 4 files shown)


packages/react/src/components/autocomplete/autocomplete-option.tsx
```diff
@@ -3,6 +3,7 @@ import type { CSSUIObject, FC } from "../../core"
 import type { UseAutocompleteOptionProps } from "./use-autocomplete-option"
 import { ui } from "../../core"
 import { cx } from "../../utils"
+import { CheckIcon } from "../icon/icons"
```
108yen, line 6: Shouldn't we first remove the icons that are deprecated since we will have to change the import path again?

illionillion: If you jump the code, you'll see that this is a Lucide icon, so there's no need to delete it.

108yen: Shouldn't the writing style be consistent?

illionillion: How is it?

```diff
- import { CheckIcon } from "../icon/icons"
+ import { CheckIcon } from "../icon/"
```

108yen: That is good.
However, the depercated icon in `icon/icons.tsx` takes priority, so you must remove it before making this change.

illionillion: I understand.
But if we remove that deprecated icon, we will have to deal with everything that uses it, so I think we should just merge it as it is now and remove the deprecated one after all the replacements are done and then unify the paths.

108yen: @illionillion 
OK.
Don't forget to change the import path.

108yen (approved):

hirotomoyamada (approved):

PR #4196, opened 2024-12-24
refactor: move styles to theme for avatar

### LINKED ISSUE #3877: Make components headless for `Avatar` component
### Description

- Move static styles from component to theme.
- Remove unnecessary fallback() calls.

### Proposed Solution or API
*No response*

### Alternatives
*No response*

Yes

### Additional Information
*No response*

chida09:
Closes #3877

## Description


packages/react/src/components/avatar/avatar.tsx
```diff
@@ -191,7 +173,11 @@ const AvatarImage: FC<AvatarImageProps> = ({
       loading={loading}
       referrerPolicy={referrerPolicy}
       rounded={rounded}
-      __css={css}
+      __css={{
+        height: "100%",
+        objectFit: "cover",
+        width: "100%",
+      }}
     />
```
odendayoko, line 181: @chida09 
Is it necessary to define the object in `packages/react/src/theme/components/avatar.ts` as well as others?
Is it ok to still have CSS defined in component?

odendayoko (changes requested):
@chida09 
I have one question.
I did not understand the purpose of the related Issue and did not know if it needed to be corrected.Could you please confirm this for me?

packages/react/src/components/avatar/avatar.tsx
chida09: @hirotomoyamada @odendayoko
When migrating styles to theme, how should we handle cases where the styles variable isn't being used? I see two potential options:

1. Create new properties in the theme like styles.image (though this would currently cause type errors)
2. Leave these styles in components without migrating them to theme in this PR

Which approach would you recommend for this migration?

hirotomoyamada: @chida09 @odendayoko 

No need to create a new key. If you don't want to port it to a theme and just remove the `css` variable and move it into the file, it is not necessary.

odendayoko (approved):
@chida09 
Thanks!! LGTM🧡

@hirotomoyamada 
>No need to create a new key. If you don't want to port it to a theme and just remove the css variable and move it into the file, it is not necessary.

I'll keep that in mind👍

hirotomoyamada (approved):

PR #4206, opened 2024-12-24
`calendar`: remove depracated props

### LINKED ISSUE #3856: Remove deprecated props in `calendar` package
### Description

### Proposed Solution or API

Example:
1. the following props should be removed:
```tsx
  /**
   * If `true`, any expanded accordion item can be collapsed again.
   *
   * @default false
   *
   * @deprecated Use `toggle` instead.
   */
  isToggle?: boolean
```
2. The internal logic should be removed
```tsx
 toggle ??= isToggle
```

### Alternatives

_No response_

None

### Additional Information

_No response_

nathanlao:
Closes #3856 

## Description

This PR is to remove the deprecated props in `calendar` and correct the props in storybook examples

No


packages/react/src/components/calendar/use-calendar-picker.ts
```diff
@@ -171,8 +170,7 @@ export const useCalendarPicker = <T extends UseCalendarProps<any>>(
     hiddenOutsideDays,
     holidays,
     inputFormat = "YYYY/MM/DD",
-    isLazy,
-    isOpen,
+    lazy,
     lazyBehavior,
     locale = themeLocale ?? defaultLocale,
     matchWidth,
```
nathanlao, line 176: When I remove the deprecated props `isLazy` (from popover package) in the hook, I realized that the `lazy` prop is not passed in and used, so I added back here.

hirotomoyamada (approved):

PR #4214, opened 2024-12-25
refactor: move styles to theme for form control

chida09:
Closes https://github.com/yamada-ui/yamada-ui/issues/3896

## Description


chida09:
I close it because it was not my charge.

PR #4216, opened 2024-12-26
`checkbox`: remove deprecated props

### LINKED ISSUE #3859: Remove deprecated props in `checkbox` package
### Description

### Proposed Solution or API

Example:
1. the following props should be removed:
```tsx
  /**
   * If `true`, any expanded accordion item can be collapsed again.
   *
   * @default false
   *
   * @deprecated Use `toggle` instead.
   */
  isToggle?: boolean
```
2. The internal logic should be removed
```tsx
 toggle ??= isToggle
```

### Alternatives

_No response_

None

### Additional Information

_No response_

nathanlao:
Closes #3859 

## Description

Removed `defaultIsChecked`, `isChecked`, `isIndeterminate`, `isNative`, 

Removed `isDisabled`, `isInvalid`, `isReadOnly`, `isRequired` from `form-control` package

No


packages/react/src/components/checkbox/use-checkbox-group.ts
```diff
@@ -87,17 +76,15 @@ export const useCheckboxGroup = <
...
-    { checked?: boolean; isChecked?: boolean; value?: Y }
+    { checked?: boolean; value?: Y }
   > = useCallback(
     (props = {}, ref = null) => ({
       ...props,
       ref,
-      [isNative ? "checked" : "isChecked"]: value.some(
-        (val) => String(props.value) === String(val),
-      ),
+      checked: value.some((val) => String(props.value) === String(val)),
```
nathanlao, line 84: Since we are removing `isNative` prop here, I am assuming `checked` is used only by default

Shunii85: I think it is true, too.

Shunii85 (approved):

hirotomoyamada (approved):

PR #4220, opened 2024-12-26
refactor: move styles to theme for form list

chida09:
Closes # 

## Description

### DIFF
--- a/packages/react/src/components/list/list.tsx
+++ b/packages/react/src/components/list/list.tsx
@@ -55,22 +55,13 @@ export const List = forwardRef<ListProps, "ul">((props, ref) => {
   const {
     className,
     children,
-    gap = "fallback(2, 0.5rem)",
+    gap = 2,
     stylePosition: listStylePosition,
     styleType: listStyleType = "none",
     ...rest
   } = omitThemeProps(mergedProps)
 
   const validChildren = getValidChildren(children)
-
-  const css: CSSUIObject = {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/list/list.tsx
```diff
@@ -55,23 +55,14 @@ export const List = forwardRef<ListProps, "ul">((props, ref) => {
   const {
     className,
     children,
-    gap = "fallback(2, 0.5rem)",
+    gap = 2,
     stylePosition: listStylePosition,
```
odendayoko, line 59: @chida09 
Why did you modify this?I thought there was no need to correct it.

odendayoko (changes requested):

packages/react/src/components/list/list.tsx
chida09: @odendayoko 
I apologize for not explaining this clearly. 

There was a comment from Yamada-san on Discord requesting changes to the fallback implementation, so I was working on addressing that together with them!

https://discord.com/channels/1114670051900010526/1318207809707577408/1321202786213433417

odendayoko (approved):
@chida09 
> There was a comment from Yamada on Discord requesting changes to the fallback implementation, so I was working on addressing that together with them!

Understood!
Thanks! LGTM🧡

hirotomoyamada (approved):

PR #4265, opened 2024-12-30
`image`: remove `size` prop and rename internal variable

### LINKED ISSUE #3500: Remove `size` prop in `@yamada-ui/image`
### Description

Remove the `size` prop from the `Image` component.

### LINKED ISSUE #3870: Remove deprecated props in `image` package
### Description

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

- Variable names are identified and require changes in this package
   - `packages/components/image/src/image.tsx`

nathanlao:
Closes #3870 
Closes #3500 

## Description

Removed `size` prop
Renamed invernal variable

No


packages/react/src/components/image/image.tsx
```diff
@@ -60,14 +53,11 @@ export const Image: FC<ImageProps> = (props) => {
     ...rest
   } = omitThemeProps(mergedProps)
   const ignoreFallback = loading != null || _ignoreFallback || !fallback
-  const css = useMemo(
-    () => ({ ...styles, boxSize, objectFit }),
-    [styles, boxSize, objectFit],
```
odendayoko, line 65: @nathanlao @hirotomoyamada 
Thank you for the changes!

It works fine, but I have a question.
Rather than passing `boxSize` directly as `props`, I feel that passing it integrated into `css` as before allows for centralized style management and is more maintainable.

I would like to hear your opinions.

odendayoko (dismissed):

packages/react/src/components/image/image.tsx
hirotomoyamada: @odendayoko 

It is no longer necessary to insert it into the `css` object, so you can leave it as is.

hirotomoyamada (approved):

PR #4314, opened 2025-01-13
Update reset css

### LINKED ISSUE #3821: Update reset css

odendayoko:
Closes #3821 

## Description
Update to modern reset css

No

### Base Reset CSS
- This reset CSS is based on the following repository:
  - [**modern-normalize**](https://github.com/sindresorhus/modern-normalize)

### Comparison with Other Reset CSS Libraries
- During the implementation, we compared the following libraries:
  - [**reset.css**](https://github.com/mayank99/reset.css)
  - [**Tailwind's `preflight.css`**](https://github.com/tailwindlabs/tailwindcss/blob/next/packages/tailwindcss/preflight.css)
  - [**Chakra UI's `preflight.ts`**](https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/styled-system/preflight.ts)

### Layout Issues After Changes
- `cursor: "pointer"` was only included in `reset.css`, and after removing it, the `Tab` `cursor` changed to `default`.
```diff
- "button, [type='button'], [type='reset'], [type='submit']": {
-    appearance: "none",
-    cursor: "pointer",
-    WebkitAppearance: "none",
-  },
```
- `display: "block"` was applied in all libraries except `modern-normalize-css`. After adopting it, the “Hokuto no Ken” logo image in `Bleed` could no longer be centered using `text-align`.
```diff js
+  "img, svg, video, canvas, audio, iframe, embed, object": {
+    display: "block",
+  },
```
- Since it was not adopted by other libraries, `th` styles were removed. However, this caused `Table` headers to be `center-aligned`.
```diff
-  th: {
-    fontWeight: "bold",
-    textAlign: "left",
-  },
```


odendayoko:
@taiseiiiii  @taroj1205 
I received feedback at the 1/14 meeting and are in the process of revising it.
Please do not review yet🙇‍♀️

hirotomoyamada:
@odendayoko 

Hi, there is nothing to reset the `mark` style.

odendayoko:
> @odendayoko
> 
> Hi, there is nothing to reset the `mark` style.

The `backgroundColor` and `color` were specified in the browser standard.
Which do you think is the best way to reset `mark` style?
```js
// option1
 mark: {
    backgroundColor: "transparent",
    color: "inherit",

// option2
 mark: {
    backgroundColor: "inherit",
    color: "inherit",
```

hirotomoyamada:
@odendayoko 

I think `option2` is the way to go.

taroj1205:
Hi @odendayoko, is this PR ready for review yet?

packages/react/src/theme/styles/reset-style.ts
```diff
@@ -1,233 +1,109 @@
...
-    fontFamily: "monospace, monospace",
-    fontSize: "inherit",
-  },
-
-  dd: {
-    marginLeft: 0,
-  },
-
-  details: {
+  "img, svg, video, canvas, audio, iframe, embed, object": {
```
koralle, line 64: @odendayoko 
Should we include the `<picture>` element in this selector?

packages/react/src/theme/styles/reset-style.ts
```diff
@@ -1,233 +1,109 @@
...
     color: "inherit",
-    height: 0,
-    margin: 0,
     overflow: "visible",
   },
 
   html: {
-    lineHeight: 1.15,
+    fontFamily:
+      "system-ui, 'Segoe UI', Roboto, Helvetica, Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji'",
```
koralle, line 83: @odendayoko 
What was the reasoning behind selecting `Helvetica` rather than `Helvetica Neue`, its enhanced version?

packages/react/src/theme/styles/reset-style.ts
odendayoko: @koralle @hirotomoyamada 
Thanks!!

This setting is based on [Tailwind CSS](https://github.com/tailwindlabs/tailwindcss/blob/main/packages/tailwindcss/preflight.css#L212) and [Chakra UI](https://github.com/chakra-ui/chakra-ui/blob/main/packages/react/src/styled-system/preflight.ts#L45), and is based on the `replaced elements`.As commented on the Tailwind CSS side, 
> 1. Make replaced elements display: block by default. (https://github.com/mozdevs/cssremedy/issues/14)

On the other hand, mordern-normalize does not add this setting.It is assumed that the browser's default is used and there is no need to override it.

Therefore, if we follow mordern-normalize, we may not need this setting in the first place, but what do you think?

packages/react/src/theme/styles/reset-style.ts
odendayoko: @koralle @hirotomoyamada 
Thanks!!

>What was the reasoning behind selecting Helvetica rather than Helvetica Neue, its enhanced version?

Sorry, I did not know this.

>Since Yamada UI is a Japanese-origin UI component library with many users in Japan, I believe we should specify font-family that considers Japanese typefaces, not just Western fonts.

You are certainly right👍

This setup is based on [modern-normalize](https://github.com/sindresorhus/modern-normalize/blob/main/modern-normalize.css#L25).The purpose is as follows.
> Improve consistency of default fonts in all browsers.

Since the role of reset CSS is to reset browser-specific initial styles, it may not be necessary to specify font families.

hirotomoyamada (approved):

PR #4410, opened 2025-01-22
refactor: move styles to theme for multi select

chida09:
Closes https://github.com/yamada-ui/yamada-ui/issues/4409

## Additional Information
Due to type errors in branch v2.0 code, reviewing only the source code changes will be sufficient.

### DIFF
--- a/packages/react/src/theme/components/multi-select.ts
+++ b/packages/react/src/theme/components/multi-select.ts
@@ -24,6 +24,9 @@ export const MultiSelect: ComponentMultiStyle<"MultiSelect"> = mergeMultiStyle(
           pointerEvents: "none",
         },
       },
+      inner: {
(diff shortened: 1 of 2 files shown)


108yen (approved):

packages/react/src/theme/components/multi-select.ts
```diff
@@ -7,6 +7,9 @@ export const MultiSelect: ComponentMultiStyle<"MultiSelect"> = mergeMultiStyle(
   {
     baseStyle: {
       clearIcon: {
+        inner: {
```
hirotomoyamada, line 10: Do not include `inner` in `clearIcon`, but set it on the same level.

hirotomoyamada:
I will fix it together with another work, so I will close it.

Issue #4070, opened 2024-12-13
Internationalize `aria-label` within the `Avatar`


taiseiiiii:
### Description

Internationalize `aria-label` within the `Avatar`

### Problem Statement/Justification

None

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

None

PR #4420, opened 2025-01-24
feat(avatar): Internationalize `aria-label` in `avatar`

taiseiiiii:
Closes [#4070](https://github.com/yamada-ui/yamada-ui/issues/4070)

## Description

Internationalized `aria-label` in `avatar` component.

## Current behavior (updates)

`aria-label` is a fixed value. (In this case, `Avatar Icon`)

## New behavior

If developer define preferable locale (ex. `ja-JP`), support it.

Example:

```json
{
  "en-US": {
    "avatar": {
      "avatarIcon": "Avatar Icon"
    }
  },
  "ja-JP": {
    "avatar": {
      "avatarIcon": "アバターのアイコン"
    }
  }
}
```

No


taiseiiiii:
I've just modify according to feedback in previous meeting.
Please review.

packages/react/src/providers/i18n-provider/i18n-provider.tsx
```diff
@@ -163,46 +143,62 @@ export const I18nProvider: FC<I18nProviderProps> = ({
...
-}
+export function useI18n<Y extends IntlKey>(key?: Y) {
+  const context = useContext(I18nContext)
 
-export function isRtl(_locale: string) {
-  const locale = new Intl.Locale(_locale).maximize()
+  const translate = useCallback(
+    <M extends Path<IntlData[Y]>>(
+      path: M,
+      replaceValue?: ReplaceValue<M>,
```
Shunii85, line 186: @hirotomoyamada
For example, when I edit `en-US.json` like following, I can confirm that the inference of `replaceValue` dose'nt works correctly.
```json
{
  "avatar": {
    "Avatar Icon": "Avatar Icon {label}"
  }
```

using value like following
```ts
t("Avatar Icon", "something")
```

Do you envision using the feature this way too?

hirotomoyamada: @Shunii85 

Hi, My assumptions are as follows

```
{
    "avatar": {
      "Avatar Icon {label}": "Avatar Icon {label}"
  }
}
```

The `{}` is present in the key of the object.

However, that is unnatural, so the type definition has been modified.

https://github.com/yamada-ui/yamada-ui/pull/4420/commits/2e984f536db6d3f5f93fa878a6fc885d66e60c68

Shunii85: @hirotomoyamada LGTM!! Genius

Shunii85 (approved):

hirotomoyamada (approved):

PR #4448, opened 2025-02-05
Refactored `Highlight` component

### LINKED ISSUE #3354: Separation of UI and Logic in the `Highlight` Component
### Description

Separate logic into separate files such as `use-xxx.ts`.
In addition, modify the hooks to function on their own.

### Additional Information

Create a new branch from `v2.0` branch for development.

108yen:
Closes #3354 

## Description

Refactored `Highlight` component to new structure.


packages/react/src/components/highlight/highlight.stories.tsx
```diff
@@ -30,7 +30,7 @@ export const WithMultiQuery: Story = () => {
   )
 }
 
-export const WithIsFragment: Story = () => {
+export const IsFragment: Story = () => {
```
odendayoko, line 33: @108yen
Change `IsFragment` to `Fragment`.
It seems that `IsDisabled` has all been changed to `Disable` in v2.0, so it would be better to align with `props`.

packages/react/src/components/highlight/highlight.tsx
odendayoko, line 1: @108yen
Shouldn't `Highlight` use `createComponent`?
I didn't want to underplay `displayName` or `__ui__` but I guess I don't have a choice 🤔.

odendayoko (dismissed):

hirotomoyamada (approved):

PR #4575, opened 2025-03-06
test: add additional tests for InfiniteScrollArea component

bmthd:
Closes #


packages/react/src/components/infinite-scroll-area/infinite-scroll-area.test.tsx
bmthd, line 1: I am concerned about the lack of confirmation of InfiniteScrollTrigger that has not been exported.

hirotomoyamada (approved):

PR #4611, opened 2025-03-16
Test/enhance-text

bmthd:
Closes #


hirotomoyamada (approved):

packages/react/src/components/text/text.test.tsx
```diff
@@ -6,6 +6,23 @@ describe("<Text />", () => {
     await a11y(<Text>Text</Text>)
   })
 
+  test("sets `displayName` and `__ui__` correctly", () => {
```
hirotomoyamada, line 9: Please change "`__ui__`" to "`__styled__`".

packages/react/src/components/text/text.test.tsx
```diff
@@ -6,6 +6,23 @@ describe("<Text />", () => {
...
+    expect(Text.displayName).toBe("Text")
+    expect(Text.__ui__).toBe("Text")
```
hirotomoyamada, line 11: Please change "`__ui__`" to "`__styled__`".

bmthd: Changed that way.

hirotomoyamada: Please change the title as well.

bmthd: Fixed.

hirotomoyamada (approved):

PR #4633, opened 2025-03-28
Refactored `Switch` component

### LINKED ISSUE #3987: Separation of UI and Logic in the `Switch` Component
### Description

Separate logic into separate files such as `use-xxx.ts`.
In addition, modify the hooks to function on their own.

### Additional Information

Create a new branch from `v2.0` branch for development.
https://react-spectrum.adobe.com/react-aria/useSwitch.html#useswitch

108yen:
Closes #3987 

## Description

Refactored `Switch` component to new structure.
## Additional Information

`Label` story is not working because the `FieldLabel` component is not available without wrapping in the `FieldRoot`.


packages/react/src/components/switch/switch.stories.tsx
```diff
@@ -24,51 +24,50 @@ export const Basic: Story = () => {
...
 }
 
-export const WithLabel: Story = () => {
+export const Label: Story = () => {
   const id = useId()
 
   return (
     <HStack gap="sm">
-      <Label htmlFor={id} userSelect="none">
+      <FieldLabel htmlFor={id} userSelect="none">
```
illionillion, line 56: If you do not use `<Field.Label>` here, you will get an error.

image

```suggestion
      <Field.Root>
        <Field.Label htmlFor={id} userSelect="none">
          Please Click
        </Field.Label>
      </Field.Root>
```

illionillion (changes requested):
@108yen 

Once the Storybook error is resolved, I think everything else is fine.

hirotomoyamada (approved):

PR #4674, opened 2025-06-02
feat(storybook): update Storybook to v9

koralle:
## Description

Update Storybook to v9

No

## Additional Information

We needed to patch the latest version v4.0.2 of `storybook-dark-mode`, so we are applying a patch by running `pnpm patch storybook-dark-mode`. 
This patch will likely be removed when the latest version of is released. 

> [!NOTE]
> So, we need to watch https://github.com/hipstersmoothie/storybook-dark-mode/pull/298 .

## References

- https://storybook.js.org/docs/configure/upgrading
- https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#from-version-8x-to-900

### DIFF
--- a/packages/react/src/components/steps/steps.stories.tsx
+++ b/packages/react/src/components/steps/steps.stories.tsx
@@ -1,4 +1,4 @@
-import type { Meta, StoryFn } from "@storybook/react"
+import type { Meta, StoryFn } from "@storybook/react-vite"
 import { useMemo, useState } from "react"
 import { Steps } from "."
 import { PropsTable } from "../../../storybook/components"


eslint.config.ts
```diff
@@ -2,6 +2,7 @@ import type { TSESLint } from "@typescript-eslint/utils"
 import type { Linter } from "eslint"
 import pluginVitest from "@vitest/eslint-plugin"
 import prettierConfig from "eslint-config-prettier"
+// For more info, see https://github.com/storybookjs/eslint-plugin-storybook#configuration-flat-config-format
```
hirotomoyamada: Why is this comment necessary?

packages/react/storybook/preview.tsx
```diff
@@ -78,6 +78,8 @@ const preview: Preview = {
           </DocsContainer>
         )
       },
+
```
hirotomoyamada: Delete any unnecessary lines.

packages/react/storybook/main.ts
```diff
@@ -1,21 +1,25 @@
...
-  framework: "@storybook/react-vite",
+  framework: getAbsolutePath("@storybook/react-vite"),
   // TODO: Remove legacy-components
   stories: ["../src/!(legacy-components)/**/*.@(mdx|stories.@(tsx))"],
   typescript: { reactDocgen: false },
 }
 
 export default config
+
+function getAbsolutePath(value: string): any {
```
hirotomoyamada: Why is this function needed? It should have been removed a while ago.

eslint.config.ts
koralle: > Why is this comment necessary? 

Sorry, it is auto-generated by `pnpm dlx @storybook/upgrade`.

packages/react/storybook/preview.tsx
koralle: Deleted.

packages/react/storybook/main.ts
koralle: Sorry, it is auto-generated by pnpm dlx @storybook/upgrade.

eslint.config.ts
koralle: Deleted.

packages/react/storybook/main.ts
koralle: Deleted.

PR #4702, opened 2025-07-07
fix(multi-select): use inline display for selected values

### LINKED ISSUE #4701: `MultiSelect` custom component width error on safari

@yamada-ui/react

### Package Version

1.7.7

### Description

When I was on the Yamada-UI documentation about [`MultiSelect` custom component](https://yamada-ui.com/components/forms/multi-select#use-custom-component), and tested the multi-select on Safari, I was facing this issue:

Image

It should be like this (Chrome):

Image

### Link to Reproduction

https://yamada-ui.com/components/forms/multi-select#use-custom-component

### Steps to Reproduce

Click all three options, and the bug will show. 

### Browser

Safari

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

No

Yes

### Additional Information

I suppose that the maximum width is buggy and doesn't actually reflect the actual width of the component, hence the `MultiSelect` component tries to compress the items into `...`, but the `x` sign will still show.

jeffplays2005:
Closes #4701 

## Description

This PR fixes the Safari issue with the `MultiSelect` component where items would be truncated when the maximum width was greater than the items width. 

## Current behavior (updates)

The truncation is applied when the width is greater than the components. 

Screenshot 2025-07-07 at 6 46 32 PM

## New behavior

There is no longer a truncation when the width is greater than the components. 

image

No

### DIFF
--- a/packages/components/select/src/multi-select.tsx
+++ b/packages/components/select/src/multi-select.tsx
@@ -282,7 +282,7 @@ const MultiSelectField = forwardRef<MultiSelectFieldProps, "div">(
                   dangerouslySetInnerHTML={{
                     __html: `${value}${!isLast ? separator : ""}`,
                   }}
-                  display="inline-block"
+                  display="inline"
                   me="0.25rem"
                 />
               )


taroj1205 (dismissed):
LGTM Thanks @jeffplays2005!

hirotomoyamada (approved):
Thanks!

Issue #4707, opened 2025-07-12
Migrate v1 documentation


kaehehehe:
### Subject

components, hooks

### Components
- [ ] accordion
- [ ] airy
- [ ] aspect-ratio
- [ ] avatar
- [ ] badge
- [ ] blockquote
- [x] box
- [ ] breadcrumb
- [ ] button
- [ ] card
- [ ] carousel
- [ ] center
- [ ] circle-grogress
- [ ] client-only
- [ ] close-button
- [ ] code
- [ ] collapse
- [ ] color-swatch
- [ ] container
- [ ] data-list
- [ ] drawer
- [ ] editable
- [ ] em
- [ ] empty-state
- [ ] fade
- [ ] fade-scale
- [ ] field
- [ ] fieldset
- [ ] file-button
- [ ] file-input
- [ ] flex
- [ ] flip
- [ ] float
- [ ] focus-lock
- [ ] for
- [ ] format-byte
- [ ] format-number
- [ ] grid
- [ ] group
- [ ] heading
- [ ] highlight
- [ ] icon
- [ ] icon-button
- [ ] image
- [ ] indicator
- [ ] infinte-scroll-area
- [ ] input
- [ ] kbd
- [ ] link
- [ ] link-box
- [ ] list
- [ ] loading
- [ ] mark
- [ ] modal
- [ ] motion
- [ ] native-select
- [ ] native-table
- [ ] number-input
- [ ] pagination
- [ ] password-input
- [ ] picture
- [ ] pin-input
- [ ] popover
- [ ] portal
- [ ] progress
- [ ] reorder
- [ ] resizable
- [ ] ripple
- [ ] rotate
- [ ] scroll-area
- [ ] segmented-control
- [ ] separator
- [ ] show
- [ ] simple-grid
- [ ] skeleton
- [ ] slide
- [ ] slide-fade
- [ ] slot
- [ ] stack
- [ ] stat
- [ ] status
- [ ] steps
- [ ] switch
- [ ] tabs
- [ ] tag
- [ ] text
- [ ] textarea
- [ ] toggle
- [ ] tooltip
- [ ] visually-hidden
- [ ] wrap

### hooks
- [ ] use-animation
- [ ] use-async
- [ ] use-async-callback
- [x] use-boolean
- [ ] use-breakpoint
- [ ] use-clipboard
- [ ] use-controllable-state
- [ ] use-counter
- [ ] use-descendant
- [ ] use-disclosure
[...]

PR #4719, opened 2025-07-22
docs: add `Container` usage

odendayoko:
## Description
#4707

## Current behavior (updates)
https://yamada-ui.com/components/layouts/container

## Additional Information
- In `v2.0` `Container` is a slot component
- [Storybook](https://yamada-ui.github.io/yamada-ui/?path=/story/components-container--basic)


odendayoko:
> Please add Variant, Size, ColorScheme section.

Can I just use one example?Can I choose from any props I like?

Container_-_Yamada_UI

hirotomoyamada (approved):

PR #4739, opened 2025-08-01
Remove redundant notes from docs

hirotomoyamada:
## Description

Remove redundant notes from docs.


taroj1205 (approved):
LGTM
I won't include it when writing useFormatByte docs

PR #4756, opened 2025-08-06
Fixed `Avatar` component style

108yen:
## Description

Fixed a bug where the style of the `Avatar` component was reversed between loading and after loading.


hirotomoyamada:
@108yen There is no problem with the style being applied while it is being loaded.

PR #4806, opened 2025-08-20
Migrate `For` documentation

108yen:
Related #4707 

## Description

Migrate `For` documentation to v2.


www/contents/components/(utilities)/for.ja.mdx
```diff
@@ -5,4 +5,250 @@ storybook: components-for--basic
 source: components/for
 ---
 
+```tsx preview functional client
```
bmthd: If you mark it as a client, a message saying “If you use this code, you need to add "use client" to the top of the file.” will be displayed.
useMemo can also be used in Server Components for compatibility, so I don't think it needs to be marked as a client.
What do you think?

108yen: I am matching the format with others.
For example: https://v2.yamada-ui.com/ja/docs/components/collapse

bmthd: Since useBoolean used in Collapse is a hook that can only be used on the client side, it is worth marking it as client.
However, although useMemo is a hook, it does not cause an error on the server side and can be used.

108yen: OK. Fixed.

www/contents/components/(utilities)/for.ja.mdx
```diff
@@ -5,4 +5,250 @@ storybook: components-for--basic
...
+        </Card.Root>
+      )}
+    </For>
+  </VStack>
+)
+```
+
+### 方向を反転させる
+
+方向を反転させる場合は、`reverse`を`true`に設定します。
```
bmthd: 些細な点ですが、以下の方が適切だと思いました。
```suggestion
### 並び順を反転させる

並び順を反転させる場合は、`reverse`を`true`に設定します。
```

www/contents/components/(utilities)/for.mdx
```diff
@@ -5,4 +5,250 @@ storybook: components-for--basic
...
+        </Card.Root>
+      )}
+    </For>
+  </VStack>
+)
+```
+
+### Reverse Direction
+
+To reverse the direction, set `reverse` to `true`.
```
bmthd: It's a minor point, but I thought the following would be more appropriate.

```suggestion
### Reverse Order

To reverse the order, set `reverse` to `true`.
```

108yen:
@bmthd 
Fixed.

bmthd (dismissed):
LGTM

PR #4829, opened 2025-08-25
feat(ci): use `Dependabot` instead of `Renovate`

koralle:
## Description

We transitioned from Renovate to Dependabot because Renovate became too difficult to manage.

## Additional Information

* [Renovate Docs](https://docs.renovatebot.com/)
* [Dependabot オプション リファレンス - GitHub Docs](https://docs.github.com/ja/code-security/dependabot/working-with-dependabot/dependabot-options-reference)


hirotomoyamada:
@koralle 

Does this mean that if there are updates to dependencies in pull requests, those updates will also be included?

Pull requests should generally contain a single change, so please do not include updates to dependencies in pull requests for other tasks.

taroj1205 (approved):
LGTM
Should we customise commit message so it follows our convention?
https://docs.github.com/ja/code-security/dependabot/working-with-dependabot/dependabot-options-reference#commit-message--

PR #4997, opened 2025-09-01
Update transition properties.

hirotomoyamada:
Closes # 

## Description

Update transition properties.


packages/react/src/components/segmented-control/segmented-control.style.ts
```diff
@@ -44,7 +44,6 @@ export const segmentedControlStyle = defineComponentSlotStyle({
         rounded: "md",
         top: "anchor(--checked top)",
         transitionDuration: "moderate",
```
taroj1205, line 46: this is not related to this PR, but are we sure we can use `anchor`? I think this breaks in firefox (and potentially on Safari as well) because it's not yet supported
https://developer.mozilla.org/en-US/docs/Web/CSS/anchor#browser_compatibility
Image

taroj1205 (approved):
LGTM

packages/react/src/components/segmented-control/segmented-control.style.ts
hirotomoyamada: @108yen @taroj1205 

Indeed, Safari and others are not supported. Could you create an issue for this matter?

PR #4999, opened 2025-09-01
refactor(grid, simple-grid): use namespaces for exports

jeffplays2005:
Closes # 

## Description

I've converted the `Grid` component to use a namespace for exporting such as `Grid.Root`, and `Grid.Item`. 

## Current behavior (updates)

Currently using `Grid` and `GridItem`, `GridProps`, etc.

## New behavior

Now using `Grid.Root`, `Grid.Item` and props use `GridRootProps`, etc. 

## Additional Information

How I should re-export `Item`, would a user use `Grid.Item` or `SimpleGrid.Item`? 

I get a lint error when I try to re-export `Item` from the `Grid` component. Perhaps a user would use `SimpleGrid.Root` and then use `Grid.Item` as the children or proxy export `Grid.Item` in `SimpleGrid` but rename to `SimpleGrid.Item`?


hirotomoyamada:
Hi, thanks for the PR.

I've thought about this issue internally. 

To cut to the chase, I'd prefer to undo this change.

The reasons are: using `<flex>` for the `Flex` component while using `<grid.root>` for the `Grid` component lacks consistency, and since the `Grid` component is widely used, the impact would be significant.

I apologize for the work already done, but I'd like to close this PR.

PR #5232, opened 2025-10-07
feat(for): add `repeat` prop for repeating

### LINKED ISSUE #5152: Add a prop that repeats `For` a specified number of times
### Description

Support iterating by count with `<For times={n}>` syntax, similar to existing `<For each={array}>`.

---

Currently `<For>` only supports iterating over an array via `each` prop. There is no first-class API for repeating a block a fixed number of times. Users must fall back to manual `Array.from` or `lodash.times`, which is less declarative and inconsistent with `<For each>`.

---

### Expected

```tsx
<For times={3}>
  {(i) => <Box key={i}>{i}</Box>}
</For>
```

---

### Actual

```tsx
{Array.from({ length: 3 }, (_, i) => (
  <Box key={i}>{i}</Box>
))}
```

---

### Proposed Solution or API

```ts
interface ForProps<T> {
  each?: T[]
  times?: number
  children: (item: T | number, index: number) => React.ReactNode
}
```

* When `times` is provided, `children` receives the index (0..n-1).
* When `each` is provided, `children` receives the element and index (current behavior).

---

### Alternatives

* Continue using `Array.from` or `lodash.times` outside of JSX.
* Create a separate component such as `<Repeat times={n}>` to avoid overloading `<For>`.

---

Yes

---

### Additional Information

This would unify iteration patterns in JSX, making `<For>` a more expressive DSL for both array iteration and count-based repetition.

a

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

taroj1205:
Closes #5152

## Description

I have added `repeat` props for repeating children.

## New behavior

Adds a new `repeat` prop.

## Additional Information

React will throw a warning
```
Encountered two children with the same key, `0`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```
This is because when I copy the elements the key is duplicated.

### DIFF
--- a/packages/react/src/components/for/for.tsx
+++ b/packages/react/src/components/for/for.tsx
@@ -1,5 +1,5 @@
 import type { ReactNode } from "react"
-import { isArray } from "../../utils"
+import { isArray, isNumber } from "../../utils"
 
 export interface ForProps<Y> {
   /**
@@ -28,6 +28,13 @@ export interface ForProps<Y> {
    */
   offset?: number
   /**
+   * The number of times to repeat the component.
+   * Values less than or equal to 1 will have no effect.
+   * The repetition is applied after all other transformations (filter, sort, reverse, limit, offset).
+   * This means the entire processed array will be repeated the specified number of times.
+   */
+  repeat?: number
+  /**
    * The boolean value to determine the order of the items in the array.
    * If `true`, the items will be reversed.
    * If `sortBy` is provided, inversion is applied to the sorted array.
@@ -40,10 +47,6 @@ export interface ForProps<Y> {
    * If `reverse` is `true`, the inversion is applied to the sorted array.
    */
   sort?: (a: Y, b: Y) => number
-  /**
-   * The number of times to repeat the component.
-   */
-  times?: number
 }
 
 /**
@@ -58,9 +61,9 @@ export const For = <Y,>({
   filter,
   limit,
   offset = 0,
+  repeat,
   reverse = false,
   sort,
-  times,
 }: ForProps<Y>): ReactNode => {
   if (!each || !isArray(each) || !each.length) return fallback || null
 
@@ -74,18 +77,20 @@ export const For = <Y,>({
 
   const rendered = sliced.map(children)
 
-  if (typeof times === "number" && times > 1) {
-    const unitLength = rendered.length
-    const totalLength = unitLength * times
-    const repeated = new Array<ReactNode>(totalLength)
-    for (let i = 0; i < times; i += 1) {
-      const baseIndex = i * unitLength
-      for (let j = 0; j < unitLength; j += 1) {
-        repeated[baseIndex + j] = rendered[j]
-      }
-    }
-    return repeated
+  if (!isNumber(repeat) || repeat <= 1) {
+    return rendered
   }
 
-  return rendered
+  const unitLength = rendered.length
+  const totalLength = unitLength * repeat
+  const repeated = new Array<ReactNode>(totalLength)
+
+  for (let i = 0; i < repeat; i += 1) {
+    const baseIndex = i * unitLength
+    for (let j = 0; j < unitLength; j += 1) {
+      repeated[baseIndex + j] = rendered[j]
+    }
+  }
+
+  return repeated
 }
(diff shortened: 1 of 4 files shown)


bmthd:
How about overloading `repeat` and `each`?

Among the array methods included in `For`, aside from `reverse`, `sort` at least doesn't seem to be usable with `repeat`.
Furthermore, allowing both props to be set would represent an impossible state.
I believe it would be better to isolate `<For repeat={i} />` and `<For each={items} />`, what do you think?

Like this [TypeScript Playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAKjgQwM5wEoFNkGN4BmUEIcARFDvmQFA3AB2MWUBeWcAYtAMoAWySgBMACsTDoA3jTiy4lAG4tUWAPwAuOACMIEADY4GNAL51GzVuy7QAonj5iIEgDwAVAHxwsAD2YMh6NxQ-IJYouJSMnJUfJquANoAuiZmTCxsuBxB2GA4MI4SXr5Y-oG8AsIFkXLyWLnIMJoMAK4gWiwpNATNDPjAEAzWUG7uABRgEZpBdrgOESMAlJrYeDAAdCv4AHIQQlhdPX0DQ+OTQzl5VUuYVOubMDt7B70w-YNBI6dOqFPloeHfa73Da3R4caQ1SgwZpQQYtPR6Tq4AaoeAzPhwAC8cFGCyxnlGUWcQS89kxkniZGQZAANGQtGREsY4AB6dw0BZ0ZEMVE3erwbG4-E4qKyYnQWr88kAZmZbI5QA)

taroj1205:
> How about overloading `repeat` and `each`?
> 
> Among the array methods included in `For`, aside from `reverse`, `sort` at least doesn't seem to be usable with `repeat`. Furthermore, allowing both props to be set would represent an impossible state. I believe it would be better to isolate `<For repeat={i} />` and `<For each={items} />`, what do you think?

I think something to think about is, whether `repeat` should do:
1. [1,2,3,1,2,3,1,2,3]
2. [1,1,1,2,2,2,3,3,3]
when used with `sort`, or `reverse` etc.

taroj1205:
Furthermore, if you isolate `each` and `repeat`, what are we going to be reapeating with `repeat`?

bmthd:
My idea was to replace what was written with the conventional For loop like this:
Simply repeating children.

```tsx
<For each={Array.from({ length: 10})}>
  {( i ) => <Text key={i} >Item: { i }</Text>}
</For>
```

Certainly, it also seems natural for children to be rendered multiple times when used in conjunction.

```tsx
<For repeat={3} items={blocks}>
  {( block, i ) => <Block key={`${block.id}-${i}`} >: { i }</Text>}
</For>
```

However, I feel it's not very intuitive as users need to set keys to be unique.
If you want to do that with overloading, you can write it like this:

```tsx
<For repeat{3}>
  {(i) => (
  <For items={blocks}>
  {( block ) => <Block key={`${block.id}-${i}`} >: { i }</Text>}
  </For>
  )}
</For>
```

This allows us to recall that it is a double loop from the signature and write it intuitively.

packages/react/src/components/for/for.tsx
```diff
@@ -67,5 +72,20 @@ export const For = <Y,>({
...
+    const repeated = new Array<ReactNode>(totalLength)
+    for (let i = 0; i < repeat; i += 1) {
+      const baseIndex = i * unitLength
+      for (let j = 0; j < unitLength; j += 1) {
+        repeated[baseIndex + j] = rendered[j]
+      }
+    }
+    return repeated
+  }
+
```
bmthd, line 83: You can use isNumber to check if repeat is a number.
How about avoiding logic that leads to deep nesting and using early returns instead?

taroj1205: @bmthd I've updated to use `isNumber`, please check.

packages/react/src/components/for/for.tsx
```diff
@@ -67,5 +75,22 @@ export const For = <Y,>({
 
   if (!sliced.length) return fallback || null
 
-  return sliced.map(children)
+  const rendered = sliced.map(children)
+
+  if (!isNumber(repeat) || repeat <= 1) {
+    return rendered
+  }
```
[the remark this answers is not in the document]
bmthd: @taroj1205 
It seems like some changes are needed.

PR #5294, opened 2025-10-19
feat(native-popover): add native-popover component

### LINKED ISSUE #3845: Add `NativePopover` component
### Description

Using [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover), we would like to create a native popover component us.

Currently, [Popover](https://yamada-ui.com/components/overlay/popover) is implemented, but [NativeSelect](https://yamada-ui.com/components/forms/native-select) should be implemented if it can be implemented natively.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

taroj1205:
Closes #3845

## Description

I have added `NativePopover` component.

## New behavior

This adds a new `NativePopover` component like the `NativeSelect`.

## Additional Information

image

<h2>Greptile Overview</h2>

Updated On: 2025-10-23 01:17:49 UTC

### Greptile Summary

**This review covers only the changes made since the last review, not the entire PR.** The latest update moves DOM queries for `hasHeader` and `hasBody` from the hook's top-level into the `getContentProps` callback (lines 132-133 in `use-native-popover.tsx`). This change addresses the previous issue where queries executed on every render regardless of usage, but introduces a new performance concern: the queries now run on every invocation of `getContentProps`, even though header/body presence is typically stable after mount. The `getDocument()?.getElementById()` calls traverse the DOM each time the callback executes, which could impact performance in components that render frequently or call `getContentProps` repeatedly. While the dependency array correctly includes `getDocument`, `headerId`, and `bodyId` (line 147), the callback should be optimized to avoid redundant DOM queries when the structure hasn't changed—ideally through ref-based caching or mutation observers.

<details>
<summary>Changed Files</summary>

| Filename | Score | Overview |
|----------|-------|----------|
| packages/react/src/components/native-popover/use-native-popover.tsx | 3/5 | Moves DOM queries inside `getContentProps` callback (lines 132-133), fixing unconditional execution on every render but introducing repeated queries on each callback invocation—queries should be cached via refs or tracked with mutation observers |

</details>

### Confidence score: 3/5

- This PR improves on the previous version but still has performance issues that need attention before merging
- Score reflects partial improvement: (1) DOM queries at lines 132-133 now execute only when `getContentProps` is called, which is better than running on every render, but they still execute on every callback invocation without caching when header/body elements are stable—implement ref-based caching or mutation observers to track actual DOM changes; (2) the `lg` size variant bug persists (uses `fontSize: "md"` instead of `"lg"` in the style file); (3) style token inconsistency mixing `{gap}` reference syntax with `--gap` CSS variable syntax remains unaddressed
- Pay close attention to `packages/react/src/components/native-popover/use-native-popover.tsx` lines 132-133—add ref-based caching to avoid repeated DOM queries when header/body presence hasn't changed, and verify that `native-popover.style.ts` has the correct `fontSize` value for the `lg` size variant


packages/react/src/components/native-popover/use-native-popover.tsx
```diff
@@ -0,0 +1,187 @@
...
+   *
+   * @default 'end'
+   */
+  placement?: Direction
+  /**
+   * The mode of the popover.
+   *
+   * @default 'auto'
+   */
+  popover?: "auto" | "manual"
```
odendayoko: @taroj1205
To align with the `Popover API` definition, `popover?: "" | "auto" | "manual" | "hint"` seems appropriate.
`popover?: "" | "auto" | "manual" | "hint"` seems appropriate.

https://developer.mozilla.org/ja/docs/Web/HTML/Reference/Global_attributes/popover#hint

odendayoko:
@taroj1205 
Thanks for the implementation! 
Sorry for the delay—reviewing this is taking longer than expected.
Please wait a little longer.

packages/react/src/components/native-popover/use-native-popover.tsx
```diff
@@ -0,0 +1,187 @@
...
+        "aria-disabled": ariaAttr(disabled),
+        "aria-haspopup": "dialog",
+        role: "button",
+        ...props,
+        ref: mergeRefs(ref, triggerRef, (node) => {
+          if (anchorRef.current == null) refs.setReference(node)
+        }),
+        type: "button",
+        popoverTarget: disabled ? undefined : contentId,
+        popoverTargetAction: disabled ? undefined : "toggle",
```
odendayoko: @taroj1205
It seems problematic that users cannot override `popoverTarget` and `popoverTargetAction`.

packages/react/src/components/native-popover/use-native-popover.tsx
```diff
@@ -0,0 +1,187 @@
...
+      return {
+        "aria-controls": contentId,
+        "aria-disabled": ariaAttr(disabled),
+        "aria-haspopup": "dialog",
+        role: "button",
+        ...props,
+        ref: mergeRefs(ref, triggerRef, (node) => {
+          if (anchorRef.current == null) refs.setReference(node)
+        }),
+        type: "button",
```
odendayoko: @taroj1205
Why is `type` fixed to `button`? This prevents overwriting it.

packages/react/src/components/native-popover/native-popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+    | "getTriggerProps"
+  > {}
+
+export interface NativePopoverRootProps
+  extends UseNativePopoverProps,
+    ThemeProps<NativePopoverStyle> {
+  /**
+   * The children of the popover.
+   */
+  children?: ReactNodeOrFunction<{}>
```
odendayoko: @taroj1205
I don't see the point of using `ReactNodeOrFunction`.
It seems better to use `PropsWithChildren` or explicitly specify `ReactNode`.

packages/react/src/components/native-popover/native-popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+      getHeaderProps,
+      getPositionerProps,
+      getTriggerProps,
+    ],
+  )
+
+  return (
+    <StyleContext value={styleContext}>
+      <ComponentContext value={componentContext}>
+        {runIfFn(children, {})}
```
odendayoko: @taroj1205
Just `children` should be fine.

odendayoko (changes requested):

odendayoko:
@taroj1205 
This is not directly related to the current fix, but the trigger's `Button` is extending beyond the `PropsTable`.
Components___NativePopover_-_Size_⋅_Storybook

taroj1205:
> @taroj1205 
> This is not directly related to the current fix, but the trigger's `Button` is extending beyond the `PropsTable`.
> Components___NativePopover_-_Size_⋅_Storybook
> 

Yeah, I've noticed it, it seems to be an issue with styling applied from layout center in storybook, when I removed width 100% from the table it seemed to be fixed, but it was happening to few other components as well.

packages/react/src/components/native-popover/use-native-popover.tsx
taroj1205: @odendayoko 
Thanks! I have updated it.

packages/react/src/components/native-popover/use-native-popover.tsx
taroj1205: @odendayoko 
Thanks! I have updated it.

packages/react/src/components/native-popover/use-native-popover.tsx
taroj1205: @odendayoko 
Thanks! I have updated it.

packages/react/src/components/native-popover/native-popover.tsx
taroj1205: @odendayoko 
Thanks! I have updated it.

packages/react/src/components/native-popover/native-popover.tsx
taroj1205: @odendayoko 
Thanks! I have updated it.

taroj1205:
I've written tests, please review again! @odendayoko

odendayoko (dismissed):

PR #5547, opened 2025-11-19
feat(qr-code): add qr code component

### LINKED ISSUE #4601: Implement `QR Code` component
### Description

We should implement QR Code component which displays a QR Code for a given input.

```tsx
<QrCode.Root value="...">
  <QrCode.Frame>
    <QrCode.Pattern />
  </QrCode.Frame>
</QrCode.Root>
```

We don't have a QR Code component.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

None

### Additional Information

_No response_

jji05:
Closes #4601 

## Description

I've created the `QrCode` component as requested in #4601. I've taken inspiration from Chakra UI, Zag JS, and Ark UI. 

I haven't started the documentation yet to first make sure the component is correct. 

> [!NOTE] 
> The `pnpm-lock.yaml` `lockfileVersion` was automatically updated to version `9.0`.

## Current behavior (updates)

There is no `QrCode` component

## New behavior

```ts
<QrCode.Root value="...">
  <QrCode.Frame>
    <QrCode.Pattern />
  </QrCode.Frame>
</QrCode.Root>
```

~~The `pnpm-lock.yaml` `lockfileVersion` was automatically updated to version `9.0`. Quite a lot of dependencies were changed, please let me know how I can fix this.~~ No longer an issue after rebasing to v2.1.


packages/react/src/components/qr-code/qr-code.style.ts
```diff
@@ -0,0 +1,75 @@
...
+      justifyContent: "center",
+      left: "50%",
+      position: "absolute",
+      top: "50%",
+      transform: "translate(-50%, -50%)",
+    },
+    pattern: {
+      "svg path": {
+        fill: "currentColor",
+      },
```
taroj1205: Is this supposed to be targetting `pattern > svg > path`? I think the `pattern` is already the `path` element.

packages/react/src/components/qr-code/qr-code.tsx
```diff
@@ -0,0 +1,109 @@
...
+    )
+
+    return (
+      <ComponentContext value={context}>
+        <styled.div {...getRootProps(rest)} />
+      </ComponentContext>
+    )
+  },
+  "root",
+)({ border: 2, ecc: "M" })
```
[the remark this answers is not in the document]
taroj1205: Since we are already defining `border` and `ecc` in line 55, we can just do `border=2` instead.

packages/react/src/components/qr-code/qr-code.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  return <styled.div {...getOverlayProps(rest)} />
+}, "overlay")()
+
+export const QrCode = {
+  Frame: QrCodeFrame,
+  Overlay: QrCodeOverlay,
+  Pattern: QrCodePattern,
+  Root: QrCodeRoot,
+}
```
taroj1205: Please check other components like `popover/namespace.ts` for they are exported.

jji05: Fixed

hirotomoyamada:
@jeffplays2005

Hi, Thanks for the PR!

A week has passed since this PR was created, so maintainers will be joining this PR.

packages/react/src/components/qr-code/qr-code.style.ts
jji05: I'm a little confused by this comment, could you please elaborate!

taroj1205: @jeffplays2005 
If you try to change the `currentColor` to something like "red" you won't see any change.
I think you meant to do this instead
```suggestion
    pattern: {
      fill: "currentColor",
```

jji05: I must've misunderstood your initial comment as when trying, that it seems to use the `color` prop over `fill`. Is this the expected behaviour? 

Screenshot 2025-12-29 at 2 47 09 PM

packages/react/src/components/qr-code/qr-code.tsx
```diff
@@ -0,0 +1,106 @@
...
+    )
+  },
+  "root",
+)({ border: 2, ecc: "L" })
+
+export interface QrCodeFrameProps
+  extends HTMLStyledProps<"svg">,
+    ThemeProps<QrCodeStyle> {}
+
+export const QrCodeFrame = withContext<"div", QrCodeFrameProps>((rest) => {
```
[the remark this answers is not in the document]
jji05: Fixed

packages/react/src/components/qr-code/qr-code.style.ts
taroj1205: What I mean is, the css styling you have written for pattern in `qr-code.style.ts` seems wrong.
You are trying to apply `fill="currentColor"` to
```html
<path class="ui-qr-code__pattern">
  <svg>
    <path fill="currentColor" /> <-- here
  </svg>
</path>
```
but the HTML for the QR Code component is
```html
<path class="ui-qr-code__pattern">
</path>
```
which means, this is not doing anything.
```ts
pattern: {
      "svg path": {
        fill: "currentColor",
      },
}
```

taroj1205: I see what you mean, I think we can just do this.
```suggestion
    pattern: {}
```

jji05: > I see what you mean, I think we can just do this.

I've also removed `pattern: {}`, is it still required?

packages/react/src/components/qr-code/use-qr-code.tsx
```diff
@@ -0,0 +1,80 @@
...
+    (props = {}) => ({ viewBox: `0 0 ${width} ${height}`, ...props }),
+    [height, width],
+  )
+
+  const getOverlayProps: PropGetter = useCallback(
+    (props = {}) => ({ ...props }),
+    [],
+  )
+
+  const getPatternProps = useCallback(
```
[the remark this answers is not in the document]
taroj1205: Please do this @jeffplays2005

packages/react/src/components/qr-code/qr-code.style.ts
taroj1205: Yes, please keep it to show slot presence.
image

taroj1205:
@jeffplays2005 Please change the base branch to `v2.1` as we want to release this in `v2.1`.

packages/react/src/components/qr-code/qr-code.tsx
```diff
@@ -0,0 +1,106 @@
+"use client"
+
+import type { HTMLStyledProps, ThemeProps } from "../../core"
+import type { QrCodeStyle } from "./qr-code.style"
+
+import { useMemo } from "react"
+import { createSlotComponent, styled } from "../../core"
```
taroj1205, line 7: Please remove the blank line between these imports.
```suggestion
import type { QrCodeStyle } from "./qr-code.style"
import { useMemo } from "react"
import { createSlotComponent, styled } from "../../core"
```

jji05:
> @jeffplays2005 Please change the base branch to `v2.1` as we want to release this in `v2.1`.

Done! Sorry I had to force push to rebase to `v2.1`, files you reviewed might show as changed but it is the same as before. The lock file also seems to be fine now with minimal changes as expected.

packages/react/src/components/qr-code/qr-code.style.ts
```diff
@@ -0,0 +1,70 @@
...
+      justifyContent: "center",
+      left: "50%",
+      position: "absolute",
+      top: "50%",
+      transform: "translate(-50%, -50%)",
+    },
+    root: {
+      position: "relative",
+    },
+  },
```
taroj1205, line 25: Please add back `pattern: {}`

jji05:
Oops, I accidentally updated the branch with `main`. Sorry will have to make a force push.

taroj1205 (dismissed):
LGTM Thanks!

hirotomoyamada (approved):

PR #5574, opened 2025-12-12
chore: update react and next dependencies

bmthd:
Closes #


taroj1205 (changes requested):
Please write changeset.

.changeset/twelve-lizards-flow.md
koralle, line 1: nitpicks [^1]

@bmthd 
Instead of just listing the CVE number, how about also including the URL of the page with a detailed description?
It would be very helpful for those viewing the changelog.

[^1]: This is a nitpicky suggestion. Feel free to ignore it if you prefer, no action is required.

PR #5588, opened 2025-12-15
add permissions for npm trustPolicy

### LINKED ISSUE #5577: Enforce trust with `trustPolicy`
### Description

We must set `trustPolicy` to `no-downgrade` in `.npmrc` (or `pnpm-workspace.yaml`).

> To further protect your supply chain, pnpm also supports a [trustPolicy](https://pnpm.io/settings#trustpolicy) setting. When set to no-downgrade, this setting will prevent installation of a package if its trust level has decreased compared to previous releases (for example, if it was previously published by a trusted publisher but now only has provenance or no trust evidence). This helps you avoid installing potentially compromised or less trustworthy versions.
>
> If you need to allow specific packages or versions to bypass the trust policy check, you can use the [trustPolicyExclude](https://pnpm.io/settings#trustpolicyexclude) setting. This is useful for known packages that may not meet the trust requirements but are still safe to use.

https://pnpm.io/supply-chain-security#enforce-trust-with-trustpolicy

bmthd:
Closes #5577

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,6 +10,10 @@ on:
     branches:
       - main
 
+permissions:
+  id-token: write
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
 


hirotomoyamada:
@bmthd 

What exactly is this doing? 
Why does adding `permissions` to the workflow mean you've implemented a trust policy?

bmthd:
@hirotomoyamada 
I forgot to include the context.
I have responded based on this document.

https://docs.npmjs.com/trusted-publishers

`npm publish` automatically determines if you have the necessary permissions, enabling the use of trusted policies.

hirotomoyamada:
@bmthd 

In that case, I don't think it has anything to do with trustPolicy, is that correct?

Also, this approach doesn't work because the next tag has a separate yaml file. If there is a way to associate an npm package tag with each file, please let me know.

bmthd:
Since it seems difficult to achieve with the current system, I will close this for now.

PR #5620, opened 2025-12-22
chore: update next version and unify the version in the catalog

bmthd:
I have updated Next.js to version 16.
Along with this, I have made the following changes:
- Renamed middleware.ts to proxy.ts, as this is a breaking change in Next.js.
- Unified Next.js versions in the pnpm catalog for centralized management.

Closes #


pnpm-workspace.yaml
```diff
@@ -3,4 +3,9 @@ packages:
   - "playgrounds/**"
   - "www/**"
 
+catalog:
+  "@next/eslint-plugin-next": ^16.1.0
+  next: ^16.1.0
```
taroj1205, line 8: should we pin these versions?

www/package.json
```diff
@@ -8,10 +8,10 @@
   "scripts": {
     "clean": "rimraf .next .velite .eslintcache tsconfig.tsbuildinfo node_modules",
     "dev:content": "velite --watch",
-    "dev:next": "next dev --turbopack",
+    "dev:next": "next dev",
     "dev": "run-p dev:*",
     "build:content": "NODE_ENV=production velite --clean",
-    "build:next": "next build  --turbopack",
+    "build:next": "next build ",
```
taroj1205: Please remove the extra whitespace. 
```suggestion
    "build:next": "next build",
```

www/next.config.ts
```diff
@@ -30,6 +30,7 @@ const nextConfig: NextConfig = {
   experimental: {
     externalDir: true,
     optimizePackageImports: ["@yamada-ui/react"],
+    turbopackFileSystemCacheForDev: true,
```
taroj1205: Isn't this on by default with 16.1?
https://nextjs.org/blog/next-16-1#turbopack-file-system-caching-for-next-dev

bmthd: Fixed.

www/package.json
bmthd: Fixed.

pnpm-workspace.yaml
bmthd: There is no difference from what is stated in package.json. The intention is to centralize versions scattered across multiple locations.

bmthd:
@hirotomoyamada Please review this!

PR #5621, opened 2025-12-22
chore: unify package versions across packages with pnpm catalog

bmthd:
Closes #


hirotomoyamada:
@bmthd 

Why do we need to use `catalog` so much? Isn't there a concern about combining them? Honestly, I don't think we need to combine everything. While I agree with the previous PR about `next` using `catalog`, I cannot agree with this.

Also, is publishing via `npm` okay when using `catalog`? Have you verified this?

bmthd:
@hirotomoyamada As you pointed out, there is no rational reason to put everything into the catalog, so we will close it.

PR #5642, opened 2025-12-27
Remove remix files and related configs

hirotomoyamada:
Closes #


hirotomoyamada:
> I think we need to update `.changeset/config.json` too.

Fixed.

PR #5645, opened 2025-12-27
fix(color-picker): pass aria-label and aria-labelledby to input element

### LINKED ISSUE #5643: Enhance a11y for `ColorPicker`
name: ColorPicker

### Description
1 accessibility violation was found in the `ColorPicker`.

#### [Components / ColorPicker / CustomControl](https://yamada-ui.github.io/yamada-ui/?path=/story/components-colorpicker--custom-control)

Received:
Form elements must have labels (label)

Fix any of the following:
  Element does not have an implicit (wrapped) <label>
  Element does not have an explicit <label>
  aria-label attribute does not exist or is empty
  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
  Element has no title attribute
  Element has no placeholder attribute
  Element's default semantics were not overridden with role="none" or role="presentation"

You can find more information on this issue here: 
https://dequeuniversity.com/rules/axe/4.11/label?application=axeAPI

### Additional Information

0pilatos0:
## Summary
- Fixes #5643
- Extracts `aria-label` and `aria-labelledby` from props and passes them to the input element
- Adds accessibility tests to verify the fix

## Root cause
The `useFieldProps` hook returns an `ariaProps` object that only contains field-related aria attributes (`aria-describedby`, `aria-disabled`, `aria-invalid`, `aria-readonly`, `aria-required`). 

The `aria-label` and `aria-labelledby` attributes were ending up in the `rest` object, which gets passed to `useCombobox` but not to `getInputProps`. This meant these accessibility attributes never reached the actual input element.

The fix explicitly extracts `aria-label` and `aria-labelledby` from the props and passes them directly to the input element in `getInputProps`.

## Test plan
- [x] Added test `passes a11y test with aria-label` - verifies no a11y violations with aria-label
- [x] Added test `applies aria-label to input element` - verifies aria-label is present on the input
- [x] All existing tests pass


hirotomoyamada:
@0pilatos0 

Hi, thanks for the PR.

The issue linked to this PR was created in error. Therefore, there is no problem to resolve.
I'm sorry to close it, especially since you took the trouble to create the PR.

PR #5668, opened 2026-01-22
fix: improve Turborepo caching and task dependencies

anthonyshew:
## Summary

- Enable caching for `lint`, `typecheck`, `format`, `test`, and `test:coverage` tasks (previously all had `cache: false`)
- Add transit node pattern for correct cache invalidation when dependency source changes
- Update schema URL to current Turborepo v2 schema

## Why

The previous config disabled caching for most tasks, causing unnecessary re-runs in CI. The transit node pattern allows `lint`, `typecheck`, and `format` to run in parallel while still invalidating cache correctly when upstream package source code changes.

## Testing

Run `turbo run lint typecheck format --dry` to verify tasks are recognized and dependencies are correct.

## AI usage
Chat transcript: https://opncd.ai/share/gObkWcya

### DIFF
--- a/www/contents/components/(components)/resizable.ja.mdx
+++ b/www/contents/components/(components)/resizable.ja.mdx
@@ -524,4 +524,28 @@ return (
 
 ## アクセシビリティ
 
-現在、v2の移行に伴い、このセクションは更新中です。
+`Resizable`は、アクセシビリティに関して[WAI-ARIA - Window Splitter Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/)に従います。
+
+### キーボード操作
+
+| キー         | 説明                                                                | 状態                       |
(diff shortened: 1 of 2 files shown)


taroj1205:
@anthonyshew Thanks!
I have a few questions if you don't mind;
1. What is the new transit task that was added?
2. Should we add coverage directory to the test:coverage task as output?

taroj1205:
I've just read the documentation about transit. 
From my understanding from the docs, transit nodes are used to properly include a package that does not include the script e.g. `build` in the dependency graph, allowing for more accurate caching. 
Is this correct? 
https://turborepo.dev/docs/core-concepts/package-and-task-graph#transit-nodes

hirotomoyamada:
@anthonyshew

Hi, Thanks for the PR!

A week has passed since this PR was created, so maintainers will be joining this PR.

PR #5681, opened 2026-01-26
pin 3rd-party action

koralle:
## Summary
  - Add workflow linting using actionlint to validate GitHub Actions workflow syntax and catch common
  errors
  - Pin all third-party GitHub Actions to specific commit hashes for improved supply chain security
  - Fix actionlint errors (deprecated `::set-output` → `$GITHUB_OUTPUT`)
  - Fix shell script issues (SC2090, SC2086) by properly quoting variables
  - Remove unnecessary push trigger for workflow linting
  - Improve workflow naming consistency across the repository

  ## Motivation
  Using commit hashes instead of version tags for third-party actions ensures reproducible builds and
  protects against supply chain attacks. If a malicious actor compromises a repository and updates a
  tag, our workflows will continue using the pinned commit hash.

  Additionally, actionlint provides static analysis for workflow files, catching syntax errors and
  anti-patterns that GitHub's own validation doesn't detect.

  ## Test plan
  - [x] Workflow linting passes on all `.github/**/*.yml` and `.github/**/*.yaml` files
  - [x] All third-party actions use pinned commit hashes
  - [x] No actionlint errors in existing workflows

  ## References
  - [actionlint checks](https://github.com/rhysd/actionlint/blob/v1.7.7/docs/checks.md)
  - [pinact documentation](https://github.com/suzuki-shunsuke/pinact)


.github/workflows/lint-workflow.yml
```diff
@@ -0,0 +1,38 @@
+name: Lint workflows
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/**/*.yml"
+      - ".github/**/*.yaml"
```
taroj1205: Is this necessary to run on push to main when we always make a PR to merge to main?

.github/workflows/lint-workflow.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  lint:
+    name: Lint workflows
+    runs-on: ubuntu-24.04
+
```
taroj1205: I think 'lint workflows' doesn't really make sense as a name.
Also, this seems to be the only workflow with ubuntu version specified.

.github/workflows/lint-workflow.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: jdx/mise-action@6d1e696aa24c1aa1bcc1adea0212707c71ab78a8 # v3.6.1
+        with:
+          mise_toml: |
+            [tools]
+            pinact = "latest"
+            actionlint = "latest"
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
taroj1205: Any reason for using mise specifically for this workflow?

koralle: I built a workflow based on this article. The reason I went with mise is that right now, it's the easiest way to get pinact and actionlint installed without much hassle.

https://tech.revcomm.co.jp/introduce-pinact-and-aqua

taroj1205: We should likely avoid using 'latest' in these tool versions. 
Is there also any reason for not using the `pinact-action` provided by the author?
https://github.com/suzuki-shunsuke/pinact-action

Also, what does the actionlint do that GitHub doesn't already do?

.github/workflows/lint-workflow.yml
taroj1205: @koralle This workflow naming isn't consistent with other workflows.

Image

.github/workflows/lint-workflow.yml
```diff
@@ -0,0 +1,32 @@
...
+      - ".github/**/*.yaml"
+
+jobs:
+  lint:
+    name: Lint workflows
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: jdx/mise-action@6d1e696aa24c1aa1bcc1adea0212707c71ab78a8 # v3.6.1
```
taroj1205: Please add `- name: ` like other workflow steps.

.github/workflows/lint-workflow.yml
koralle: @taroj1205 
Sorry, I forgot to delete it.

.github/workflows/lint-workflow.yml
koralle: @taroj1205 
I messed up. I fixed it.

.github/workflows/lint-workflow.yml
koralle: @taroj1205 

> We should likely avoid using 'latest' in these tool versions.

You're right.
I've decided to use suzuki-shunsuke/pinact-action for executing pinact.
Also, I've decided to hardcode and fix the version of actionlint to be installed with jdx/mise.

---

> Also, what does the actionlint do that GitHub doesn't already do?

Regarding that, please refer to this document: https://github.com/rhysd/actionlint/blob/v1.7.7/docs/checks.md#

.github/workflows/security.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+
+      - name: Check modified files
+        id: check-modified-packages
+        uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c # v46.0.5
+        with:
+          files: |
+            .github/**/*.{yml,yaml}
```
taroj1205: I think for now since we only have Pinact, we can just use `path` in the `on` field so it skips the workflow entirely.

koralle: @taroj1205 
I followed the existing workflow, but is it okay if I don't stick to it too closely?

taroj1205: @koralle 
There are workflows using `path`.
image

.github/workflows/security.yml
```diff
@@ -0,0 +1,22 @@
+name: Security
+
+on:
+  pull_request:
+    paths:
+      - ".github/**/*.(yml,yaml)"
+    branches:
```
taroj1205, line 8: It seems like the workflow is no longer running on this PR.
Maybe its supposed to be using `|`?
```suggestion
    paths:
      - ".github/**/*.(yml|yaml)"
    branches:
```

taroj1205 (approved):

PR #5702, opened 2026-02-15
docs(select): update a11y

### LINKED ISSUE #5477: Populate Accessibility section for `Select`
script: update-documentation
page: select-accessibility

### Subject

Accessibility section for `Select` (v2)

### Path

`www/contents/components/(components)/select.mdx`

`www/contents/components/(components)/select.ja.mdx`

### Description

Populate the Accessibility section for the `Select` component in v2 documentation, using the v1 docs as reference

odendayoko:
Closes #5477


www/contents/components/(components)/select.mdx
```diff
@@ -818,4 +818,73 @@ return (
...
+```
+
+### Keyboard Navigation
+
+| Key              | Description                                                                                       | State                  |
+| ---------------- | ------------------------------------------------------------------------------------------------- | ---------------------- |
+| `ArrowUp`        | Focuses the previous non-disabled item. If on the first item, focuses the last non-disabled item. | -                      |
+| `ArrowDown`      | Focuses the next non-disabled item. If on the last item, focuses the first non-disabled item.     | -                      |
+| `Home`           | Focuses the first non-disabled item.                                                              | -                      |
+| `End`            | Focuses the last non-disabled item.                                                               | -                      |
```
chida09: "non-disabled" should be "that is not disabled", and "If on the" should be "If it's the" to match the pattern used in other component docs.

Reference (Accordion):

https://github.com/yamada-ui/yamada-ui/blob/main/www/contents/components/(components)/accordion.mdx#L339-L344

odendayoko: fixed.

PR #5726, opened 2026-02-16
Updated `react-resizable-panels`

hirotomoyamada:
Closes # 

## Description

Updated `react-resizable-panels`.

Yes


packages/react/src/components/resizable/use-resizable.ts
```diff
@@ -31,243 +26,104 @@ import {
...
-      tagName: props.as ?? groupProps.as,
-      onLayout: fnAll(props.onLayout, groupProps.onLayout, onLayout),
+      style: {
+        height: undefined,
+        width: undefined,
+        ...rest.style,
+        ...props.style,
+      },
+      elementRef: mergeRefs(props.elementRef, ref),
+      groupRef: mergeRefs(props.groupRef, groupRef),
```
koralle: ask [^1]

Rather than executing `assignRef()` inside `useEffect()`, wouldn't it be more appropriate to add `controlRef` to `mergeRef()`?

```suggestion
      groupRef: mergeRefs(props.groupRef, groupRef, controlRef),
```

[^1]: Could you explain why you made this change?

hirotomoyamada: Fixed.

PR #5958, opened 2026-03-08
Refactor dependencies interface to extend dependents

hirotomoyamada:
Closes # 

## Description

Refactor dependencies interface to extend dependents.


claude-for-yamada-ui (approved):
LGTM. Clean refactoring — `Dependencies` correctly extends `Dependents` since it shares all three fields (`components`, `hooks`, `providers`) and only adds `externals`. The change is applied consistently in both `packages/cli/src/index.type.ts` and `scripts/generate/registries/index.ts`. All CI checks pass.

PR #6054, opened 2026-03-11
feat(docs): enhance search functionality with default locale support

### LINKED ISSUE #6053: feat(www): Support cross-language search — search both English and current locale indexes
## Description

When users view the documentation in a non-English locale (e.g., Japanese) and type an
English keyword into the search bar, no results appear even when matching pages exist in
the localized version. The search should also query the English index and return
deduplicated results — so users can find pages regardless of which language they search in.

Non-English users who are familiar with English component names or API terminology are
unable to find documentation pages by searching in English when the site language is set
to their native language. This creates a friction point where users have to either switch
the site language to English just to search or guess the translated term.

Image

## Proposed Solution or API

When the active locale is non-English, query **both** the English index and the current
locale's index simultaneously, then merge and deduplicate the results before displaying:

1. Run the search query against the current locale's index (e.g., `ja`).
2. Run the same query against the English (`en`) index.
3. Merge both result sets, removing duplicates (same page URL).
4. Display the deduplicated results to the user.

This approach requires **no language detection and no translation** — it simply broadens
the search scope to include English content while deduplicating entries so the same page
does not appear twice.

## Alternatives

- Search only the current locale index (current behavior — misses English queries).
- Full translation/detection pipeline (too complex, out of scope).

## Additional Information

- Affects: docs site search (`www` package)
- No external translation API or language detection library is needed

bmthd:
Closes #6053

## AI used


hirotomoyamada:
@bmthd 

Did you use AI to make these changes?

bmthd:
@hirotomoyamada Yes, i use.

PR #6060, opened 2026-03-12
ci: submit `coverage-final.json` to Codecov explicitly

koralle:
This PR adds tests for `@yamada-ui/cli`.
It seems Codecov was attempting to collect coverage for `@yamada-ui/cli` even though it wasn't sending `coverage-final.json`.

This PR explicitly modifies `@yamada-ui/cli` to send the `coverage-final.json` it generates to Codecov.


.github/workflows/codecov.yml
hirotomoyamada: Please change this line.

koralle: @hirotomoyamada 
Fixed.

PR #6093, opened 2026-03-13
Updated page titles for each section

hirotomoyamada:
Closes # 

## AI used

## Description

Updated page titles for each section.

No


claude-for-yamada-ui (approved):
LGTM! Clean and well-scoped changes:

1. Fixes incorrect `PageProps` type parameter from `"/[locale]/icons"` to `"/[locale]"` across all example pages
2. Updates each page to use section-specific titles (`menu.<section>`) instead of the generic `title` key

Both fixes are correct and consistent across all 7 example pages.

PR #6110, opened 2026-03-15
feat(cli): add --dry-run/-n option for CLI commands

### LINKED ISSUE #6069: feat(cli): Add --dry-run option to all CLI commands
## Description

Add a `--dry-run` option to all CLI commands (`add`, `update`, `init`, `theme`, `tokens`, `diff`) that simulates the command's execution without writing any files or installing any dependencies. This helps users preview what changes will be made before committing to them.

Currently, CLI commands like `add`, `update`, and `init` make file system changes and install dependencies without giving users a way to preview the effects first. A `--dry-run` flag is a common CLI pattern (used by tools like npm, git, rsync) that lets users safely inspect what a command would do before actually running it — reducing risk of unintended changes.

## Proposed Solution or API

Add a `--dry-run` / `-n` flag to each command:

```bash
yamada-ui add button --dry-run
# Output: (dry run) Would write: src/components/button/index.tsx
# Output: (dry run) Would install: @yamada-ui/button

yamada-ui update --dry-run
# Output: (dry run) Would update: src/components/button/index.tsx

yamada-ui init --dry-run
# Output: (dry run) Would create: yamada.config.ts
# Output: (dry run) Would install: @yamada-ui/core
```

Affected commands:
- `add` — show files that would be written and dependencies that would be installed
- `update` — show files that would be updated/overwritten
- `init` — show files that would be created and dependencies that would be installed
- `theme` — show files that would be created
- `tokens` — show files that would be generated
- `diff` — (already read-only, `--dry-run` could be a no-op or skipped)

## Alternatives

None considered — `--dry-run` is a widely understood and expected pattern for CLI tools.

## Additional Information

- Related pattern: `--yes` / `-y` option was added in #5980
- `--force` option was added to `update` in #5866
- Existing commands use `listr2` for task runners and `ora` for spinners — dry-run output should follow the same visual style

This issue is open to community contribution.

Related #5980
Related #5866

sh-1101:
Closes #6069

## AI used

## Description
Add a --dry-run option to all CLI commands (add, update, init, theme, tokens, diff) that simulates the command's execution without writing any files or installing any dependencies. This helps users preview what changes will be made before committing to them.

## Current behavior (updates)
Running any CLI command (e.g. `yamada-cli add button`) immediately writes files and installs dependencies with no way to preview what will happen.

## New behavior
Passing `--dry-run` (or `-n`) prints a log of every file write and dependency install that **would** occur, without actually performing any of them.

- `dryRun` follows the same pattern as existing flags (`--yes`, `--force`) and is passed via dedicated options types (`WriteFileOptions`, `InstallDependenciesOptions`, `GenerateSourceOptions`, `UpdateFilesOptions`).
- In `update-files.ts`, the early return occurs before `mkdtemp` so no temporary files are created either.

No


hirotomoyamada:
### Code review

Found 1 issue:

1. `--dry-run` does not guard `rimraf(outdirPath)` in `theme` and `init` commands — directories are actually deleted even when `--dry-run` is specified

Both `theme/index.ts` and `init/index.ts` call `rimraf(outdirPath)` when the user confirms overwriting an existing directory. The PR adds `dryRun` guards to `writeFileSafe` and `installDependencies`, but `rimraf` is called unconditionally. Running `--dry-run --yes` (or confirming the overwrite prompt) with an existing output directory will delete that directory, which is destructive and defeats the purpose of `--dry-run`.

- theme: https://github.com/yamada-ui/yamada-ui/blob/a2b3a1fb2dd93347bc70f791f96eda69a48ef349/packages/cli/src/commands/theme/index.ts#L180-L185
- init: https://github.com/yamada-ui/yamada-ui/blob/a2b3a1fb2dd93347bc70f791f96eda69a48ef349/packages/cli/src/commands/init/index.ts#L262-L267

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sh-1101:
Thank you for pointing this out. I had missed the `rimraf` calls. Wrapped the relevant code in the   `init` and `theme` commands with a `dryRun` guard.

hirotomoyamada:
### Code review (follow-up)

2 questions about the current implementation:

1. **Should `--dry-run` imply `--yes` (skip prompts)?**
Currently, running `--dry-run` without `--yes` still shows interactive prompts (e.g., overwrite confirmation in `init` and `theme`). Since `--dry-run` doesn't actually write anything, asking the user to confirm overwrites feels contradictory. Would it make sense to automatically skip prompts when `--dry-run` is set, or is the current behavior (requiring `--yes` separately) intentional?

2. **Is it intentional that merge conflict detection still runs during `--dry-run` in `update-files.ts`?**
In `updateFiles`, the `dryRun` guard is applied inside each task (early return before `mkdtemp`), but the overall flow — including merge conflict detection and diff3 validation — still executes normally. Is this by design (i.e., to show the user what *would* conflict), or should conflict handling also be skipped/reported differently during a dry run?

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

sorafujitani:
@sh-1101 
take over this PR.
make it a clone once the PR it's being taken over from is completed.

PR #6124, opened 2026-03-16
feat(cli): add view command to inspect and download component files

### LINKED ISSUE #6071: feat(cli): Add view command to inspect component source files
## Description

Add a `view` command to `@yamada-ui/cli` that allows users to inspect component source files before or without installing them — inspired by the shadcn CLI's `view` command.

Currently, there is no way to inspect the source files of a Yamada UI component from the CLI. Users must navigate the repository manually to understand what files make up a component. Since Yamada UI components consist of multiple files, a `view` command would help users explore component structure and show the contents of individual files.

## Proposed Solution or API

```
# Show directory listing of the component's files
yamada-cli view button

# Download/copy a specific file from the component
yamada-cli view button button.tsx
```

- Running \`view <component>\` with no file argument displays the list of files that make up the component (directory view).
- Running \`view <component> <file>\` downloads or copies the specified file to the user's project.

## Alternatives

Users currently browse the GitHub repository manually or rely on the \`add\` command, which installs everything at once without allowing selective inspection.

## Additional Information

Reference: [shadcn CLI \`view\` command](https://ui.shadcn.com/docs/cli#view)

SahilJat:
Closes #6071 

## AI used

- [ .] I did not use AI to create this PR.

## Description
This PR introduces the `view` command to the Yamada UI CLI, allowing users to inspect or download component source files directly from the registry without needing to install them via the `add` command.

## Current behavior (updates)
Yamada ui doesnt have a view command in cli currently

## New behavior

- Added `view` command logic in `packages/cli/src/commands/view/index.ts`

- **Flow 1:** Running `yamada-cli view <component>` prints a clean directory tree of available files.
- **Flow 2:** Running `yamada-cli view <component> <file>` downloads the raw code to the user's current working directory.
- Added full unit test coverage in `index.test.ts` mocking `fs` and network requests.

## Additional Information
1. Run `pnpm cli dev view button` to see the directory listing.
2. Run `pnpm cli dev view button button.tsx` to download the specific file to your current directory.


packages/cli/src/commands/view/index.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    consoleErrorSpy = vi
+      .spyOn(console, "error")
+      .mockImplementation(() => undefined)
+    processExitSpy = vi
+      .spyOn(process, "exit")
+      .mockImplementation((() => undefined) as any)
+
+    vi.clearAllMocks()
+  })
+
```
bmthd: Please avoid using any for the spy.
For better predictability, each test case should declare only the mocks it needs.

Also, it would be great if the test style could stay close to the existing module tests. Keeping the approach consistent with the current codebase makes it easier for us to review and merge.

packages/cli/src/index.ts
```diff
@@ -9,7 +9,7 @@ import { init } from "./commands/init"
 import { theme } from "./commands/theme"
 import { tokens } from "./commands/tokens"
 import { update } from "./commands/update"
-
```
bmthd, line 12: Please insert a line break.

packages/cli/src/index.ts
```diff
@@ -34,7 +34,7 @@ export function run() {
   program.addCommand(diff)
   program.addCommand(theme)
   program.addCommand(tokens)
-
```
bmthd, line 37: Please insert a line break.

packages/cli/src/commands/view/index.ts
```diff
@@ -0,0 +1,60 @@
...
+          `${c.green("✔")} Successfully downloaded ${c.bold(file)} from ${c.bold(component)} to your current directory.\n`,
+        )
+      } else {
+        console.log(
+          `\n${c.blue("Directory listing for:")} ${c.bold(component)}\n`,
+        )
+
+        sources.forEach((f: any) => {
+          console.log(`  ${c.dim("├─")} ${c.green(f.name)}`)
+        })
```
bmthd: Please also support nested structures.

packages/cli/src/commands/view/index.ts
```diff
@@ -0,0 +1,60 @@
...
+import c from "picocolors"
+import { fetchRegistry } from "../../utils"
+
+export const view = new Command()
+  .name("view")
+  .description(
+    "inspect component source files before or without installing them",
+  )
+  .argument("<component>", "the name of the component to view (e.g., button)")
+  .argument("[file]", "specific file to view or download (optional)")
```
bmthd: Please add tag support, just like add and diff.

packages/cli/src/commands/view/index.ts
```diff
@@ -0,0 +1,60 @@
...
+          console.error(
+            c.red(
+              `\nError: File '${file}' not found in component '${component}'.`,
+            ),
+          )
+          console.log(
+            `Available files are: ${sources.map((f: any) => f.name).join(", ")}\n`,
+          )
+          process.exit(1)
+        }
```
bmthd, line 40: Since we are mocking `process.exit()`, we should add a return statement to ensure that the control flow works correctly.

bmthd (commented):
Thanks for the PR! I left a few comments.

SahilJat:
hi @bmthd All set! Added the -t, --tag option to mirror the add command's architecture, and updated the directory listing to dynamically parse paths and print a cleanly formatted, recursive nested tree structure. Please let me know if there is something else i need to do.
Thanks

SahilJat:
@bmthd Added return statements after process.exit(1) to properly halt the control flow in the mocked test environment.

packages/cli/src/commands/view/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  .description(
+    "inspect component source files before or without installing them",
+  )
+  .argument("<component>", "the name of the component to view (e.g., button)")
+  .argument("[file]", "specific file to view or download (optional)")
+  .option("-t, --tag <name>", "tag for the registries (e.g. dev, next).")
+  .action(
+    async (component: string, file: string | undefined, { tag }: Options) => {
+      try {
+        const registryData = (await fetchRegistry(component, { tag })) as any
```
bmthd: Do not use `any`.

packages/cli/src/commands/view/index.ts
```diff
@@ -0,0 +1,102 @@
...
+              ),
+            )
+            console.log(
+              `Available files are: ${sources.map((f: any) => f.name).join(", ")}\n`,
+            )
+            return process.exit(1)
+          }
+
+          const targetPath = path.join(process.cwd(), file)
+          fs.writeFileSync(targetPath, targetFile.content, "utf-8")
```
bmthd: The expected behavior of this feature is to view the source file, not download it.
Please do not write any files.

SahilJat: hii @bhmtd in the original issue in the proposed solution this was mentioned there : 
`Running `view ` downloads or copies the specified file to the user's project.`
but if its not a requirement anymore , i will make sure to remove this specific code and will let the user watch the code in their terminal instead of download . Thanks for the comment.

SahilJat: hi @bmthd  updated the code as requested , removed the download and writing to the users project, please let me know if its good to go or do i need to do some more changes . Thanks

bmthd: Sorry about that.
The plan was incorrect — the view command should be dedicated solely to viewing.
Thanks for fixing it!

SahilJat:
hi @bmthd thanks for the refactor ! I noticed you split the `the view tree` in a different file - is this the standard pattern we want to use in the future cli commands? Just want to make sure I match this structure for my future contributions.

bmthd:
@SahilJat As you’ve already noticed, when multiple operations are performed within a single command, they should be separated into distinct functions based on each concept. Also, the error handling and testing approach needed to be aligned with the existing patterns, so I’ve taken care of that on my end.

We’re planning to merge this after Yamada’s review. Looking forward to your continued contributions.
thank you!

bmthd (dismissed):

bmthd (approved):

bmthd (dismissed):

bmthd (approved):

PR #6207, opened 2026-03-17
fix(icons): clear URL query params on search reset

### LINKED ISSUE #6205: I want to be able to save my search history
### Subject

Icons

### Description

I want to be able to save my search history. For example, I want the entire search results to be restored when I search for `"file"` and reload the page.

### Are you willing to participate in creating this documentation page and create a pull request?

Yes

### AI used

- [x] I did not use AI to create this issue.

hirotomoyamada:
Closes #6205

## AI used

## Description

Fixed `onReset` in the icons page to also clear the URL query parameter. Previously, resetting the search only cleared the local React state but left `?query=...` in the URL, causing the old filter to reappear on page refresh or when sharing the URL.

No

### DIFF
--- a/www/app/[locale]/(apps)/icons/list.tsx
+++ b/www/app/[locale]/(apps)/icons/list.tsx
@@ -123,10 +123,11 @@ export function List({ ...rest }: ListProps) {
 
   const onReset = useCallback(() => {
     setValue("")
+    replaceQuery({ query: "" })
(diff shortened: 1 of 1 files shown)


claude-for-yamada-ui (approved):
LGTM! The fix correctly clears URL query params on search reset by calling `replaceQuery({ query: "" })`, and the dependency array is properly updated.

Issue #6040, opened 2026-03-11
There is a discrepancy between Codecov and local coverage.


hirotomoyamada:
For example, the Codecov coverage for [avatar-group.tsx](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages%2Freact%2Fsrc%2Fcomponents%2Favatar%2Favatar-group.tsx) and [number-input.tsx](https://app.codecov.io/gh/yamada-ui/yamada-ui/blob/main/packages%2Freact%2Fsrc%2Fcomponents%2Fnumber-input%2Fnumber-input.tsx) is less than 100%, but the local coverage is 100%.

PR #6208, opened 2026-03-18
ci(test/coverage): refine test workflow with sharding

koralle:
Related to #6040  

## AI used


.github/workflows/codecov.yml
```diff
@@ -29,9 +78,89 @@ jobs:
...
-        run: pnpm test:coverage -- --shard=${{ matrix.shard_index }}/4
+        working-directory: packages/react
+        run: |
+          pnpm vitest run \
+            --browser=chromium \
+            --coverage \
+            --reporter=blob \
+            --reporter=verbose \
+            --shard=${{ matrix.shard_index }}/${{ matrix.shard_total }}
+
```
taroj1205, line 89: We should support turborepo cache for this too.

.github/workflows/codecov.yml
```diff
@@ -29,9 +78,89 @@ jobs:
...
+      - name: Merge reports
+        working-directory: packages/react
+        run: |
+          pnpm vitest run \
+            --browser=chromium \
+            --coverage \
+            --merge-reports
+
+      - name: Upload merged coverage to GitHub Actions Artifacts
```
taroj1205, line 127: I think we should consider using turborepo cache for this too.

taroj1205 (commented):
Maybe these are better namings? The current one is much longer than other job names, making it inconsistent.

- Check Coverage Changes
- Utils
- CLI
- Chromium ${{ matrix.shard_index }}/4
- Merge Coverage
- Codecov Upload

PR #6221, opened 2026-03-19
docs(instructions): prohibit use of --no-verify for ai commits

### LINKED ISSUE #6028: Add Skill to Enforce ESLint and Prettier on Every Code Change
## Description

Ensure that ESLint and Prettier are always applied when Claude Code modifies files in the yamada-ui repository.

Code changes made by AI agents may not follow the project's formatting and linting rules. A skill can automatically run `pnpm exec prettier <file> --write` after any file edit and check for ESLint violations.

## Proposed Solution or API

- Automatically run `pnpm exec prettier <file> --write` after any file edit
- Run ESLint checks and auto-fix where possible
- Warn on any remaining lint errors

## Alternatives

None considered.

Yes

## Additional Information

Parent issue: #6023

Jin5555C:
Closes #6028

## AI used

## Description
This PR adds clear AI commit safety guidance to prevent bypassing repository hooks.

## Current behavior (updates)
There was no explicit, shared instruction requiring AI agents to avoid git commit --no-verify. As a result, hook bypass behavior was not explicitly prohibited in AI-facing guidance.

## New behavior
* Added a Git commit policy for AI agents that explicitly prohibits using git commit --no-verify.
* Clarified that local hooks (pre-commit, commit-msg) must not be bypassed.
Added guidance to fix hook failures instead of skipping them.
Added an explicit exception path: if bypass is absolutely necessary, the agent must request user approval first.
Aligned instruction intent across AI-facing instruction files to reduce policy drift.
Is this a breaking change (Yes/No):
No.

## Additional Information
* This change is documentation/instruction policy only.
* No runtime code paths, public APIs, or package behavior were modified.
* The purpose is to improve consistency and safety of AI-assisted commit workflows.
* Copilot uses `copilot-instructions.md` as its repository instruction source, while other agents in this project follow `AGENTS.md`. This update keeps the commit policy aligned across both instruction paths to ensure consistent agent behavior.


Jin5555C:
Results of the execution:

### step 1. make demo file which has eslint error.

```
/* eslint eqeqeq: "error" */

export const shouldFailLintDemo = (): boolean => {
  // Intentional ESLint violation for hook behavior testing.
  return 1 == 1
}
```

### step 2. try to commit with asking AI " please commit this file with `--no-verify`  option"

### Result: 
image

AGENTS.md
```diff
@@ -8,6 +8,13 @@ Yamada UI is a React UI component library built with CSS-in-JS (Emotion).
 - **Accessibility is required**: All components must support ARIA attributes, keyboard navigation, and screen readers. Report any concerns.
 - **Do not bundle multiple fixes**: If you encounter a separate issue while working on a fix, do not fix it in the same PR. Create a separate issue and submit a separate PR.
 
+## Git Commit Policy
+
+- Never use `git commit --no-verify`.
+- Never bypass local hooks (`pre-commit`, `commit-msg`) during commits.
+- If hooks fail, fix the underlying issues and commit again without bypass options.
+- If a bypass seems absolutely necessary, stop and ask the user for explicit approval before proceeding.
+
```
chida09: Thank you. 

Since AGENTS.md is designed for code-based contexts (structure, conventions, workflows), such operational constraints feel somewhat out of scope.

With Claude Code, commands can be mechanically blocked using `permissions.deny` in `.claude/settings.json`, making that a more reliable approach. Since other agents do not yet have equivalent mechanisms, AGENTS.md likely won’t provide sufficient enforcement in any case.               

For now, I think starting with Claude Code's settings.json would be sufficient — what do you think? We could also wait for other members' reviews before deciding.

suborajin:
@chida09 
Thank you for your suggestion. I fixed it based on our discussion.

.claude/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "permissions": {
+    "deny": ["Bash(git commit --no-verify*)"]
```
chida09: It would be better to include -n as well, since users might use the short flag.

https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---no-verify
https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--n

chida09: Since `.claude` is a symlink pointing to `.agents`, placing `settings.json` directly in `.agents/` would make it accessible as .`claude/settings.json` without removing the symlink. 

Could you move `.claude/settings.json` to `.agents/settings.json` and keep the `.claude` symlink intact?

Jin5555C:
@chida09 
Thank you for your review.
I fixed as the suggestion. Please check it.

.claude
chida09, line 1: This change is from a symbolic link to a regular file.
I want it to be a symbolic link, so please cancel the change.

.agents/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "permissions": {
+    "deny": ["Bash(git commit *--no-verify*)", "Bash(git commit *-n*)"]
```
chida09: The pattern `"Bash(git commit *-n*)"` is intended to block `git commit -n` (the short form of `--no-verify`), but it may also match unintended cases such as commit messages containing `-n` (e.g., `git commit -m "fix-name"`).                                                                          
                                                                                                                                                       
Consider using more specific patterns like `"Bash(git commit *--no-verify*)"` and `"Bash(git commit *-n)"` to keep both forms while avoiding false positives.

.agents/settings.json
```diff
@@ -0,0 +1,5 @@
...
+  }
```
taroj1205, line 9: What happens if agents do `git commit -mn 'message'`?

Jin5555C:
@chida09  @taroj1205 
I added three matching patterns:

"Bash(git commit -n *)",
"Bash(git commit * -n)",
"Bash(git commit -*n*)"

However, this causes false positives such as `git commit -m "-name fix"`.

When trying to cover combined short options (like `-mn`), this issue seems unavoidable.

If you have any ideas to improve this, please let me know.

taroj1205:
> @chida09  @taroj1205 
> I added three matching patterns:
> 
> "Bash(git commit -n *)",
> "Bash(git commit * -n)",
> "Bash(git commit -*n*)"
> 
> However, this causes false positives such as `git commit -m "-name fix"`.
> 
> When trying to cover combined short options (like `-mn`), this issue seems unavoidable.
> 
> If you have any ideas to improve this, please let me know.

maybe you can try 
```sh
Bash(git commit *-[a-zA-Z]*n* *)
```

could you test it locally?

Jin5555C:
@taroj1205 @chida09 
Thank you for reviewing this multiple times and advice.
I’ve made the updates. Could you please review it again?

and I tested the short-form command locally.
It works as below

**result**
スクリーンショット 2026-04-06 164622

chida09:
@taroj1205  @Jin5555C 

Does `git commit -mn "message"` even work as a Git command, regardless of Claude Code's settings?
https://git-scm.com/docs/git-commit

Screenshot 2026-04-06 at 23 23 46

chida09:
On the other hand, `git commit -nm "message"` does work as a bypass (equivalent to `--no-verify -m "message"`), but it is not caught by the current patterns. 

`"Bash(git commit -*n*)"` causes false positives — any commit message containing `n` is incorrectly blocked.

Examples:
  - `git commit -m "fix(button): note"` → blocked ❌
  - `git commit -m "fix navigation"` → blocked ❌

To cover `git commit -nm "message"` without false positives, consider using a specific pattern instead:
`"Bash(git commit -nm *)"`

taroj1205:
@chida09 You're right, `-mn` doesn't seem to work, please ignore it.

.agents/settings.json
taroj1205: Doesn't work.

chida09 (approved):

taroj1205 (approved):

hirotomoyamada:
@taroj1205 @Jin5555C @chida09 

Which AI was used to verify this? Also, which AI systems are compatible, and which are not?

chida09:
@hirotomoyamada
Thanks for the review!

**Which AI was used to verify this:**
I verified this with Claude Code.

**Compatibility:**
We looked into how each agent enforces command-level deny rules:

- **Hard block (cannot be bypassed):** Claude Code — `settings.json` `permissions.deny` blocks execution at the system level.
- **Soft block (auto-execution prevented, but user approval can override):** GitHub Copilot, Windsurf, RooCode, Cline (CLI).
- **No deny mechanism:** Cursor (removed), Cline (VS Code), Aider, JetBrains Junie, OpenAI Codex — for these, instruction files (`AGENTS.md` / `copilot-instructions.md`) are the only available guardrail, and compliance depends on the model's reasoning.

**Background on the design choice:**
A few considerations led us here. First, instruction files referenced from `AGENTS.md` have context limits (CLAUDE.md ~200 lines, copilot-instructions.md 4,000 chars), so we wanted to keep them focused on always-on rules (structure, naming, etc.) rather than individual command-level restrictions. 

Second, even when a rule is written in `AGENTS.md`, compliance ultimately depends on the model's reasoning. Trying to cover every agent uniformly is hard, so we prioritized mechanisms that can actually enforce the rule. 

Claude Code is the only agent that provides a hard block, so we chose to make `.claude/settings.json` the source of enforcement, since several of us on the discussion thread use Claude Code.

Details of the discussion: 
https://discord.com/channels/1114670051900010526/1484175466662203513/1485492862471634975

hirotomoyamada:
@chida09 

First, does Claude reference `.agents`? `.claude` isn't linked to `.agents`.

Also, if this only works for Claude, I think you should configure it separately in `settings.local.json`.

chida09:
@hirotomoyamada 

## On the symlink

You're right. I did add a fix to restore the symlink in 7acc05ab, but it ended up not making it into the final state of this PR. I'll re-add the symlink to address this.

## On settings.local.json

Just to make sure I understand your proposal correctly: are you suggesting that instead of committing this as a shared project setting(`.claude/settings.json`), each developer using Claude Code should configure it locally in their own `.claude/settings.local.json`?

`.claude/settings.local.json` is automatically gitignored by Claude Code and treated as personal overrides ([docs](https://code.claude.com/docs/en/settings#settingslocaljson)).

hirotomoyamada:
@chida09 

If these are settings specific to Claude, isn’t it unnecessary to put them in the repository? It seems like each contributor could just configure their own `settings.local.json` file.

chida09:
@hirotomoyamada 
For context on where this came from: the original issue (#6028) was about ensuring ESLint are consistently applied when AI agents modify files. 

The pre-commit hooks already enforce this on normal commits, but AI agents can bypass them with --no-verify, so this PR tried to block that at the platform level for Claude Code.

That said, if you'd prefer to leave this up to each contributor's own configuration, I'm fine with closing this PR.

taroj1205:
@Jin5555C @chida09 @hirotomoyamada how about using block-no-verify package with pre tool use hook?
https://www.npmjs.com/package/block-no-verify

hirotomoyamada:
@taroj1205 @chida09 

Since this isn't a major issue, I'm closing this PR. If the problem occurs frequently, we can discuss it then.

PR #6298, opened 2026-03-29
perf(react): reduce test execution time

koralle:
## AI used

### Test execution time

In `packages/react`:

#### Before

```
 Test Files  6 failed | 226 passed (232)
      Tests  6 failed | 2762 passed | 2 todo (2770)
   Start at  18:27:04
   Duration  216.90s (transform 300.70s, setup 15.00s, import 1781.94s, tests 712.78s, environment 113.37s)
```

#### After

```
 Test Files  232 passed (232)
      Tests  2768 passed | 2 todo (2770)
   Start at  18:08:57
   Duration  93.51s (transform 68.30s, setup 12.37s, import 387.11s, tests 469.25s, environment 
84.49s)
```

### DIFF
--- a/.changeset/light-pianos-talk.md
+++ b/.changeset/light-pianos-talk.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+perf(react): reduce test execution time


taroj1205 (commented):
I can see what you are trying to do, and I agree with the goal, but this causes a lot of inconsistencies across the codebase.
I think we should discuss about this more carefully in our weekly meeting.

PR #6309, opened 2026-03-29
Updated button sizes to use dynamic height variables

hirotomoyamada:
Closes # 

## AI used

## Description

Updated button sizes to use dynamic height variables.

No


claude-for-yamada-ui (approved):
LGTM! Clean refactoring — extracting the repeated size token into a `--height` CSS variable for each button size eliminates duplication across `h`, `lineHeight`, and `minW` without changing the rendered styles. All CI checks pass.

PR #6326, opened 2026-03-31
Handle CSS variables and interpolations correctly

hirotomoyamada:
Closes # 

## AI used

## Description

Handle CSS variables and interpolations correctly.

No


claude-for-yamada-ui (approved):
LGTM! The fix correctly prevents double-wrapping of CSS variables and interpolation values in `injectVars`. When a value is already a `var(--...)` or contains `{...}` interpolation syntax, it's now passed through as-is instead of being incorrectly wrapped in `{token.value, value}` format.

PR #6340, opened 2026-04-03
Run eslint and prettier after writing files

hirotomoyamada:
Closes # 

## AI used

## Description

With the current approach, the generated file ends up showing differences every time. This is because, even though ESLint resolves the issues, Priter still flags them as formatting violations.

No


claude-for-yamada-ui (approved):
LGTM! ESLint の後に Prettier を実行することで、生成ファイルの差分が毎回出る問題を正しく解決しています。ESLint → Prettier の実行順序も適切です。

PR #6344, opened 2026-04-04
fix(next): correct scroll behavior on page navigation

### LINKED ISSUE #6300: Incorrect scrolling on page navigation
### Provide the exact quote of the error or issue

When navigating to a different section from scrolled state, the page does not correctly scroll to the top.
This results in the H1 being hidden under the header.

https://github.com/user-attachments/assets/040fbe15-bc09-4ea8-82a8-b7bb484a36dc

### Does the docs page already exist? Please link to it.

_No response_

None

### AI used

- [x] I did not use AI to create this issue.

aryanjsingh:
## Summary

This fix addresses an issue where the page would not scroll to the top upon navigation in the Next.js playground. The root cause was the custom scroll reset behavior in `UIProvider`, which conflicts with layouts using a sticky header. By disabling this custom behavior via the `config` prop, we allow Next.js's native scroll restoration to function correctly, ensuring the page content is not hidden on navigation.

## Changes

- **playgrounds/next/app/app/layout.tsx**: Disable the custom scroll reset in `UIProvider` by setting `config={{ resetScroll: false }}`. This prevents the library's scroll handling from interfering with Next.js's native scroll restoration, which correctly handles scrolling to the top of the page on navigation, even with sticky headers.

## Related Issue

Closes #6300

### DIFF
--- a/playgrounds/next/app/app/layout.tsx
+++ b/playgrounds/next/app/app/layout.tsx
@@ -33,7 +33,9 @@ export default async function RootLayout({
         <ColorModeScript type="cookie" />
         <ThemeSchemeScript type="cookie" />
 
-        <UIProvider cookie={cookieStore.toString()}>{children}</UIProvider>
+        <UIProvider config={{ resetScroll: false }} cookie={cookieStore.toString()}>
+          {children}
+        </UIProvider>
       </body>
     </html>
   )


hirotomoyamada:
Have you read the [AI Policy](https://github.com/yamada-ui/yamada-ui/blob/main/AI_POLICY.md)? This PR violates the policy and does not resolve the issue. I am marking this PR as spam and closing it.

PR #6371, opened 2026-04-06
chore(www): improve cache for build:content script

taroj1205:
Closes # 

## AI used

## Description

I have improved cache for `build:content` script by updating the input files.


hirotomoyamada:
Thanks for working on this! The direction is good — currently `build:content` has no explicit `inputs` in `www/turbo.json`, so Turborepo hashes every file in the package and unrelated changes (e.g. `app/**`, `components/**`, `theme/**`) invalidate the cache. Narrowing the inputs is a worthwhile improvement, and the list looks comprehensive: it covers everything `velite.config.ts` actually reads (`contents/**`, `@/constants`, `@/utils/*`, `@/scripts/generate/docs/*`, `@yamada-ui/workspace/prettier`). Excluding `data/doc-map.*.json` from inputs is also correct, since it's an output of this task and would otherwise cause self-invalidation.

### Concern: this may unintentionally drop `outputs` and `dependsOn` from the root config

The root `turbo.json` defines:

```json
"build:content": {
  "dependsOn": ["transit"],
  "outputs": [".velite/**", "data/doc-map.*.json"]
}
```

When a package-level `turbo.json` redefines a task via `extends`, the task definition replaces the root one rather than deep-merging field by field. If that's the case here, this PR would silently lose both `outputs` and `dependsOn: ["transit"]`, which would cause:

1. On a fresh CI checkout, a cache hit would replay logs but **not restore** `.velite/**` or `data/doc-map.*.json`, breaking the downstream `format` / `lint` / `typecheck` tasks that depend on them.
2. The dependency on `transit` (which builds workspace packages) would be removed, so changes to `@yamada-ui/workspace/prettier` (already listed in `inputs`) wouldn't actually trigger a rebuild of its dist before `build:content` runs.

### Suggested verification

Could you verify this locally before we merge? Something like:

```bash
rm -rf www/.velite www/data/doc-map.*.json
pnpm turbo run build:content --filter=www   # first run, cache miss
rm -rf www/.velite www/data/doc-map.*.json
pnpm turbo run build:content --filter=www   # second run, should be a cache HIT and must restore outputs
ls www/.velite www/data/doc-map.*.json       # confirm files are back
```

If the outputs aren't restored on the second run, please also copy `outputs` and `dependsOn: ["transit"]` into `www/turbo.json` explicitly:

```json
"build:content": {
  "dependsOn": ["transit"],
  "outputs": [".velite/**", "data/doc-map.*.json"],
  "inputs": [
    "velite.config.ts",
    "contents/**",
    ...
  ]
}
```

Other than that, LGTM once the caching behavior is confirmed. Thanks!

taroj1205:
@hirotomoyamada Hi I've tested this locally, and it seems to work as expected.
CleanShot 2026-04-07 at 13 54 27@2x

PR #6373, opened 2026-04-06
ci(quality): add 10 minute timeouts to quality and test workflows

taroj1205:
## AI used

## Description

Adds explicit `timeout-minutes: 10` to Quality workflow jobs and to jobs in the reusable Test workflow so hung runs fail fast instead of running until the default 6-hour limit.

## Current behavior (updates)

Quality and Test workflows rely on the default job timeout (up to 6 hours).

## New behavior

Each job has a 10-minute cap. If a job stalls, it is cancelled instead of consuming runner time indefinitely.

No

## Additional Information

If CI times out on install, Playwright, or coverage, consider raising timeouts only for the test jobs (and the Quality job that calls `test.yml`), not lint/format/typecheck.

### DIFF
--- a/www/app/[locale]/(apps)/icons/preview-drawer.tsx
+++ b/www/app/[locale]/(apps)/icons/preview-drawer.tsx
@@ -201,7 +201,11 @@ function CopyButton({ value }: CopyButtonProps) {
   const { copied, onCopy } = useClipboard(value)
 
   return (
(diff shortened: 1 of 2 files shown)


hirotomoyamada (changes requested):
Please delete `timeout-minutes` from `test`.

PR #6576, opened 2026-04-11
refactor(calendar): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6419: refactor(calendar): propagate suppressHydrationWarning to hidden elements
### Description

Some internal elements inside the `Calendar` component are not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to them.

Affected elements:

- `CalendarMonth` `<styled.thead aria-hidden>` at [calendar.tsx:539](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/calendar/calendar.tsx#L539) — no props spread.
- `CalendarMonth` day label `<styled.span>` at [calendar.tsx:561](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/calendar/calendar.tsx#L561) — default day cell wrapper, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6419

## AI used

## Description

Propagated `suppressHydrationWarning` set on the `Calendar` root to hidden inner elements inside `CalendarMonth` that previously did not receive the merged rest props.

- `CalendarMonth` `<styled.thead aria-hidden>` now spreads `useHydrationContext()`.
- `CalendarMonth` default day label `<styled.span>` (used when no custom day renderer is provided) now spreads `useHydrationContext()`.

No


claude-for-yamada-ui (changes requested):
## Review — request changes

### 1. Necessity: `suppressHydrationWarning` on `<styled.thead>` is ineffective (anti-pattern #2)

**File:** `packages/react/src/components/calendar/calendar.tsx` (the `<styled.thead aria-hidden>` hunk)

`suppressHydrationWarning` only suppresses mismatches on the element's **own attributes** and its **direct text content** (React docs are explicit — it does not cascade to children). The `<thead>` element:

- Has only the static attribute `aria-hidden`.
- Has no direct text content — its children are `<CalendarWeekDays>` / `<CalendarWeekday>` components.

Therefore `suppressHydrationWarning` has **no observable effect** on this element. The prop should be removed from the `<thead>`, and the corresponding test ("propagates suppressHydrationWarning to the hidden thead") should be removed as well.

### 2. Correctness: prop precedence regression in `getMonthProps` call

**File:** `packages/react/src/components/calendar/calendar.tsx` (return value of `CalendarMonth`)

Before the change:
```ts
getMonthProps({ ...monthProps, ...rest })
```
`rest` contains user-passed props (including `suppressHydrationWarning`), so user props win over `monthProps` (slot/context props). After the change:
```ts
getMonthProps({ suppressHydrationWarning, ...monthProps, ...rest })
```
Now `suppressHydrationWarning` is placed **before** `...monthProps`, so `monthProps` could shadow it. This reverses the intended precedence.

**Suggested fix:** Do not destructure `suppressHydrationWarning` out of `rest`. Instead, extract it non-destructively for use in the JSX children:

```ts
)(undefined, ({ day: dayProp, ...rest }) => {
  const { suppressHydrationWarning } = rest
  // ... use suppressHydrationWarning in <styled.span> only ...
  return {
    ...getMonthProps({ ...monthProps, ...rest }), // unchanged from before
    children,
  }
})
```

This preserves the original prop merge order while still giving access to the value for the `<span>`.

### 3. The `<styled.span>` propagation is justified

The `<styled.span>{label}</styled.span>` renders `label` as direct text content. When `format.day` is configured, `label` is produced by `dateTimeFormat` which is locale/timezone-sensitive and can legitimately differ between SSR and hydration. `suppressHydrationWarning` has a real effect here. This site is correct.

### Summary of requested changes

1. Remove `suppressHydrationWarning` from `<styled.thead>` and delete the thead-specific test.
2. Fix the prop precedence regression by not destructuring `suppressHydrationWarning` out of `rest` (extract it non-destructively instead).

codex-for-yamada-ui (changes requested):
### Request Changes

1. **Correctness / Consistency: `suppressHydrationWarning` precedence changed unintentionally**
   - **File:** `packages/react/src/components/calendar/calendar.tsx` (around new `getMonthProps` call in `CalendarMonth`, diff hunk near line 576)
   - **Issue:** The new merge order:
     ```ts
     getMonthProps({ suppressHydrationWarning, ...monthProps, ...rest })
     ```
     makes `monthProps.suppressHydrationWarning` override the explicitly passed `CalendarMonth` prop, which is opposite to the existing slot merge convention (`{ ...slotProps, ...rest }`) used across this file.
   - **Why this matters:** This is a behavioral regression in prop precedence and can produce inconsistent behavior between the table and propagated hidden elements when both sources provide the prop.
   - **Suggested fix:** Resolve the value explicitly, then apply it last:
     ```ts
     const resolvedSuppressHydrationWarning =
       suppressHydrationWarning ?? monthProps?.suppressHydrationWarning

     getMonthProps({
       ...monthProps,
       ...rest,
       suppressHydrationWarning: resolvedSuppressHydrationWarning,
     })
     ```
     and use `resolvedSuppressHydrationWarning` for `<styled.thead>` and `<styled.span>` as well.
   - **Test suggestion:** Add a case covering conflict between direct `CalendarMonth` prop and `monthProps` to lock expected precedence.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

Issue #6457, opened 2026-04-11
`Dropzone` overwrites user props via JSX double spread


hirotomoyamada:
### Which package has an issue?
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Dropzone`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction
N/A — code audit; reproducible by passing `className` and `onClick` to the affected sub-component and observing that the context-side value is replaced.

### Steps to Reproduce
1. Render the component with `className` and `onClick` (or `style`) set by the user.
2. Inspect the DOM node.
3. Observe that the context-side props are replaced instead of merged.

### Browser
Chrome

### Operating System
- [x] macOS

### Do you know how to fix the issue?
Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix?
Yes

### Additional Information
Use `mergeProps(contextProps, userProps)()` from `packages/react/src/core/components/props.ts` instead of plain spread. See `packages/react/src/components/sidebar/sidebar.tsx` for the reference fix pattern.

**Affected locations:**

- [`packages/react/src/components/dropzone/dropzone.tsx:100`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/dropzone/dropzone.tsx#L100) — `{...varProps} {...getRootProps()}`

Parent tracking: #6430

### AI used
- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

hirotomoyamada:
Closing as not-a-bug.

The reported spread `{...varProps} {...getRootProps()}` at line 100 is not actually unsafe:

```tsx
const varProps = useInputBorder({ errorBorderColor, focusBorderColor })
const { ..., getRootProps } = useDropzone(rest)
return (
  <styled.div data-group {...varProps} {...getRootProps()}>
    ...
  </styled.div>
)
```

1. `varProps` from `useInputBorder` only contains CSS custom properties (`--error-border-color`, `--focus-border-color`), no `className`/`style`/`css`/`ref`/`on*` keys — spreading it cannot clobber anything.
2. `getRootProps()` is called with no arguments. The user's top-level props flow into `rest` (in `useDropzone`'s closure) and are folded into the getter's output internally.

[...]

PR #6583, opened 2026-04-11
fix(radio-card): merge user props with context props via mergeProps

### LINKED ISSUE #6434: `RadioCardRoot` overwrites user `className`/`style` via unsafe spread
@yamada-ui/react

### Package Version
2.2.0

### Description
In `RadioCard`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/radio-card/radio-card.tsx:136`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/radio-card/radio-card.tsx#L136) — `RadioCardRoot`

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6434

## AI used

## Description

Replaces unsafe object spread with `mergeProps` in `RadioCardRoot` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.

Follows the reference pattern established in `sidebar.tsx` and the sibling fix in `radio.tsx` (#6433).

## Current behavior (updates)

Passing `className` / `style` / `onClick` via `rootProps` to `RadioCardRoot` would replace context-side values instead of merging with them.

## New behavior

User-provided props are now merged with context props, preserving internal className, styles, refs, and handlers. Added a regression test that renders a `RadioCard.Root` with `rootProps={{ className, style, onClick }}` and asserts that both the internal and user-provided values are present.

No

## Additional Information

### DIFF
--- a/.changeset/radio-card-root-merge-props.md
+++ b/.changeset/radio-card-root-merge-props.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Use `mergeProps` in `RadioCardRoot` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.
--- a/packages/react/src/components/radio-card/radio-card.test.tsx
+++ b/packages/react/src/components/radio-card/radio-card.test.tsx
@@ -1,4 +1,5 @@
-import { a11y, render, screen } from "#test"
+import { a11y, fireEvent, render, screen } from "#test"
+import { vi } from "vitest"
 import { RadioCard, RadioCardGroup } from "."
 
 const items = [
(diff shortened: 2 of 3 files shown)


hirotomoyamada:
Closing as a false positive, per the same rule applied to #6585 / #6437 / #6454–#6457.

The only change in this PR is on `packages/react/src/components/radio-card/radio-card.tsx`:

```tsx
-      <styled.label {...getRootProps({ ...varProps, ...rootProps })}>
+      <styled.label {...mergeProps(getRootProps(varProps), rootProps)()}>
```

`varProps` comes from `useInputBorder` and only contains CSS custom properties (`--error-border-color`, `--focus-border-color`) — it has no mergeable keys (`ref`, `className`, `style`, `css`, or `on*`). Since `varProps` carries nothing that `rootProps` could clobber, `{ ...varProps, ...rootProps }` is already safe, and wrapping it in `mergeProps` changes nothing at runtime.

The associated tracking issue #6434 is being closed for the same reason.

PR #6585, opened 2026-04-11
fix(input): merge user props with context props in InputElement and InputAddon

### LINKED ISSUE #6437: `InputElement` and `InputAddon` overwrite user props via `{ ...varProps, ...rest }`
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Input`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/input/input-element.tsx:32`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/input/input-element.tsx#L32) — `InputElement`
- [`packages/react/src/components/input/input-addon.tsx:32`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/input/input-addon.tsx#L32) — `InputAddon`

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6437

## AI used

## Description

Replaces unsafe object spread with `mergeProps` in `InputElement` and `InputAddon` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.

Follows the reference pattern established in `sidebar.tsx`.

## Current behavior (updates)

Passing `className` / `style` / `onClick` to `InputElement` or `InputAddon` would replace context-side values instead of merging with them.

## New behavior

User-provided props are now merged with context props, preserving internal className, styles, refs, and handlers. Added regression tests.

No

## Additional Information

### DIFF
--- a/.changeset/input-element-addon-merge-props.md
+++ b/.changeset/input-element-addon-merge-props.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Use `mergeProps` in `InputElement` and `InputAddon` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.
--- a/packages/react/src/components/input/input-addon.tsx
+++ b/packages/react/src/components/input/input-addon.tsx
@@ -3,7 +3,7 @@
 import type { HTMLStyledProps, ThemeProps } from "../../core"
 import type { InputProps } from "./input"
 import type { InputAddonStyle } from "./input-addon.style"
-import { createComponent } from "../../core"
+import { createComponent, mergeProps } from "../../core"
 import { inputAddonStyle } from "./input-addon.style"
 import { useInputBorder } from "./use-input-border"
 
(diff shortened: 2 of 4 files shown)


hirotomoyamada:
False positive — closing without merging.

`InputElement` and `InputAddon` use `useInputBorder` to produce `varProps`, which only contains CSS custom properties (`--error-border-color`, `--focus-border-color`). These are NOT mergeable keys (`className`, `style`, `css`, `ref`, `on*`), so the original `{ ...varProps, ...rest }` cannot clobber any user-provided handler/style/class — there is nothing in `varProps` to clobber. `mergeProps(varProps, rest)()` produces the same result.

The user-provided `className`/`style`/`onClick` flow through `rest` (which is the user props captured by the context mapper) and are returned untouched, then `withContext` merges them with the slot className via the existing infrastructure. There is no real bug at this site.

PR #6592, opened 2026-04-11
refactor(indicator): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6420: refactor(indicator): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `Indicator` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `Indicator` overflow `<styled.span>` at [indicator.tsx:118](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/indicator/indicator.tsx#L118) — rendered inside `computedLabel` `useMemo` when `label > overflowCount`, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6420

## AI used

## Description

Propagate `suppressHydrationWarning` to the hidden overflow `<span>` rendered inside `Indicator` when the numeric label exceeds `overflowCount`. Without this, users who passed `suppressHydrationWarning` on the `Indicator` root would still see hydration warnings originating from the overflow span.

## Current behavior (updates)

The overflow `<styled.span>` inside `Indicator`'s `computedLabel` did not receive any props from the root, so `suppressHydrationWarning` set on the root was never forwarded to it.

## New behavior

`Indicator` now destructures `suppressHydrationWarning` from its own props, explicitly forwards it to both the root `<styled.div>` and the hidden overflow `<styled.span>`, and re-includes it in the `useMemo` dependency array. Tests cover both the positive and negative propagation cases.

No.

## Additional Information

Follows the same pattern established for `Accordion` in #6575.

### DIFF
--- a/.changeset/indicator-suppress-hydration-hidden.md
+++ b/.changeset/indicator-suppress-hydration-hidden.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Propagate `suppressHydrationWarning` to the hidden overflow span in `Indicator`.
--- a/packages/react/src/components/indicator/indicator.test.tsx
+++ b/packages/react/src/components/indicator/indicator.test.tsx
@@ -1,4 +1,4 @@
-import { a11y, render, screen } from "#test"
+import { a11y, hasSuppressHydrationWarning, render, screen } from "#test"
 import { Indicator } from "."
 
 describe("<Indicator />", () => {
@@ -221,5 +221,40 @@ describe("<Indicator />", () => {
     )
     const label = screen.getByTestId("indicator")
     expect(label).toHaveClass("custom-label")
+  })
+
+  test("propagates `suppressHydrationWarning` to the overflow span when `label > overflowCount`", () => {
+    render(
+      <Indicator
+        label={100}
+        overflowCount={99}
+        suppressHydrationWarning
+        labelProps={{ "data-testid": "indicator" }}
+      >
+        <div />
+      </Indicator>,
+    )
+
+    const label = screen.getByTestId("indicator")
+    const overflowSpan = label.querySelector("span")
+    expect(overflowSpan).toHaveTextContent("+")
+    expect(hasSuppressHydrationWarning(overflowSpan)).toBeTruthy()
+  })
+
+  test("does not set `suppressHydrationWarning` on the overflow span when omitted from the root", () => {
+    render(
+      <Indicator
+        label={100}
+        overflowCount={99}
+        labelProps={{ "data-testid": "indicator" }}
+      >
+        <div />
+      </Indicator>,
+    )
+
+    const label = screen.getByTestId("indicator")
+    const overflowSpan = label.querySelector("span")
+    expect(overflowSpan).toHaveTextContent("+")
+    expect(hasSuppressHydrationWarning(overflowSpan)).toBeFalsy()
   })
 })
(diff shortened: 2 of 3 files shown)


claude-for-yamada-ui (changes requested):
This PR is an instance of **anti-pattern #2** from [review-anti-patterns.md](https://github.com/yamada-ui/yamada-ui/blob/main/.agents/knowledge/review-anti-patterns.md#2-suppresshydrationwarning-propagated-to-elements-without-dynamic-content) — propagating `suppressHydrationWarning` to elements that have no dynamic content.

### 1. The overflow `<styled.span>` does not need `suppressHydrationWarning`

The span renders a static `"+"` character. It has:
- No dynamic attributes (no `id`, no `aria-*` injected by a hook, nothing computed).
- No dynamic text content — it is always the literal string `"+"`.

`suppressHydrationWarning` only suppresses mismatches in an element's own attributes and its direct text content. Since nothing about this span can differ between SSR and hydration, adding the prop has no effect. It is dead code.

### 2. The root `<styled.div>` change is a no-op

Before this PR, `suppressHydrationWarning` flowed through `...rest` onto the root `<styled.div>`. The PR destructures it out of rest and then re-adds it explicitly — the net result on the rendered output is identical. This is unnecessary churn.

### 3. The `useMemo` dependency addition is unnecessary

Adding `suppressHydrationWarning` to the `useMemo` dependency array means the memoized label fragment is recomputed whenever `suppressHydrationWarning` changes. Since the prop is typically a static boolean (or absent), this is harmless but pointless — it exists only to support the unnecessary span prop from point 1.

### Recommendation

All three changes (span prop, root prop extraction, memo dep) are unnecessary. The PR should be closed rather than modified.

codex-for-yamada-ui (commented):
### Request changes

_(Posted as comment because GitHub API rejected request-changes review)_

1. **Unnecessary / ineffective propagation target**
   - File: `packages/react/src/components/indicator/indicator.tsx` (lines 118, 122)
   - `suppressHydrationWarning` is being forwarded to the overflow `+` `<styled.span>`, but this element has static direct text (`"+"`) and no dynamic attrs in this code path. Per React behavior, this prop only affects that element's own attrs/direct text, so this hunk does not earn its place.
   - **Fix suggestion:** remove `suppressHydrationWarning` from the overflow `<styled.span>` and remove it from the `useMemo` dependency array. If suppression is needed, apply it to the element that actually owns dynamic direct content/attrs in this branch.

2. **Tests lock in dead behavior instead of validating meaningful behavior**
   - File: `packages/react/src/components/indicator/indicator.test.tsx` (lines 1, 225, 243)
   - The new tests/assertions only verify propagation to the overflow `span`, which is the unnecessary target above.
   - **Fix suggestion:** remove these span-propagation tests (and `hasSuppressHydrationWarning` import) and replace with tests that assert suppression on an element where mismatches can actually occur.

hirotomoyamada:
Closing this PR because both reviewers independently determined it is unnecessary.

The `<styled.span>` that renders the static `"+"` character has no dynamic attributes and no dynamic text content. `suppressHydrationWarning` only suppresses mismatches in an element's own attributes and its direct text content — since nothing about this span can differ between SSR and hydration, adding the prop has no effect. This is anti-pattern #2 from our review knowledge base.

The root `<styled.div>` change is also a no-op: `suppressHydrationWarning` was already reaching the root via `...rest` before this PR destructured it out and re-added it explicitly.

PR #6593, opened 2026-04-11
refactor(native-accordion): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6421: refactor(native-accordion): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `NativeAccordion` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `NativeAccordionPanel` inner `<styled.div>` at [native-accordion.tsx:215](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/native-accordion/native-accordion.tsx#L215) — inner wrapper around children, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6421

## AI used

## Description

Propagate `suppressHydrationWarning` from `NativeAccordion.Root` to the hidden inner `<div>`
wrapper rendered inside `NativeAccordionPanel`, so users can suppress hydration mismatch
warnings on every DOM node rendered by the component (not only the outer panel wrapper).

This follows the same canonical pattern that was already applied to `Accordion` in PR #6575:

1. Destructure `suppressHydrationWarning` from the slot component's own props. The
   `createSlotComponent` / `withContext` infrastructure merges the root's
   `suppressHydrationWarning` into each slot component's props, so destructuring from props
   is sufficient (no `useHydrationContext` is needed).
2. Spread it on the hidden inner `<styled.div>` wrapper.
3. Re-include it on the outer/visible element via `getPanelProps({ suppressHydrationWarning, ...rest })`
   so that direct usage (`<NativeAccordion.Panel suppressHydrationWarning>`) still reaches
   the visible element.

## Current behavior (updates)

Setting `suppressHydrationWarning` on `NativeAccordion.Root` (or on
`NativeAccordion.Panel` directly) did not propagate to the hidden inner `<div>` wrapper
inside `NativeAccordionPanel`. This caused React hydration mismatch warnings to still be
emitted for that inner wrapper even when the user had explicitly opted in to
`suppressHydrationWarning`.

## New behavior

`NativeAccordionPanel` now destructures `suppressHydrationWarning` from its props and
applies it to both the outer panel `<div>` (via `getPanelProps`) and the hidden inner
`<div>` wrapper, so hydration warnings are suppressed on every DOM node rendered by the
component.

Two new tests verify:

- `suppressHydrationWarning` on `NativeAccordion.Root` propagates to the inner `<div>`.
- The inner `<div>` does not receive `suppressHydrationWarning` when the root omits it.

No.

## Additional Information

Reference: PR #6575 applied the same pattern to `Accordion`.

### DIFF
--- a/.changeset/native-accordion-suppress-hydration-hidden.md
+++ b/.changeset/native-accordion-suppress-hydration-hidden.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Propagate `suppressHydrationWarning` to the hidden inner panel `<div>` in `NativeAccordion`.
--- a/packages/react/src/components/native-accordion/native-accordion.test.tsx
+++ b/packages/react/src/components/native-accordion/native-accordion.test.tsx
@@ -1,4 +1,4 @@
-import { a11y, render, screen } from "#test"
+import { a11y, hasSuppressHydrationWarning, render, screen } from "#test"
 import { BoxIcon } from "../icon"
 import { NativeAccordion } from "./"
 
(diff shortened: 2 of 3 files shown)


claude-for-yamada-ui (changes requested):
This PR falls into **anti-pattern #2** from the review knowledge base: propagating `suppressHydrationWarning` to elements without dynamic content.

### Inner `<styled.div>` (line 215 in the diff)

The inner `<styled.div>` that wraps `children` is a completely static element — no dynamic attributes, no hook-injected IDs, no `aria-*`, no `hidden`. `suppressHydrationWarning` only suppresses mismatches in the element's own attributes and its direct text content; it does not cascade to children. Adding it here has no effect.

### Outer `<styled.div>` — the explicit destructuring is a no-op

`getPanelProps` in `NativeAccordion` is defined as:

```ts
const getPanelProps: PropGetter = useCallback((props = {}) => props, [])
```

It is an identity function — it injects **no dynamic attributes** (unlike the regular `Accordion`'s `getPanelProps` which injects `id`, `aria-labelledby`, `hidden`, etc.). Before this PR, `suppressHydrationWarning` already flowed through `rest` into `getPanelProps(rest)`. Explicitly destructuring it and passing `getPanelProps({ suppressHydrationWarning, ...rest })` produces the same result — it is a refactor with no behavioral change.

### Conclusion

Neither change has a meaningful effect:
- The outer div change is a no-op refactor (prop was already in `rest`).
- The inner div has no dynamic attributes, so `suppressHydrationWarning` does nothing there.

This is exactly the scenario described in [review-anti-patterns.md #2](https://github.com/yamada-ui/yamada-ui/blob/main/.agents/knowledge/review-anti-patterns.md): mechanical propagation of `suppressHydrationWarning` to static wrapper elements where it has no effect.

**Recommendation**: Close this PR. The inner `<div>` does not need `suppressHydrationWarning` because it has no dynamic attributes that could mismatch during hydration.

codex-for-yamada-ui (approved):

PR #6594, opened 2026-04-11
refactor(rating): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6423: refactor(rating): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `Rating` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `RatingItem` `<styled.input>` at [rating.tsx:249](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/rating/rating.tsx#L249) — sibling of the label wrapper, only receives `getInputProps(inputProps)` from the hook, not the root `suppressHydrationWarning`.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6423

## AI used

- [x] ~I did not use AI to create this PR.~

## Description

Propagates `suppressHydrationWarning` from the `Rating` root to the hidden `<input>` element inside `RatingItem`, following the canonical pattern introduced for `Accordion` in #6575.

## Current behavior (updates)

`Rating` accepts `suppressHydrationWarning` at the root, but the hidden `<styled.input>` rendered inside `RatingItem` only received props from `getInputProps(inputProps)` and never the root's `suppressHydrationWarning`. This caused React hydration warnings when `Rating` was rendered on the server and its value was resolved on the client.

## New behavior

`RatingItem` now destructures `suppressHydrationWarning` from its own props (which `createSlotComponent` already merges from the root) and forwards it explicitly to:

- The hidden `<styled.input>` via `getInputProps({ suppressHydrationWarning, ...inputProps })`
- The outer `<styled.label>` via `getLabelProps({ suppressHydrationWarning, ...itemProps })`

Two tests are added to verify the prop is set when the root passes `suppressHydrationWarning` and remains unset otherwise.

No.

## Additional Information

- Uses the shared `hasSuppressHydrationWarning` test helper from `#test`.
- No `useHydrationContext` is introduced; the existing slot-component propagation is sufficient.


claude-for-yamada-ui (approved):

codex-for-yamada-ui (changes requested):
### Findings

1. **Unnecessary hunk (Role 1: Necessity)**
   - File: `packages/react/src/components/rating/rating.tsx`
   - The PR's stated problem is propagation to the hidden `<input>`, but this change also forwards `suppressHydrationWarning` into `getLabelProps(...)` for `<styled.label>`.
   - This label forwarding does not appear required to fix #6423, increases diff surface, and is not covered by the new tests (tests only assert radio inputs).
   - **Concrete fix**: remove `suppressHydrationWarning` from the `getLabelProps` call and keep propagation only in `getInputProps`:
     - keep: `getInputProps({ suppressHydrationWarning, ...runIfFn(inputProps, groupValue) })`
     - revert label call to: `getLabelProps(runIfFn(itemProps, groupValue))`

codex-for-yamada-ui (approved):
Dismissing my previous concern after deeper analysis. `getLabelProps` returns `htmlFor: id` where `id` is derived from `useId()` — this is a dynamic attribute that can differ between SSR and hydration, so `suppressHydrationWarning` on the `<styled.label>` is justified. The propagation to both elements earns its place. LGTM.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

Issue #6412, opened 2026-04-10
Identify components that contain elements hidden within them


hirotomoyamada:
Identify components that contain elements hidden within them (elements that users cannot access from the root element).

This means that developers cannot set `suppressHydrationWarning` on specific elements.

Furthermore, even though `suppressHydrationWarning` now propagates to each slot via [this issue](https://github.com/yamada-ui/yamada-ui/issues/6411) not only for components created with `createComponent` but also for those created with `createSlotComponent`, this issue does not cover cases where elements within a slot do not reference the `createSlotComponent` context that is being hidden within the slot.

For now, we will link the components that fall under the above categories to this issue and address them one by one.

PR #6595, opened 2026-04-11
refactor(picture): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6422: refactor(picture): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `Picture` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `Picture` `sourceElements` `<styled.source>` at [picture.tsx](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/picture/picture.tsx) — generated from the `sources` prop, only spreads per-source rest (excluding `maxW` / `minW`), not the root `suppressHydrationWarning`.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6422

## AI used

## Description

Propagates `suppressHydrationWarning` from the `Picture` root down to the hidden `<source>` elements generated from the `sources` prop, mirroring the pattern applied to `Accordion` in #6575.

## Current behavior (updates)

When `suppressHydrationWarning` is set on `<Picture>`, the flag is not forwarded to the `<source>` elements rendered from the `sources` prop. Only per-source rest props (excluding `maxW`/`minW`) are spread, so React still warns about hydration mismatches on those hidden elements.

## New behavior

`Picture` now destructures `suppressHydrationWarning` from its own props and explicitly forwards it to:

- the visible `<picture>` root,
- each hidden `<source>` rendered inside the `useMemo`-generated `sourceElements` (also added to the dep array),
- the inner `<Image>` fallback.

Direct usage of `<Picture suppressHydrationWarning>` continues to work as before.

Tests added in `picture.test.tsx` using the shared `hasSuppressHydrationWarning` helper:

- Propagates `suppressHydrationWarning` to each `<source>` when the root sets it.
- Does not set `suppressHydrationWarning` on `<source>` elements when the root omits it.

No.

## Additional Information

Quality checks run locally against `packages/react`:

- `pnpm react lint`
- `NODE_OPTIONS=--max-old-space-size=8192 pnpm react typecheck`
- `pnpm react test:jsdom --run src/components/picture/` (12 passed)


claude-for-yamada-ui (changes requested):
## Review — `suppressHydrationWarning` propagation in Picture

Applying anti-pattern #2 from review knowledge: for each element receiving `suppressHydrationWarning`, does the prop have a meaningful effect on *that specific element*?

### 1. `<styled.source>` elements (sourceElements) — Justified

These are internal elements generated from the `sources` prop that users cannot reach directly. The issue (#6422) specifically identifies these as the affected elements. Since they are Emotion-wrapped and receive computed `media` attributes, propagating `suppressHydrationWarning` here is consistent with the design intent of #6412.

### 2. `<styled.picture>` root — Unnecessary

The `<styled.picture>` element already accepts `pictureProps`, which means the user can already pass `suppressHydrationWarning` via `pictureProps={{ suppressHydrationWarning: true }}`. This element is not "hidden" or "unreachable" — it is explicitly exposed. Adding the prop here is out of scope of #6422 (which only mentions `<styled.source>`).

**Suggestion**: Remove the `suppressHydrationWarning` addition from `<styled.picture>`. The user already has access via `pictureProps`.

### 3. `<Image>` fallback — Unnecessary / redundant restructuring

Before this PR, `suppressHydrationWarning` was **not** destructured from props, so it naturally flowed through `...rest` to `<Image {...rest} />`. The PR destructures it out and then explicitly passes it back — the net effect is identical, but the code is more complex for no reason. `<Image>` is also not identified as an affected element in #6422.

**Suggestion**: Do not destructure `suppressHydrationWarning` from the top-level props. Instead, only extract it where needed (inside the `sourceElements` memo) or pass it from `rest` into the source elements. This avoids unnecessary changes to `<styled.picture>` and `<Image>`.

### Concrete fix

Keep `suppressHydrationWarning` in `...rest` (do not destructure it at the top level). Only forward it to the `<source>` elements:

```tsx
const Picture: FC<PictureProps> = ({
  children,
  enableSorting = true,
  sources: sourcesProp = [],
  suppressHydrationWarning,
  pictureProps,
  ...rest
}) => {
```

should remain as-is in the original (without the `suppressHydrationWarning` destructuring), and then extract it inside `sourceElements`:

```tsx
const sourceElements = useMemo(
  () =>
    sources.map(({ maxW: _maxW, minW: _minW, ...rest }, index) => (
      <styled.source
        key={index}
        suppressHydrationWarning={restFromPicture.suppressHydrationWarning}
        {...rest}
      />
    )),
  [sources, restFromPicture.suppressHydrationWarning],
)
```

Or more practically: keep the destructuring but remove the additions to `<styled.picture>` and `<Image>`, and put `suppressHydrationWarning` back into the `rest` spread for `<Image>`:

```tsx
<Image suppressHydrationWarning={suppressHydrationWarning} {...rest} />
```

This line is technically fine but redundant — consider just not destructuring it at all and accessing `rest.suppressHydrationWarning` for the source elements.

### Summary

Only the `<styled.source>` propagation earns its place. The `<styled.picture>` and `<Image>` changes are unnecessary scope creep beyond what #6422 asks for.

codex-for-yamada-ui (changes requested):
### Findings

1. **Necessity / Scope**  
   **File:** `packages/react/src/components/picture/picture.tsx` (around `return` block, lines ~140-148 in this patch)  
   The additions forwarding `suppressHydrationWarning` to `<styled.picture>` and `<Image>` do not appear necessary for issue #6422 (which is specifically about hidden `<source>` elements from `sources`).  
   - If those two hunks are removed, the reported bug (root flag not reaching hidden generated `<source>`) can still be fixed.  
   - This expands behavior beyond the issue scope and makes the patch less minimal.  

   **Suggested fix:** limit the change to generated hidden `<source>` elements only.  
   - Keep `<styled.picture {...pictureProps}>` unchanged.  
   - Keep `<Image {...rest} />` unchanged.  
   - Use the root prop only for generated sources (for example, read `rest.suppressHydrationWarning` when creating `sourceElements`, or destructure only if needed for that purpose).

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6598, opened 2026-04-11
refactor(sidebar): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6425: refactor(sidebar): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `Sidebar` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `SidebarHandle` inner `<styled.button>` at [sidebar.tsx:740](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/sidebar/sidebar.tsx#L740) — only sets `tabIndex={-1}`, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6425

## AI used

## Description

Propagates `suppressHydrationWarning` from `SidebarHandle` to its hidden inner
`<button>` so the hidden element honors the root's hydration behavior, matching
the canonical pattern used in `Accordion` (#6575).

## Current behavior (updates)

`SidebarHandle` rendered a hidden inner `<styled.button tabIndex={-1} />` that
did not receive `suppressHydrationWarning`, so the hidden element still emitted
hydration warnings even when the root set `suppressHydrationWarning`.

## New behavior

`SidebarHandle` destructures `suppressHydrationWarning` from its own props and
spreads it on both the outer `SidebarTrigger` (via `mergeProps`) and the hidden
inner `<styled.button>`. `createSlotComponent` already merges the root's
`suppressHydrationWarning` into each slot component's props, so direct usage of
`Sidebar.Handle` also works.

Tests were added to `sidebar.test.tsx` to cover both propagation and the
negative case when the root omits the prop.

No.

## Additional Information

- Lint, typecheck, and `src/components/sidebar/` tests all pass locally.

### DIFF
--- a/.changeset/sidebar-suppress-hydration-hidden.md
+++ b/.changeset/sidebar-suppress-hydration-hidden.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Propagate `suppressHydrationWarning` to the hidden handle `<button>` in `Sidebar`.
--- a/packages/react/src/components/sidebar/sidebar.test.tsx
+++ b/packages/react/src/components/sidebar/sidebar.test.tsx
@@ -1,4 +1,11 @@
-import { a11y, fireEvent, render, screen, waitFor } from "#test"
+import {
+  a11y,
+  fireEvent,
+  hasSuppressHydrationWarning,
+  render,
+  screen,
+  waitFor,
+} from "#test"
 import { useRef } from "react"
 import { Sidebar } from "."
 import { Button } from "../button"
@@ -685,5 +692,33 @@ describe("<Sidebar />", () => {
     const link = screen.getByRole("link", { name: "1-1" })
 
     expect(link).toHaveAttribute("data-selected")
+  })
+
+  test("propagates `suppressHydrationWarning` to the `SidebarHandle` `<button>`", () => {
+    const { container } = render(
+      <Sidebar.Root suppressHydrationWarning>
+        <Sidebar.SidePanel>
+          <Sidebar.Handle />
+        </Sidebar.SidePanel>
+      </Sidebar.Root>,
+    )
+
+    const handle = container.querySelector(".ui-sidebar__handle")
+    expect(handle?.tagName).toBe("BUTTON")
+    expect(hasSuppressHydrationWarning(handle)).toBeTruthy()
+  })
+
+  test("does not set `suppressHydrationWarning` on the `SidebarHandle` `<button>` when the root omits it", () => {
+    const { container } = render(
+      <Sidebar.Root>
+        <Sidebar.SidePanel>
+          <Sidebar.Handle />
+        </Sidebar.SidePanel>
+      </Sidebar.Root>,
+    )
+
+    const handle = container.querySelector(".ui-sidebar__handle")
+    expect(handle?.tagName).toBe("BUTTON")
+    expect(hasSuppressHydrationWarning(handle)).toBeFalsy()
   })
 })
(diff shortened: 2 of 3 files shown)


claude-for-yamada-ui (changes requested):
## Review — request changes

### Role 1 (Necessity): the inner `<styled.button>` does not need `suppressHydrationWarning`

This PR falls into **anti-pattern #2** from `review-anti-patterns.md` — propagating `suppressHydrationWarning` to an element that has no dynamic content.

The inner element:

```tsx
<styled.button
  suppressHydrationWarning={suppressHydrationWarning}
  tabIndex={-1}
/>
```

is completely static. It has only `tabIndex={-1}`, no dynamic attributes (no `id`, no `aria-*` from hooks, no text content). `suppressHydrationWarning` only suppresses mismatches in an element's own attributes and direct text content. Since this element has nothing that can differ between SSR and hydration, the prop has no effect and should not be added.

### Role 1 (Necessity): the explicit destructuring for the outer `SidebarTrigger` is redundant

`SidebarHandle` is created via `withContext`, which uses the internal `HydrationContext`. Looking at `create-component.tsx`, every slot component automatically receives `hydrationProps` (including `suppressHydrationWarning`) from the root via `useHydrationContext()`. These are spread into the `props` parameter of the function body **before** the function runs.

The original code:

```tsx
<SidebarTrigger {...mergeProps(handleProps, props)()}>
```

already carries `suppressHydrationWarning` from the hydration context as part of `props`. The PR's explicit destructuring:

```tsx
({ suppressHydrationWarning, ...rest }) => {
  ...
  <SidebarTrigger {...mergeProps(handleProps, { suppressHydrationWarning, ...rest })()}>
```

reconstructs the same object that `props` already was. This is unnecessary.

### Role 5 (Test)

The tests assert `hasSuppressHydrationWarning` on `.ui-sidebar__handle`, which is the outer trigger button — but this propagation already works through the hydration context without any code changes. The tests pass on the base branch as well.

### Summary

Both the source change and its tests are unnecessary. The `withContext` / `HydrationContext` mechanism already propagates `suppressHydrationWarning` to `SidebarHandle` and its outer `SidebarTrigger`. The inner `<styled.button>` is static and does not benefit from the prop (anti-pattern #2). This PR should be closed or revised to address a genuine hydration-warning site.

codex-for-yamada-ui (changes requested):
### Findings (request changes)

1. **Test may target the wrong element, so it can pass without validating the hidden button behavior**
   - File: `packages/react/src/components/sidebar/sidebar.test.tsx` (line 705)
   - The new tests query `.ui-sidebar__handle` and then check `hasSuppressHydrationWarning`. This selector does not guarantee you are asserting against the hidden inner handle button (`tabIndex={-1}`), which is the element mentioned in the PR description.
   - **Fix suggestion:** Select/assert the hidden button explicitly (for example, `.ui-sidebar__handle[tabindex="-1"]`) and assert that element's hydration-warning state. Keep the positive and negative cases.

2. **Redundant prop propagation in `SidebarHandle`**
   - File: `packages/react/src/components/sidebar/sidebar.tsx` (line 736)
   - `suppressHydrationWarning` is propagated both through `mergeProps(handleProps, { suppressHydrationWarning, ...rest })()` and directly on `<styled.button suppressHydrationWarning={...} />`. That duplicates responsibility in the same hunk.
   - **Fix suggestion:** Keep a single propagation path and document intent via a focused test. For this specific hidden-button fix, applying it directly to the hidden `<styled.button>` is the clearest path.

PR #6599, opened 2026-04-11
refactor(segmented-control): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6424: refactor(segmented-control): propagate suppressHydrationWarning to hidden elements
### Description

Some internal elements inside the `SegmentedControl` component are not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to them.

Affected elements:

- `SegmentedControlItem` `<styled.input>` at [segmented-control.tsx:150](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/segmented-control/segmented-control.tsx#L150) — sibling of the label wrapper, only receives `getInputProps()` from the hook.
- `SegmentedControlItem` `<styled.span>` at [segmented-control.tsx:152](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/segmented-control/segmented-control.tsx#L152) — children wrapper, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6424

## AI used

## Description

Propagate `suppressHydrationWarning` from `SegmentedControl.Item` to the
hidden `<input>` and `<span>` elements it renders, so that hydration warnings
are suppressed across every DOM node produced by the component.

## Current behavior (updates)

When `suppressHydrationWarning` is set on `SegmentedControl.Root` (or
`SegmentedControl.Item`), it reaches the outer `<label>` via
`createSlotComponent` / `withContext` prop merging, but the hidden
`<input>` sibling and the `<span>` children wrapper do not receive it,
so mismatched content inside those nodes can still produce hydration
warnings.

## New behavior

`SegmentedControlItem` now destructures `suppressHydrationWarning` from
its props and forwards it explicitly to:

- the hidden `<input>` via `getInputProps({ suppressHydrationWarning })`
- the children `<span>` wrapper via a direct prop
- the visible `<label>` via `getLabelProps({ suppressHydrationWarning })`
  so direct per-item usage keeps working

Two tests were added to verify the propagation (and its absence when the
root omits the prop).

No.

## Additional Information

- Follows the canonical pattern established in #6575 (accordion).
- Quality checks (`pnpm react lint`, `pnpm react typecheck`,
  `pnpm react test:jsdom --run src/components/segmented-control/`) all
  pass locally.


claude-for-yamada-ui (changes requested):
## Review: request changes

### Role 1 — Necessity (anti-pattern #2: `suppressHydrationWarning` on static wrapper)

The `<label>` and `<input>` sites are justified — both receive dynamic attributes from `getLabelProps` / `getInputProps` (`aria-disabled`, `data-checked`, `checked`, `tabIndex`, etc.) that can differ between SSR and hydration.

However, the `<styled.span>` wrapper at line 153 has **no dynamic attributes** — no `id`, no `aria-*`, no `data-*` computed from hooks. It is a plain wrapper whose only content is `{children}` (the caller's label text). `suppressHydrationWarning` suppresses mismatches in the element's own attributes and its direct text content only; it does **not** propagate to children (React docs are explicit). Since the `<span>` has nothing dynamic about it, the prop has no effect.

This is exactly anti-pattern #2 from our review knowledge base (the accordion `<styled.p>` case). The fix: remove `suppressHydrationWarning` from the `<styled.span>`, and update the corresponding test assertion that checks the `<span>`.

```diff
-      <styled.span suppressHydrationWarning={suppressHydrationWarning}>
+      <styled.span>
```

### Role 2–9

No issues found with correctness, accessibility, type safety, API surface, style/theme, scope, or consistency (aside from the necessity issue above).

### Summary

The `<label>` and `<input>` propagation is correct and justified. Please remove `suppressHydrationWarning` from the `<styled.span>` and adjust the test accordingly.

codex-for-yamada-ui (changes requested):
### Findings

1. **Test gap for the item-level propagation path introduced by this diff (request changes)**
   File: `packages/react/src/components/segmented-control/segmented-control.test.tsx`:90
   The new tests only verify root-level propagation (`<SegmentedControl.Root suppressHydrationWarning />` with `items`), but this PR also changes item-level handling in `SegmentedControlItem` (`suppressHydrationWarning` is destructured and manually threaded into `getLabelProps`/`getInputProps`). That item-level path is currently untested.
   **Concrete fix:** add a test rendering explicit children (`<SegmentedControl.Item ...>`) and assert that `suppressHydrationWarning` on an item reaches that item's hidden `<input>` and `<span>` (and a negative case where it is omitted). This validates the new threading logic and prevents silent regressions.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6600, opened 2026-04-11
refactor(steps): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6427: refactor(steps): propagate suppressHydrationWarning to hidden elements
### Description

An internal element inside the `Steps` component is not reachable via the root element's props, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `StepsList` items-mode inner `<styled.div>` at [steps.tsx:228](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/steps/steps.tsx#L228) — wraps `StepsTitle` / `StepsDescription` when the `items` prop is used, no props spread.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6427

## AI used

## Description

Propagate `suppressHydrationWarning` to the hidden inner wrapper in the `Steps` component when using the `items` prop, so hydration mismatches inside `StepsTitle` / `StepsDescription` can be suppressed from the root.

## Current behavior (updates)

When `Steps.Root` receives `suppressHydrationWarning`, the inner `<div>` wrapper inside `StepsList` items-mode (around `StepsTitle` / `StepsDescription`) did not receive the flag, so hydration warnings could still be surfaced there.

## New behavior

`StepsList` now destructures `suppressHydrationWarning` from its own props (which `createSlotComponent` already merges from the root via `withContext`) and spreads it on the hidden wrapper `<div>`. It is also re-included via `getListProps({ suppressHydrationWarning, ...rest })` so direct usage on `Steps.List` still works.

No.

## Additional Information

- Tests added in `packages/react/src/components/steps/steps.test.tsx` verify propagation (and non-propagation when omitted) in items-mode.
- Follows the same pattern as PR #6575 (accordion).

### DIFF
--- a/.changeset/steps-suppress-hydration-hidden.md
+++ b/.changeset/steps-suppress-hydration-hidden.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Propagate `suppressHydrationWarning` to the hidden items-mode wrapper in `Steps`.
--- a/packages/react/src/components/steps/steps.test.tsx
+++ b/packages/react/src/components/steps/steps.test.tsx
@@ -1,5 +1,5 @@
 import type { FC } from "react"
-import { a11y, render, screen } from "#test"
+import { a11y, hasSuppressHydrationWarning, render, screen } from "#test"
 import { useMemo } from "react"
 import { Steps } from "."
 import { ButtonGroup } from "../button"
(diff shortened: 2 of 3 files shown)


claude-for-yamada-ui (changes requested):
## Review — request changes

### Anti-pattern #2: `suppressHydrationWarning` on a static element

**File:** `packages/react/src/components/steps/steps.tsx`

The inner `<styled.div>` wrapper (around `StepsTitle` / `StepsDescription`) is a **static layout element** with no dynamic attributes and no direct text content. Adding `suppressHydrationWarning` to it has no effect.

Per React docs, `suppressHydrationWarning` only suppresses mismatches in **the element's own attributes and its direct text content** — it does not cascade to children. This `<styled.div>` receives no props from any hook (no `id`, no `aria-*`, no `hidden`, no dynamic data attributes). Its children are React elements (`StepsTitle`, `StepsDescription`), not text nodes. Therefore `suppressHydrationWarning` on this element is dead code.

This is the same pattern documented in [review-anti-patterns.md #2](https://github.com/yamada-ui/yamada-ui/blob/main/.agents/knowledge/review-anti-patterns.md), which was identified after PR #6575 (accordion). The PR description explicitly says "Follows the same pattern as PR #6575 (accordion)" — but that pattern was retroactively flagged as an anti-pattern for exactly this reason.

**What to keep:**
- Propagating `suppressHydrationWarning` to the `<styled.ol>` via `getListProps({ suppressHydrationWarning, ...rest })` is acceptable — the `<ol>` is the main container rendered by `StepsList` and receives `data-orientation` from `getListProps`.

**What to remove:**
- `suppressHydrationWarning={suppressHydrationWarning}` on the inner `<styled.div>` (line 228 in the diff).
- The corresponding `suppressHydrationWarning` destructure can stay since it's still needed for the `<ol>` propagation.
- The test "propagates `suppressHydrationWarning` to the items-mode inner wrapper `<div>`" should be removed or rewritten to test the `<ol>` instead, since the wrapper `<div>` should not receive the prop.
- The `suppressHydrationWarning` dependency in the `useMemo` array should be removed if the only use inside the memo was on the `<styled.div>`.

### Summary

The `<styled.ol>` propagation earns its place. The inner `<styled.div>` propagation does not — it is a no-op on a static element.

codex-for-yamada-ui (dismissed):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6601, opened 2026-04-11
refactor(slider): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6426: refactor(slider): propagate suppressHydrationWarning to hidden elements
### Description

An internal helper component inside the `Slider` file is not wrapped by the `createSlotComponent` factory, so `suppressHydrationWarning` set on the root does not propagate to it.

Affected elements:

- `SliderInput` FC `<styled.input>` at [slider.tsx:226](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/slider/slider.tsx#L226) — plain `FC` that returns a raw `<styled.input>` (or an array of them when `range` is true) with only `getInputProps()` from the hook; the `HydrationContext` propagated by `withContext` never reaches it.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6426

## AI used

## Description

Forwards `suppressHydrationWarning` from `Slider.Root` to the hidden `<input>` element(s) rendered by `SliderInput`, matching the pattern already used by `Accordion` (#6575).

## Current behavior (updates)

When `Slider.Root` is given `suppressHydrationWarning`, the prop reaches the visible root `<div>` (via `getRootProps` spreading `rest`), but the internal `SliderInput` FC is not a slot component, so the hidden `<input>` does not inherit the hydration setting. SSR-rendered sliders whose value differs between server and client therefore emit hydration mismatch warnings.

## New behavior

- `SliderRoot` now destructures `suppressHydrationWarning` from its own props and re-applies it to both the root element (via `getRootProps({ suppressHydrationWarning })`) and the `SliderInput` child.
- `SliderInput` accepts `suppressHydrationWarning` and forwards it to `getInputProps` so it lands on the hidden `<styled.input>` — in both single and `range` modes.
- Users can still override per-input behavior via `inputProps.suppressHydrationWarning`.

No.

## Additional Information

- Tests added in `packages/react/src/components/slider/slider.test.tsx` cover the single-value, range, and default-omitted cases using `hasSuppressHydrationWarning`.
- Quality checks run locally: `pnpm react lint`, `pnpm react typecheck`, `pnpm react test:jsdom --run src/components/slider/` — all green.


claude-for-yamada-ui (dismissed):

codex-for-yamada-ui (changes requested):
### Request Changes

1. **Missing regression test for the override contract (`inputProps.suppressHydrationWarning`)**
   - **Files:** `packages/react/src/components/slider/slider.tsx` (L182), `packages/react/src/components/slider/slider.test.tsx` (L135)
   - **Why this matters:** The implementation intentionally sets root-level `suppressHydrationWarning` and then allows `inputProps` to override it on hidden input(s) via prop spread order. That override behavior is part of the PR description and is easy to regress in future refactors, but there is no test for it.
   - **Concrete fix suggestion:** Add a test like:
     - render `<Slider.Root suppressHydrationWarning inputProps={{ suppressHydrationWarning: false }} />`
     - assert hidden input does **not** have suppress-hydration-warning while root still does.
     - (Optional but stronger) add the same assertion in range mode for both hidden inputs.

hirotomoyamada:
Addressed @codex-for-yamada-ui's review: added a test that verifies `inputProps.suppressHydrationWarning` can override the root-level prop on the hidden `<input>` (commit 626b474c86).

claude-for-yamada-ui (dismissed):

codex-for-yamada-ui (dismissed):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6606, opened 2026-04-12
refactor(table): propagate suppressHydrationWarning to hidden elements

### LINKED ISSUE #6428: refactor(table): propagate suppressHydrationWarning to hidden elements
### Description

Several internal helper components inside the `Table` file are plain `FC`s returning raw `<styled.*>` elements, so `suppressHydrationWarning` set on the root does not propagate to them.

Affected elements:

- `SortingIcon` `<styled.button>` at [table.tsx:1102](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/table/table.tsx#L1102)
- `ResizableTrigger` `<styled.div>` at [table.tsx:1147](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/table/table.tsx#L1147)
- `TruncatedText` `<styled.span>` at [table.tsx:1175](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/table/table.tsx#L1175)

All three are local FCs without `createSlotComponent` wrapping; the `HydrationContext` provided by the root never reaches them.

### Proposed Solution or API

_No response_

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Part of #6412.

hirotomoyamada:
Closes #6428

## AI used

## Description

Propagated `suppressHydrationWarning` set on the `Table` root to three hidden internal FCs that previously rendered raw `<styled.*>` elements without the root hydration context.

- Added `suppressHydrationWarning` to `PickedNativeTableProps` so the prop is exposed on `TableProps`.
- `SortingIcon` now destructures `suppressHydrationWarning` and spreads it on its `<styled.button>`.
- `ResizableTrigger` now destructures `suppressHydrationWarning` and spreads it on its `<styled.div>`.
- `TruncatedText` now destructures `suppressHydrationWarning` and spreads it on its `<styled.span>`.

No


claude-for-yamada-ui (changes requested):
## Review: request changes

This PR is a direct instance of **anti-pattern #2** documented in `.agents/knowledge/review-anti-patterns.md` — "`suppressHydrationWarning` propagated to elements without dynamic content."

### Analysis per target element

`suppressHydrationWarning` only suppresses mismatches in **the element's own attributes and its direct text content**. It does not cascade. Whether it earns its place depends on whether *that specific element* has anything that can differ between SSR and hydration.

#### 1. `SortingIcon` (`<styled.button>`)
- Attributes are all static literals (`type="button"`, `tabIndex={-1}`, style props, etc.)
- `aria-label` comes from `t(...)` (i18n) which is deterministic for a given locale — no SSR/hydration mismatch
- No `useId()`, no server-generated IDs
- **Verdict: `suppressHydrationWarning` has no effect on this element**

#### 2. `ResizableTrigger` (`<styled.div>`)
- `data-active` and `transform` derive from client state (`resizing`, `offset`) — on initial SSR render these are deterministic (resizing=false, offset is null/0)
- No `useId()`, no server-generated IDs
- **Verdict: `suppressHydrationWarning` has no effect on this element**

#### 3. `TruncatedText` (`<styled.span>`)
- Pure static wrapper with `lineClamp`, `truncated`, `wordBreak` style props
- No dynamic attributes whatsoever
- **Verdict: `suppressHydrationWarning` has no effect on this element**

### Additional concern: redundant destructuring

Since `SortingIconProps extends HTMLStyledProps<"button">` (and similarly for the other two), `suppressHydrationWarning` is already included in the type and would flow through `...rest` into the `{...mergeProps(rest, sortingIconProps)()}` spread. Destructuring it out of `rest` and re-adding it as an explicit prop placed *before* the spread actually changes merge priority — `sortingIconProps` would now override it, whereas before it would have been part of `rest` with equal priority. This is a subtle behavioral change for no benefit.

### Conclusion

None of the three target elements have attributes that can differ between SSR and hydration. There is no `useId()` or equivalent anywhere in the table component. Adding `suppressHydrationWarning` to these elements has zero effect and adds dead code. This is the exact same pattern that was rejected in PR #6575 for Accordion.

If there is a real hydration mismatch in the Table component, the fix should target the specific element(s) where the mismatch actually occurs, not fan out the prop mechanically to internal wrappers.

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6607, opened 2026-04-12
fix(accordion): drop redundant suppressHydrationWarning on static wrappers

hirotomoyamada:
## AI used

## Description

Follow-up to #6575. That PR forwarded `suppressHydrationWarning` from `Accordion.Root` into the static `<h3>` wrapper of `AccordionButton` and the static `<p>` wrapper that `AccordionPanel` renders for string children. React's `suppressHydrationWarning` only suppresses mismatches on the element's own attributes and direct text content, **does not propagate to children**, and is meaningless on a static wrapper that has no dynamic attributes — so those two forwards have no effect at runtime and are dead props.

The legitimate forwarding paths are kept untouched: `getButtonProps({ suppressHydrationWarning, ... })` on `AccordionButton`'s `<button>` and `getPanelProps({ suppressHydrationWarning, ... })` on `AccordionPanel`'s `<div>` both inject dynamic attributes (`id`, `aria-*`, etc.) that genuinely can mismatch between SSR and hydration.

The two tests added by #6575 asserted the dead props on the static wrappers, so they are removed along with the now-unused `hasSuppressHydrationWarning` test helper.

## Current behavior (updates)

`suppressHydrationWarning` is forwarded onto `<styled.h3>` (in `AccordionButton`) and `<styled.p>` (in `AccordionPanel`), where React ignores it because neither element has dynamic attributes or direct dynamic text.

## New behavior

`suppressHydrationWarning` is no longer forwarded to those static wrappers. The prop continues to flow through `getButtonProps` / `getPanelProps` to the elements that actually need it.

No.

### DIFF
--- a/.changeset/fix-accordion-suppress-hydration-redundant.md
+++ b/.changeset/fix-accordion-suppress-hydration-redundant.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Removed redundant `suppressHydrationWarning` forwarding to the static `<h3>` and `<p>` wrapper elements inside `AccordionButton` and `AccordionPanel`. React's `suppressHydrationWarning` does not propagate to children and has no effect on these static wrappers; the legitimate forwarding via `getButtonProps` and `getPanelProps` is preserved.
--- a/packages/react/src/components/accordion/accordion.test.tsx
+++ b/packages/react/src/components/accordion/accordion.test.tsx
@@ -1,4 +1,4 @@
-import { a11y, hasSuppressHydrationWarning, render, screen } from "#test"
+import { a11y, render, screen } from "#test"
 import { noop } from "../../utils"
 import { BoxIcon } from "../icon"
 import { Accordion } from "./"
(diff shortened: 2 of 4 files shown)


claude-for-yamada-ui (dismissed):

codex-for-yamada-ui (changes requested):
### Role 5 — Test Coverage (Issue Found)

**File:** `packages/react/src/components/accordion/accordion.test.tsx` around line 346

Two tests asserting `suppressHydrationWarning` behavior were removed with no replacement. The PR preserves the forwarding path through `getButtonProps` / `getPanelProps`, but no test now verifies that `suppressHydrationWarning` actually reaches the target elements via those paths.

**Concrete fix suggestion:** Add replacement tests that assert:
1. `Accordion.Button`'s rendered `<button>` receives `suppressHydrationWarning` via `getButtonProps`.
2. `Accordion.Panel`'s rendered inner `<div>` receives `suppressHydrationWarning` via `getPanelProps`.

The removed wrapper-level assertions (`<h3>`, `<p>`) can stay removed, but positive coverage for the preserved behavior must be added.

---

### Role 9 — Consistency (Hypothesis)

**File:** `packages/react/test/utils.ts` line 1 — `hasSuppressHydrationWarning` helper removed.

This is consistent with the PR's goal, but if similar assertions are needed elsewhere in future hydration-related tests, a local helper per test file would be preferable over reintroducing a broad shared utility. Not a blocker, but worth noting.

---

**Verdict: Request changes.** The deleted tests are not replaced by equivalent coverage for the preserved forwarding path, which leaves the intended `suppressHydrationWarning` behavior unguarded against future regressions.

claude-for-yamada-ui:
Added two tests that verify `suppressHydrationWarning` reaches the correct elements via `getButtonProps` (on the `<button>`) and `getPanelProps` (on the panel `<div>`). Each test reads the React internal fiber props to confirm the prop is forwarded to the dynamic elements where it has effect.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6610, opened 2026-04-12
fix(textarea): merge user props with context props via mergeProps

### LINKED ISSUE #6447: `Textarea` overwrites user props via `{ rows: 2, ...context, ...props }`
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Textarea`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/textarea/textarea.tsx:39`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/textarea/textarea.tsx#L39) — `Textarea` context mapper

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6447

## AI used

## Description

Use `mergeProps` in `Textarea` so user props are merged with context props instead of being overwritten.

## Current behavior (updates)

`Textarea` merged defaults/context/user props with object spread, so context props could overwrite user-provided `className`, `style`, `css`, `ref`, and `on*` handlers.

## New behavior

`Textarea` now uses the canonical `mergeProps({ rows: 2 }, context, props)()` pattern to merge these props correctly. Added a regression test to verify `className`, `style`, and click handler merging from context and user props.

No

## Additional Information

Added changeset: `.changeset/gentle-canvas-merge.md`.

### DIFF
--- a/.changeset/gentle-canvas-merge.md
+++ b/.changeset/gentle-canvas-merge.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Use mergeProps in Textarea so that user-provided className, style, css, ref, and on* event handlers are merged with context props instead of being silently overwritten.
--- a/packages/react/src/components/textarea/textarea.test.tsx
+++ b/packages/react/src/components/textarea/textarea.test.tsx
@@ -1,4 +1,5 @@
-import { a11y, render, screen } from "#test"
+import { a11y, fireEvent, render, screen } from "#test"
+import { InputPropsContext } from "../input"
 import { Textarea } from "./"
 
 describe("<Textarea />", () => {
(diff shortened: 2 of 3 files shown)


claude-for-yamada-ui (changes requested):
This PR is unnecessary. It matches **anti-pattern #1 (utility over-application)** from our review knowledge base.

**The context object has no fields that `mergeProps` exists to handle.**

`useInputPropsContext()` returns the value set by `InputGroupRoot`, which contains only: `size`, `variant`, `errorBorderColor`, `focusBorderColor`, `disabled`, `invalid`, `readOnly`, `required`, and `data-*` attributes. None of these are `className`, `style`, `ref`, or `on*` event handlers.

The original `{ rows: 2, ...context, ...props }` is correct and sufficient. Plain spread is the right tool when there is nothing to merge — `mergeProps` adds overhead for zero benefit.

The test creates an artificial scenario by injecting `className`, `style`, and `onClick` directly into `InputPropsContext`, but `InputGroupRoot` (the only real provider of this context) never sets those fields. The test proves mergeProps works in general, not that this call site needs it.

Recommend closing this PR.

claude-for-yamada-ui (changes requested):
This PR is unnecessary — it is a textbook instance of anti-pattern #1 (utility over-application) from `review-anti-patterns.md`.

`useInputPropsContext()` is provided by `InputGroupRoot`, which builds the context from `size`, `variant`, `errorBorderColor`, `focusBorderColor`, `disabled`, `invalid`, `readOnly`, `required`, and `data-*` attributes. The context never contains `className`, `style`, `ref`, or event handlers — the exact fields `mergeProps` exists to reconcile.

Because there is nothing to merge, `{ rows: 2, ...context, ...props }` is semantically identical to `mergeProps({ rows: 2 }, context, props)()` but cheaper and easier to read.

The added test manufactures a scenario that cannot occur in production: it injects `className`, `style`, and `onClick` directly into `InputPropsContext`, but the sole real provider (`InputGroupRoot`) never sets those fields.

Recommend closing.

codex-for-yamada-ui (changes requested):
This PR is unnecessary. It applies `mergeProps` where plain object spread is equivalent.

The context from `useInputPropsContext()` (provided by `InputGroupRoot`) contains only scalar theme/field props: `size`, `variant`, `errorBorderColor`, `focusBorderColor`, `disabled`, `invalid`, `readOnly`, `required`, and `data-*` attributes. It never includes `className`, `style`, `ref`, or event handlers — the fields `mergeProps` is designed to reconcile.

`{ rows: 2, ...context, ...props }` already produces the correct result. Wrapping it in `mergeProps` is dead weight (anti-pattern #1 in review-anti-patterns.md).

The regression test is artificial: it puts `className`, `style`, and `onClick` into `InputPropsContext` directly, but `InputGroupRoot` never provides those fields in practice.

Recommend closing.

Issue #6430, opened 2026-04-11
Tracking: unsafe prop spread overrides user-provided className/style/event handlers across components


hirotomoyamada:
## Summary

Across many components in `@yamada-ui/react`, user-provided props are merged with context/getter props via plain object spread (`{ ...contextProps, ...userProps }`) or JSX double spread (`<X {...a} {...b} />`). This silently **overwrites** the context-side values for `className`, `style`, `css`, `ref`, and `on*` event handlers when both sides set them, breaking the user's ability to customize these without losing the component's internal wiring.

## Root cause

The codebase already has a safe merge utility in `packages/react/src/core/components/props.ts`:

    mergeProps(a, b)()  // handles className (cx), style (merge),
                        // css (array), ref (mergeRefs), on* (handlerAll)

Plain spread bypasses all of that and just takes the last value.

## Reference fix

`packages/react/src/components/sidebar/sidebar.tsx` already demonstrates the correct pattern — `itemComponentContext` builds each prop bucket with `mergeProps(componentContext.xxxProps, xxxProps)()` instead of raw spread.

## Scope

29 affected components, tracked as sub-issues below. Fixes must be split per-component (one PR per component) per `CLAUDE.md` "Do not bundle multiple fixes".

## AI used

- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

PR #6611, opened 2026-04-12
fix(file-input): merge user props with context props via mergeProps

### LINKED ISSUE #6446: `FileInput` overwrites user props via JSX double spread and context mapper
@yamada-ui/react

### Package Version
2.2.0

### Description
In `FileInput`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/file-input/file-input.tsx:117`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/file-input/file-input.tsx#L117) — JSX double spread
- [`packages/react/src/components/file-input/file-input.tsx:127`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/file-input/file-input.tsx#L127) — context mapper

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6446

## AI used

## Description

Replaces unsafe double-spread patterns in `FileInput` with `mergeProps(contextProps, userProps)()` from `../../core`, mirroring the pattern established in PR #6578 (`fix(tree)`).

## Current behavior (updates)

`FileInput` used a raw spread (`{...contextProps, ...userProps}`) which silently overwrote user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers from the context with the component's own props, or vice versa.

## New behavior

`mergeProps` is used to properly merge context props with user props so that `className` values are concatenated, `style` objects are merged, and event handlers are chained — none are silently dropped.

Tests were added to `file-input.test.tsx` to verify that `className`, `style`, and `onClick` from both `InputPropsContext` and the user are all applied.

No

## Additional Information

Sub-issue of #6430. Canonical pattern from #6578.


packages/react/src/components/file-input/file-input.tsx
```diff
@@ -114,7 +114,7 @@ export const FileInput = withProvider<"input", FileInputProps>(
           <styled.input {...getInputProps()} />
         </Portal>
 
-        <styled.div {...varProps} {...getFieldProps()}>
+        <styled.div {...mergeProps(varProps, getFieldProps())()}>
```
claude-for-yamada-ui: This `mergeProps` call is unnecessary — it is the same anti-pattern as PR #6580.

`varProps` comes from `useInputBorder`, which returns only `--error-border-color` and `--focus-border-color` (plain CSS custom-property strings). It has no `ref`, no `className`, no `style`, no event handlers. `getFieldProps()` returns clickable/aria props. The two objects have no overlapping keys that `mergeProps` would handle differently from a plain spread.

The original `{...varProps, ...getFieldProps()}` was already correct here. Please revert this hunk and keep the plain spread.

packages/react/src/components/file-input/file-input.tsx
```diff
@@ -124,7 +124,7 @@ export const FileInput = withProvider<"input", FileInputProps>(
 )((props) => {
   const context = useInputPropsContext()
 
-  return { ...context, ...props }
+  return mergeProps(context, props)()
```
claude-for-yamada-ui, line 127: This change is correct and justified. `context` comes from `useInputPropsContext()` and can carry arbitrary user-provided props (`className`, `style`, `ref`, event handlers) that overlap with the component props. `mergeProps` ensures they are combined instead of silently overwritten.

claude-for-yamada-ui (changes requested):

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui: Unnecessary `mergeProps`. `varProps` (from `useInputBorder`) contains only CSS custom-property strings (`--error-border-color`, `--focus-border-color`). It has no `ref`, `className`, `style`, or event handlers. `getFieldProps()` returns clickable/aria props with no key overlap that `mergeProps` would handle differently. The original `{...varProps, ...getFieldProps()}` spread was correct. Revert this hunk.

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui, line 127: Correct use of `mergeProps`. The context from `useInputPropsContext()` can carry `className`, `style`, `ref`, and event handlers that would be silently overwritten by a plain spread. `mergeProps` properly chains/merges them.

codex-for-yamada-ui (changes requested):

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui: Fixed. Reverted to plain spread.

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui: Acknowledged. Kept as-is.

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui: Fixed. Reverted to `{...varProps} {...getFieldProps()}`.

packages/react/src/components/file-input/file-input.tsx
codex-for-yamada-ui: Acknowledged. Kept as-is.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6613, opened 2026-04-12
fix(popover): merge user props with context props via mergeProps

### LINKED ISSUE #6448: `Popover` root uses `{ ...props, ...styleProps }` which drops user `className`/`style`/handlers
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Popover`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/popover/popover.tsx:188`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/popover/popover.tsx#L188) — `useRootComponentProps` call

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6448

## AI used

## Description

Use `mergeProps` in `PopoverRoot` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with style props instead of being silently overwritten by plain object spread.

## Current behavior (updates)

`PopoverRoot` uses `{ ...props, ...styleProps }` which drops user-provided `className`/`style`/event handlers when both sides set them.

## New behavior

Uses `mergeProps(props, styleProps)()` to safely merge all overlapping props.

No

## Additional Information

Part of tracking issue #6430.

### DIFF
--- a/.changeset/calm-waves-popover.md
+++ b/.changeset/calm-waves-popover.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Use `mergeProps` in `Popover` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.
--- a/packages/react/src/components/popover/popover.test.tsx
+++ b/packages/react/src/components/popover/popover.test.tsx
@@ -259,8 +259,35 @@ describe("<Popover />", () => {
     )
     expect(screen.getByTestId("fn-child")).toHaveTextContent("open")
   })
-})
 
+  test("should merge `onOpen` and `onClose` from props context and root props", async () => {
+    const onContextOpen = vi.fn()
+    const onContextClose = vi.fn()
+    const onRootOpen = vi.fn()
+    const onRootClose = vi.fn()
+    const { user } = render(
+      <Popover.PropsContext
+        value={{ onClose: onContextClose, onOpen: onContextOpen }}
+      >
+        <Component onClose={onRootClose} onOpen={onRootOpen} />
+      </Popover.PropsContext>,
+    )
+
+    const triggerButton = await screen.findByRole("button", {
+      name: "Popover Trigger",
+    })
+
+    await user.click(triggerButton)
+
+    expect(onContextOpen).toHaveBeenCalledTimes(1)
+    expect(onRootOpen).toHaveBeenCalledTimes(1)
+
+    await user.keyboard("{escape}")
+
+    expect(onContextClose).toHaveBeenCalledTimes(1)
+    expect(onRootClose).toHaveBeenCalledTimes(1)
+  })
+})
 describe("usePopupAnimationProps", () => {
   test("returns scale animation props by default", () => {
     const { result } = renderHook(() => usePopupAnimationProps())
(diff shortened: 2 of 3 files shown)


codex-for-yamada-ui (changes requested):
## Necessity (Role 1) — PR is unnecessary

The single code change replaces `{ ...props, ...styleProps }` with `mergeProps(props, styleProps)()` at `packages/react/src/components/popover/popover.tsx:188`.

`styleProps` is the return value of `usePopoverStyleProps`, which contains only `{ flip, gutter, matchWidth, placement, strategy }`. None of these are `ref`, `className`, `style`, `css`, or event handlers. `mergeProps` exists to handle exactly those special fields — since none are present in `styleProps`, the `mergeProps` call is equivalent to a plain spread. This is dead weight.

The PR description claims user-provided `className`, `style`, `css`, `ref`, and `on*` handlers are silently overwritten. That overwrite would occur between `props` and `styleProps`, but `styleProps` never carries any of those fields, so no overwrite happens with the existing spread.

The merging of PropsContext values with component props (which *does* involve event handlers) already happens inside `useRootComponentProps` via `mergeProps(contextProps, props)()` (line 352 of `create-component.tsx`), and that path is unaffected by this PR.

This is anti-pattern #1 from `review-anti-patterns.md`: utility over-application.

## Test (Role 5) — test does not validate the change

The added test (`popover.test.tsx:262`) verifies that `onOpen`/`onClose` from `Popover.PropsContext` and root props are both called. This merge is handled by `useRootComponentProps` internally, not by the `props`/`styleProps` merge this PR modifies. The test passes with or without this PR's code change.

## Scope / Convention (Role 8)

The changeset claims behavioral improvement that does not exist at this call site.

claude-for-yamada-ui (changes requested):
### Findings (request changes)

1. **Unnecessary change at `PopoverRoot` call site (Role 1: Necessity)**
   - File: `packages/react/src/components/popover/popover.tsx:188`
   - Diff hunk replaced:
     - `useRootComponentProps({ ...props, ...styleProps })`
     - with `useRootComponentProps(mergeProps(props, styleProps)())`
   - `styleProps` from `usePopoverStyleProps` only contains `flip/gutter/matchWidth/placement/strategy`. It does not carry `className`, `style`, `css`, `ref`, or event handlers, so `mergeProps`' special merging behavior is not used here.
   - If this hunk is deleted, behavior appears unchanged for the claimed bug class; this is dead-weight complexity similar to the documented anti-pattern.
   - **Suggested fix:** revert this hunk (and the new `mergeProps` import) unless you can point to a concrete overlapping prop at this exact site that requires `mergeProps`.

2. **New test does not validate the claimed fix (Role 5: Test)**
   - File: `packages/react/src/components/popover/popover.test.tsx:262`
   - The added test checks `onOpen`/`onClose` merging between `Popover.PropsContext` and root props.
   - That merge already happens inside `useRootComponentProps` (`mergeProps(contextProps, props)`), independent of this PR's `props + styleProps` change. So this test is likely green before and after the code change and does not prove the introduced hunk.
   - **Suggested fix:** replace with a test that fails before the fix at the exact changed boundary, or remove this test if no real bug exists at this call site.

3. **Changeset should not be shipped for a non-justified/no-op behavioral change (Role 1 + Scope/Convention)**
   - File: `.changeset/calm-waves-popover.md`
   - The changeset claims user `className/style/css/ref/on*` are now merged in `Popover`, but the touched hunk does not substantiate that claim.
   - **Suggested fix:** remove/update this changeset after correcting the implementation to a genuinely necessary fix.

PR #6616, opened 2026-04-12
fix(sidebar): merge user props with context props in SidebarItemTooltip via mergeProps

### LINKED ISSUE #6450: `SidebarItemTooltip` overwrites user props via `{ ...tooltipProps, ...props }`
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Sidebar`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/sidebar/sidebar.tsx:1362`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/sidebar/sidebar.tsx#L1362) — `SidebarItemTooltip`

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6450

## AI used

## Description

Use `mergeProps` in `SidebarItemTooltip` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with tooltip context props instead of being silently overwritten by plain object spread.

## Current behavior (updates)

`SidebarItemTooltip` uses `{ ...tooltipProps, ...props }` which drops context-side values when users pass the same props.

## New behavior

Uses `mergeProps(tooltipProps, props)()` to safely merge all overlapping props.

No

## Additional Information

Part of tracking issue #6430.

### DIFF
--- a/.changeset/soft-tooltip-merge.md
+++ b/.changeset/soft-tooltip-merge.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Use `mergeProps` in `SidebarItemTooltip` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.
--- a/packages/react/src/components/sidebar/sidebar.test.tsx
+++ b/packages/react/src/components/sidebar/sidebar.test.tsx
@@ -631,6 +631,57 @@ describe("<Sidebar />", () => {
     expect(screen.getByRole("link", { name: "3" })).toBeInTheDocument()
   })
 
+  test("should merge `tooltipProps` from side panel and item without overwriting user props", async () => {
+    const onSidePanelClick = vi.fn()
+    const onItemClick = vi.fn()
+
+    render(
+      <Sidebar.Root
+        disclosure={{ desktop: { defaultOpen: false } }}
+        mode="icon"
+      >
+        <Sidebar.SidePanel
+          tooltipProps={{
+            className: "from-side-panel",
+            style: { color: "red" },
+            onClick: onSidePanelClick,
+          }}
+        >
+          <Sidebar.Content>
+            <Sidebar.Item
+              label="tooltip-item"
+              value="/tooltip-item"
+              tooltipProps={{
+                className: "from-item",
+                style: { backgroundColor: "blue" },
+                onClick: onItemClick,
+              }}
+            />
+          </Sidebar.Content>
+        </Sidebar.SidePanel>
+      </Sidebar.Root>,
+    )
+
+    fireEvent.pointerEnter(screen.getByRole("link", { name: "tooltip-item" }), {
+      pointerType: "mouse",
+    })
+
+    await waitFor(() => {
+      expect(screen.getByRole("tooltip")).toBeInTheDocument()
+    })
+
+    const tooltip = screen.getByRole("tooltip")
+
+    expect(tooltip).toHaveClass("from-side-panel", "from-item")
+    expect(tooltip).toHaveStyle({ color: "rgb(255, 0, 0)" })
+    expect(tooltip).toHaveStyle({ backgroundColor: "rgb(0, 0, 255)" })
+
+    fireEvent.click(tooltip)
+
+    expect(onSidePanelClick).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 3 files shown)


taroj1205 (changes requested):
@hirotomoyamada 
Typecheck is failing, please fix.

claude-for-yamada-ui (changes requested):
## Necessity (Role 1) — PR should be closed

This change is an instance of **anti-pattern #1 (utility over-application)**.

At `packages/react/src/components/sidebar/sidebar.tsx:1362`, `mergeProps(tooltipProps, props)()` replaces `{ ...tooltipProps, ...props }`.

**Neither operand carries fields that `mergeProps` exists to merge:**

- `tooltipProps` comes from `useItemComponentContext()`. Its type is `Omit<SidebarItemTooltipProps, "children" | "content">`, which extends `TooltipProps`. `TooltipProps` extends `UseTooltipProps`, `PropsWithChildren`, `UsePopupAnimationProps`, and `ThemeProps` — none of which include `ref`, `className`, `style`, `css`, or event handlers.
- `props` is always just `{children}` — `SidebarItemTooltip` is used internally at lines 1389 and 1400 with only children, no additional props.

The upstream merge at line 1270 (`...mergeProps(componentContext.tooltipProps, tooltipProps)()`) already handles merging user-provided tooltip props from `SidebarSidePanel` and `SidebarItem`. By the time `tooltipProps` reaches `SidebarItemTooltip`, it is a flat object of tooltip configuration values. There is nothing for `mergeProps` to merge that plain spread doesn't already handle identically.

The test asserts `className`/`style`/`onClick` merging on the tooltip element, but these assertions test the upstream `mergeProps` at line 1270, not the changed line 1362. Deleting the change at line 1362 would not break the test (the upstream merge already handles it).

**What breaks if we delete this hunk?** Nothing.

Additionally, CI reports TypeScript and Test/Chromium 4/4 failures on this branch.

codex-for-yamada-ui (changes requested):
### 1. Unnecessary/no-op `mergeProps` application in `SidebarItemTooltip`
**File:** `packages/react/src/components/sidebar/sidebar.tsx`  
**Line:** `1359` (hunk `const { children, disabled, ...rest } = ...`)

`SidebarItemTooltip` already receives `tooltipProps` that were merged upstream in `SidebarItem` (`tooltipProps: { ...mergeProps(componentContext.tooltipProps, tooltipProps)() }` around line `1270`).  
At this call site, `props` is effectively just `children` from internal usage (`<SidebarItemTooltip>...</SidebarItemTooltip>`), so replacing `{ ...tooltipProps, ...props }` with `mergeProps(tooltipProps, props)()` does not change behavior.

If this hunk is deleted, nothing functionally breaks for the exposed Sidebar API.

**Concrete fix:** revert this line to the original spread merge, or remove the PR change entirely unless you can show a real runtime path where `SidebarItemTooltip` receives overlapping props requiring `mergeProps`.

---

### 2. Added test does not exercise the changed line
**File:** `packages/react/src/components/sidebar/sidebar.test.tsx`  
**Line:** `631` (new test block)

The new test validates merge behavior between `Sidebar.SidePanel.tooltipProps` and `Sidebar.Item.tooltipProps`. That merge already happens in `SidebarItem` via `mergeProps(componentContext.tooltipProps, tooltipProps)()` before `SidebarItemTooltip` runs.

So this test would pass even without the code change at line `1359`, meaning it does not prove the fix claimed in the PR.

**Concrete fix:** either:
1. remove this test with the no-op code change, or  
2. add a test that fails on the old implementation and passes on the new one (currently there is no clear public path to do that for `SidebarItemTooltip` props merging).

---

### 3. Changeset is not justified for a no-op behavioral change
**File:** `.changeset/soft-tooltip-merge.md`  
**Line:** `1`

Because the implementation change appears behaviorally redundant, publishing a patch note claiming a user-visible fix is misleading.

**Concrete fix:** remove this changeset unless a real behavioral regression is demonstrated with a failing-then-passing test.

PR #6618, opened 2026-04-12
fix(data-list): merge user props with context props via mergeProps

### LINKED ISSUE #6452: `DataList` term and description overwrite user props via JSX double spread
@yamada-ui/react

### Package Version
2.2.0

### Description
In `DataList`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/data-list/data-list.tsx:172`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/data-list/data-list.tsx#L172) — `{...termProps} {...customTermProps}`
- [`packages/react/src/components/data-list/data-list.tsx:177`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/data-list/data-list.tsx#L177) — `{...termProps} {...customTermProps}`
- [`packages/react/src/components/data-list/data-list.tsx:188-189`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/data-list/data-list.tsx#L188-L189) — `{...descriptionProps} {...customDescriptionProps}`
- [`packages/react/src/components/data-list/data-list.tsx:196-197`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/data-list/data-list.tsx#L196-L197) — `{...descriptionProps} {...customDescriptionProps}`

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6452

## AI used

## Description

Fix DataList term/description prop merging to avoid user props being overwritten by context props.

## Current behavior (updates)

`DataList.Item` used JSX double spreads for `termProps` and `descriptionProps`, which overwrote conflicting props instead of merging event handlers and class names safely.

## New behavior

`DataList.Item` now uses `mergeProps(contextProps, userProps)()` for `DataList.Term` and `DataList.Description`, and includes regression tests to verify merged event handlers and class names.

No

## Additional Information

- Added a changeset for `@yamada-ui/react` patch release.
- Targeted tests passed: `pnpm react test:jsdom --run src/components/data-list/`.


claude-for-yamada-ui (changes requested):
### Necessity (Role 1)

All four `mergeProps` call sites are justified. Both `termProps` (from context) and `customTermProps` (from item) are typed as `HTMLStyledProps`, meaning they can carry `ref`, `className`, `style`, `css`, and event handlers. Plain spread would overwrite these — `mergeProps` correctly merges them. This is not a mechanical over-application.

### Test (Role 5)

The fix modifies four render paths but the new tests only exercise the **scalar** `term`/`description` branches. The **array** branches (`isArray(term)` at `data-list.tsx:171` and `isArray(description)` at `data-list.tsx:188`) were also changed to use `mergeProps` but have no corresponding regression tests.

Add tests for the array paths that pass both context-level and item-level props with overlapping `className` and `onClick`, then assert both class names are present and both handlers fire on each generated element — mirroring the existing scalar tests.

All other roles (Correctness, Accessibility, Type Safety, API/Breaking Change, Style/Theme, Scope/Convention, Consistency) have no issues.

codex-for-yamada-ui (changes requested):
### Findings

1. **Test coverage gap for changed call sites (Role 1: Necessity, Role 5: Test)**
   - **Where:** `packages/react/src/components/data-list/data-list.tsx:172`, `packages/react/src/components/data-list/data-list.tsx:188`, and tests added at `packages/react/src/components/data-list/data-list.test.tsx:256`.
   - **Issue:** The fix changes **four** render paths (term single/array and description single/array), but the new regression tests exercise only the single-value paths. The array paths are part of the same bug surface and were modified too, so this leaves part of the fix unproven.
   - **Concrete fix:** Add regression tests for `term` and `description` array branches that assert merged behavior (both class names present and both click handlers called) on generated `DataList.Term`/`DataList.Description` items. This should validate each changed hunk, not just the scalar branches.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6619, opened 2026-04-12
fix(rating): merge user props with context props via mergeProps

### LINKED ISSUE #6453: `Rating` group overwrites user props via JSX double spread
@yamada-ui/react

### Package Version
2.2.0

### Description
In `Rating`, user-provided props are merged with context/getter props via plain object spread (or JSX double spread). This overwrites `className`, `style`, `css`, `ref`, and `on*` event handlers set on the context side when the user also passes them, so users cannot extend styling or attach handlers without destroying the component's internal wiring.

### Link to Reproduction

### Steps to Reproduce
2. Inspect the DOM node.

### Browser
Chrome

### Operating System
- [x] macOS

Yes

Yes

### Additional Information

**Affected locations:**

- [`packages/react/src/components/rating/rating.tsx:213`](https://github.com/yamada-ui/yamada-ui/blob/main/packages/react/src/components/rating/rating.tsx#L213) — `{...runIfFn(groupProps, value)} {...rest}`

Parent tracking: #6430

### AI used

hirotomoyamada:
Closes #6453

## AI used

## Description

Fixes `Rating` prop merging where JSX double spreads could overwrite user-provided props.

## Current behavior (updates)

`RatingGroup` and `RatingIcon` used double spread/object spread patterns that did not safely merge props, causing `className`, `style`, and event handlers to be overwritten instead of merged.

## New behavior

Uses `mergeProps(contextProps, userProps)()` in the affected `Rating` locations so user and context/getter props are merged safely.
Adds a regression test to verify `groupProps` user props are preserved alongside slot props.

No

## Additional Information

N/A


claude-for-yamada-ui (changes requested):
### Necessity

**`RatingIcon` call site (rating.tsx ~L252):** `mergeProps({ "--clip-path": clipPath }, runIfFn(iconProps, groupValue))()` — the left-hand object `{ "--clip-path": clipPath }` contains only a CSS custom property. It has no `ref`, `className`, `style`, `css`, or event handlers. `mergeProps` exists to merge those fields correctly; there is nothing to merge here. Plain spread is equivalent and simpler. This is utility over-application (anti-pattern #1). Revert to:
```tsx
<RatingIcon
  {...{ "--clip-path": clipPath }}
  {...runIfFn(iconProps, groupValue)}
>
```

**`RatingGroup` call site (rating.tsx ~L213):** `mergeProps(runIfFn(groupProps, value), rest)()` — both objects can carry `className`, `style`, `ref`, and event handlers, so `mergeProps` is justified here.

### Correctness

**Argument order is backwards in `RatingGroup`:** `mergeProps(runIfFn(groupProps, value), rest)()` gives `rest` (internal slot props from `withContext`) last-wins precedence over `groupProps` (user props). The PR description says it fixes "user-provided props being overwritten" but this order still lets slot props overwrite user props on non-mergeable keys. The original code `{...runIfFn(groupProps, value)} {...rest}` had the same problem, so the fix is incomplete. The correct order should be `mergeProps(rest, runIfFn(groupProps, value))()` so user-provided `groupProps` take precedence.

### Type Safety

**Test type error (rating.test.tsx ~L38):** `groupProps` is typed as `FunctionOrValue<number, RatingGroupProps>` where `RatingGroupProps` requires `value: number`. The test passes a function returning `{ className, style, onClick }` without `value`, causing a TS2322 error. The test needs to satisfy the `RatingGroupProps` constraint or the type needs to be relaxed.

### Test

The test idea is good but it does not compile due to the type issue above, and it should also verify that the merge precedence is correct (user props win over slot props on conflicting keys).

codex-for-yamada-ui (changes requested):
### Findings

1. **[Correctness / Consistency] `mergeProps` argument order is reversed in `RatingGroup`**
   - File: `packages/react/src/components/rating/rating.tsx` (around line 213)
   - Current change: `mergeProps(runIfFn(groupProps, value), rest)()`
   - Why this is a problem: across the component system, merge precedence is typically `mergeProps(contextProps, userProps)()`. Reversing that here can let slot/internal props win on conflicting scalar/style keys and changes event handler execution order.
   - Concrete fix:
     ```tsx
     <styled.div {...mergeProps(rest, runIfFn(groupProps, value))()}>
     ```
   - Test gap: please add/adjust a regression test to validate the intended precedence behavior (user props should not be overridden by slot/internal props on conflict).

2. **[Necessity] `mergeProps` is unnecessary for the `RatingIcon` hunk**
   - File: `packages/react/src/components/rating/rating.tsx` (around line 252)
   - Current change wraps:
     ```tsx
     mergeProps({ "--clip-path": clipPath }, runIfFn(iconProps, groupValue))()
     ```
   - Why this is unnecessary: the left-hand object does not contain `ref`, `className`, `style`, `css`, or event handlers. For this site, `mergeProps` adds complexity without behavioral gain over plain spread.
   - Concrete fix: revert this hunk to simple spreads:
     ```tsx
     <RatingIcon
       {...{ "--clip-path": clipPath }}
       {...runIfFn(iconProps, groupValue)}
     >
     ```

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

Issue #6460, opened 2026-04-11
test(react): migrate component tests to Vitest Browser Mode


hirotomoyamada:
### Description

Track the migration of each component test in `packages/react` from the jsdom-based `#test` utilities to Vitest Browser Mode (`#test/browser`).

> GitHub limits each parent issue to 100 sub-issues. The remaining 13 child issues are tracked in #6574 (Part 2).

### Problem Statement/Justification

Most component tests in `packages/react` still run under jsdom via `#test`. Running tests in a real browser via Vitest Browser Mode gives us more reliable results in an environment closer to the user's, especially for components that rely on layout, focus management, pointer events, or real DOM APIs. This is the follow-up work for #5285, broken down per component so each migration can be reviewed and merged independently.

The following components are already migrated and can be used as references:

- [badge.test.tsx](packages/react/src/components/badge/badge.test.tsx)
- [carousel.test.tsx](packages/react/src/components/carousel/carousel.test.tsx)
- [flex.test.tsx](packages/react/src/components/flex/flex.test.tsx)
- [list.test.tsx](packages/react/src/components/list/list.test.tsx)
- [loading.test.tsx](packages/react/src/components/loading/loading.test.tsx)
- [number-input.test.tsx](packages/react/src/components/number-input/number-input.test.tsx)
- [z-stack.test.tsx](packages/react/src/components/stack/z-stack.test.tsx)

### Proposed Solution or API

For each component's test file:

1. Replace `import { ... } from "#test"` with `import { page, render } from "#test/browser"` (add `a11y` as needed).
2. Rewrite assertions to use Browser Mode APIs (`page.getBy*`, `await expect.element(...)`, etc.).
3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/<name>/`.

Reference: [badge.test.tsx](packages/react/src/components/badge/badge.test.tsx)

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

Yes

### AI used

- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

### Additional Information

Child issues are listed in the Sub-issues section below. Due to GitHub's 100 sub-issue limit per parent, 13 additional children are tracked in #6574.

Related: #5285

PR #6639, opened 2026-04-13
test(airy): migrate tests to Vitest Browser Mode

### LINKED ISSUE #6463: test(airy): migrate to Vitest Browser Mode
### Description

Migrate the `airy` component tests to Vitest Browser Mode.

- [airy.test.tsx](packages/react/src/components/airy/airy.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/airy/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6463

## AI used

## Description

Migrates `airy.test.tsx` from jsdom `#test` helpers to Vitest Browser Mode (`#test/browser`), following the parent issue #6460 and the reference `badge.test.tsx` pattern.

## Current behavior (updates)

Component tests ran under jsdom via `#test` imports.

## New behavior

Tests use `page`, `render`, and `a11y` from `#test/browser` with `expect.element` assertions. The disabled-state test no longer performs `user.click` on a disabled button (Playwright rejects that); it asserts `toBeDisabled()` instead.

No

## Additional Information

Verified with `pnpm react test:browser --run src/components/airy/`.


claude-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/airy/airy.test.tsx
```diff
@@ -38,41 +37,38 @@ describe("<Airy />", () => {
...
 
-    const onText = screen.getByText("ON")
-    expect(onText).toBeInTheDocument()
+    const onText = page.getByText("ON")
+    await expect.element(onText).toBeVisible()
 
-    user.click(button)
-    expect(onText).toBeInTheDocument()
+    await user.click(button)
+    await expect.element(onText).toBeVisible()
```
codex-for-yamada-ui: **Issue:** This read-only assertion is a false positive in browser mode. `Airy` does not flip immediately on click; it first awaits the `useAnimation()` exit transition and only then switches from `from` to `to`. That means `"ON"` is still visible right after `await user.click(button)` even when `readOnly` is broken, so this test no longer verifies that clicks are ignored.

**Fix:** After clicking, wait past the transition and assert the state still stays on `from`, e.g. `await vi.waitFor(() => expect(button.element()).toHaveAttribute("data-value", "from"))`, or use `waitForAnimationFrame` before asserting `"ON"` is still rendered and `"OFF"` never appears.

codex-for-yamada-ui (changes requested):
The read-only test no longer verifies the contract. After migrating to Vitest Browser Mode, the assertion that `ON` remains visible right after a click does not reliably distinguish a working `readOnly` path from a broken one, because `Airy`'s click handler awaits its exit animation before flipping the value.

packages/react/src/components/airy/airy.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

Issue #6574, opened 2026-04-11
test(react): migrate component tests to Vitest Browser Mode (Part 2)


hirotomoyamada:
### Description

Track the migration of each component test in `packages/react` from the jsdom-based `#test` utilities to Vitest Browser Mode (`#test/browser`).

This is Part 2 of the tracking issue — continuation of #6460, which reached GitHub's 100 sub-issue per parent limit.

### Problem Statement/Justification

Most component tests in `packages/react` still run under jsdom via `#test`. Running tests in a real browser via Vitest Browser Mode gives us more reliable results in an environment closer to the user's, especially for components that rely on layout, focus management, pointer events, or real DOM APIs. This is the follow-up work for #5285, broken down per component so each migration can be reviewed and merged independently.

### Proposed Solution or API

For each component's test file:

1. Replace `import { ... } from "#test"` with `import { page, render } from "#test/browser"` (add `a11y` as needed).
2. Rewrite assertions to use Browser Mode APIs (`page.getBy*`, `await expect.element(...)`, etc.).
3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/<name>/`.

Reference: [badge.test.tsx](packages/react/src/components/badge/badge.test.tsx)

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

Yes

### AI used

- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

### Additional Information

Child issues are listed in the Sub-issues section below. See #6460 for the first 100 child issues.

Related: #5285, #6460

PR #6640, opened 2026-04-13
test(textarea): migrate tests to Vitest Browser Mode

### LINKED ISSUE #6565: test(textarea): migrate to Vitest Browser Mode
### Description

Migrate the `textarea` component tests to Vitest Browser Mode.

- [textarea.test.tsx](packages/react/src/components/textarea/textarea.test.tsx)
- [use-autosize.test.tsx](packages/react/src/components/textarea/use-autosize.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/textarea/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6565

## AI used

## Description

Migrate `textarea` component tests to Vitest Browser Mode.

## Current behavior (updates)

`textarea.test.tsx` and `use-autosize.test.tsx` were using jsdom-based `#test` utilities.

## New behavior

Both test files now run in Browser Mode using `#test/browser` APIs (`page`, `render`, `renderHook`) while preserving existing coverage.

No.

## Additional Information

- Verified locally in the dedicated worktree with:
  - `pnpm react test:browser --run src/components/textarea/`
- Parent tracking issue: #6574


packages/react/src/components/textarea/textarea.test.tsx
```diff
@@ -10,64 +10,69 @@ describe("<Textarea />", () => {
...
         removeEventListener: vi.fn(),
       },
       writable: true,
     })
-    render(<Textarea autosize />)
-    expect(screen.getByRole("textbox")).toHaveProperty("rows", 2)
-    Object.defineProperty(global.document, "fonts", {
+    await render(<Textarea autosize />)
+    await expect
+      .poll(() => (page.getByRole("textbox").element() as HTMLTextAreaElement).rows)
```
claude-for-yamada-ui: Issue: This `.poll(...)` callback line is 84 characters wide, over the repo's Prettier `printWidth: 80`. The inline `(page.getByRole("textbox").element() as HTMLTextAreaElement).rows` pushes past the limit.

Fix: Extract the textarea into a local before the assertion — e.g. `const textarea = page.getByRole("textbox").element() as HTMLTextAreaElement; await expect.poll(() => textarea.rows).toBe(2)` — or rewrap the arrow body so the resulting line stays within 80 columns.

packages/react/src/components/textarea/use-autosize.test.tsx
```diff
@@ -96,6 +96,32 @@ const AutoSizeTextarea = ({
...
+  } else {
+    textarea.value = value
+  }
+}
+
+const changeTextareaValue = (textarea: HTMLTextAreaElement, value: string) => {
+  setNativeValue(textarea, value)
+  textarea.dispatchEvent(new Event("input", { bubbles: true }))
+  textarea.dispatchEvent(new Event("change", { bubbles: true }))
+}
```
claude-for-yamada-ui: Issue: This file reimplements a React-Testing-Library-style `setNativeValue` + manual `dispatchEvent("input"/"change")` helper. Every other migrated browser-mode test in this repo (e.g. `number-input.test.tsx`, `pin-input.test.tsx`, `password-input.test.tsx`) drives input through the real `userEvent` API obtained from `await render(...)`. Reintroducing synthetic native-value dispatch adds ~22 lines that no sibling test carries and defeats a core reason for migrating to Browser Mode.

Fix: Drop `setNativeValue` / `changeTextareaValue`; destructure `{ user }` from `await render(<AutoSizeTextarea ... />)` and drive input with `await user.clear(textarea)` + `await user.type(textarea, "Line 1\nLine 2\n...")` (or `await user.fill(textarea, value)`), matching the existing input-family precedents.

packages/react/src/components/textarea/use-autosize.test.tsx
```diff
@@ -114,51 +140,49 @@ describe("useAutosize", () => {
...
+  test("adjusts the rows of the textarea based on content", async () => {
+    await render(<AutoSizeTextarea />)
+    const textarea = getTextarea()
 
-    const textarea = screen.getByTestId("autosize-textarea")
+    changeTextareaValue(textarea, "Line 1\nLine 2\nLine 3\nLine 4\nLine 5")
 
-    fireEvent.change(textarea, {
-      target: { value: "Line 1\nLine 2\nLine 3\nLine 4\nLine 5" },
+    await vi.waitFor(() => {
```
claude-for-yamada-ui: Issue: This file uses `await vi.waitFor(() => { expect(textarea.rows).toBe(N) })` at every wait site (here and at 160, 171, 180, 204, 238), whereas the sibling `textarea.test.tsx:73` uses `await expect.poll(() => ...).toBe(2)` for the analogous autosize assertion. Choosing two waiting idioms from one PR sets two conflicting precedents for the next migration.

Fix: Standardize on one primitive across both files. `await expect.poll(() => textarea.rows).toBe(N)` pairs naturally with the browser-mode assertions already used in `textarea.test.tsx`; apply it uniformly (or, conversely, switch `textarea.test.tsx:73` to `vi.waitFor`).

packages/react/src/components/textarea/use-autosize.test.tsx
```diff
@@ -172,101 +196,83 @@ describe("useAutosize", () => {
...
-    })
+    changeTextareaValue(textarea, "Line 1\nLine 2\nLine 3\nLine 4\nLine 5")
 
-    expect((textarea as HTMLTextAreaElement).rows).not.toBe(5)
+    expect(textarea.rows).not.toBe(5)
   })
 
-  test("resizes when value changes via rerender", () => {
-    const { rerender } = render(<AutoSizeTextarea minRows={1} />)
+  test("resizes when value changes via rerender", async () => {
```
claude-for-yamada-ui, line 213: Issue: `resizes when value changes via rerender` calls `rerender(<AutoSizeTextarea minRows={1} />)` later in the test but asserts nothing after the rerender. The only assertion (`rows === 3`) runs before the rerender, so the test name promises rerender-driven resizing but actually overlaps with `adjusts the rows of the textarea based on content`.

Fix: Either add a post-rerender assertion that exercises the rerender-driven resize path (e.g. change props so the expected `rows` differ, then `await vi.waitFor(() => expect(textarea.rows).toBe(<new value>))`), or remove the unused rerender call and rename the test to match what it covers.

claude-for-yamada-ui (changes requested):
The migration to Vitest Browser Mode is the right direction, but four spots still carry jsdom-era patterns that the rest of the repo's browser-mode tests have moved past: the `expect.poll` line in `textarea.test.tsx` exceeds the repo's 80-column formatter width, `use-autosize.test.tsx` reimplements a React-Testing-Library-style `setNativeValue` + `dispatchEvent` helper that sibling input tests drop in favor of `user` from `render`, the two migrated files disagree on whether to wait with `expect.poll` or `vi.waitFor`, and the `resizes when value changes via rerender` case never asserts after the rerender.

packages/react/src/components/textarea/use-autosize.test.tsx
codex-for-yamada-ui: Issue: `changeTextareaValue()` dispatches both `input` and `change` for each update. `useAutosize()` wires its resize logic through `getTextareaProps().onChange`, so dispatching a second event on top of `input` can invoke the resize path twice for a single edit and weakens the later "does not resize when value has not changed" assertion.

Fix: Simulate one edit per call. Prefer the browser-mode `user.type()` flow (see the other input-family tests), or keep `setNativeValue()` and dispatch only a single event that React will route to `onChange`.

packages/react/src/components/textarea/textarea.test.tsx
```diff
@@ -10,64 +10,69 @@ describe("<Textarea />", () => {
...
+      .toBe(2)
```
codex-for-yamada-ui: Issue: The new `expect.poll` assertion is wrapped differently from surrounding browser-mode tests in this file and in precedents like `loading.test.tsx` — the `.poll()` callback should land on its own indented line rather than being stuffed into the same physical line.

Fix: Rewrap to match the local style — for example, extract the textbox element into a local, or break the `.poll(...)` callback onto its own indented line so the chain reads:

```ts
await expect
  .poll(() => (page.getByRole("textbox").element() as HTMLTextAreaElement).rows)
  .toBe(2)
```

codex-for-yamada-ui (changes requested):
Two issues. The `changeTextareaValue` helper dispatches both `input` and `change` per call, which can exercise the autosize resize path twice for one edit and weaken the later "value has not changed" coverage. And the new `expect.poll` assertion in `textarea.test.tsx` is wrapped in a way that doesn't match the file's usual multi-line `.poll(...)` style.

packages/react/src/components/textarea/textarea.test.tsx
composer-for-yamada-ui: Issue: This `await expect` / `.poll` / `.toBe(2)` chain isn't wrapped the way the repo's formatter expects — the `.poll(` callback should sit on its own indented line rather than being packed into one line, matching patterns such as `number-input.test.tsx`.

Fix: Rewrap to:

```ts
await expect
  .poll(
    () => (page.getByRole("textbox").element() as HTMLTextAreaElement).rows,
  )
  .toBe(2)
```

Or extract the textarea element into a local before the assertion.

composer-for-yamada-ui (changes requested):
The `expect.poll` assertion in `textarea.test.tsx` doesn't match the multi-line `.poll(...)` shape used elsewhere in browser-mode tests.

packages/react/src/components/textarea/textarea.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/use-autosize.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/use-autosize.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/use-autosize.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/textarea.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/use-autosize.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/textarea/textarea.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6641, opened 2026-04-13
test(timeline): migrate tests to Vitest Browser Mode

### LINKED ISSUE #6566: test(timeline): migrate to Vitest Browser Mode
### Description

Migrate the `timeline` component tests to Vitest Browser Mode.

- [timeline.test.tsx](packages/react/src/components/timeline/timeline.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/timeline/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6566

## AI used

## Description

Migrates `timeline.test.tsx` from jsdom `#test` helpers to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

Timeline tests ran in jsdom.

## New behavior

Timeline tests run in a real browser via Vitest Browser Mode.

No

## Additional Information

Part of #6574.


packages/react/src/components/timeline/timeline.test.tsx
```diff
@@ -31,62 +31,73 @@ describe("<Timeline />", () => {
...
+    ).toBe("H3")
+    expect(page.getByText("Step 1 description").element().tagName).toBe("P")
   })
 
-  test("renders numbers in indicators with `number` variant", () => {
-    render(<Timeline.Root variant="number" items={items} />)
-    const listItems = screen.getAllByRole("listitem")
+  test("renders numbers in indicators with `number` variant", async () => {
+    await render(<Timeline.Root variant="number" items={items} />)
+    const listItems = document.querySelectorAll('[role="listitem"]')
```
claude-for-yamada-ui: Issue: This loop (and the one at line 82) reaches for `document.querySelectorAll('[role="listitem"]')` while the rest of this file (and sibling browser-mode migrations such as `list.test.tsx` / `tag.test.tsx`) uses the `page` locator API (`page.getByRole("listitem")`, `.first()`, `.nth(...)`). Falling back to raw DOM queries also drops the implicit non-zero-items check that `getAllByRole` used to give us — the `forEach` now becomes a no-op on an empty collection.

Fix: Replace with locator-based iteration, e.g. iterate `page.getByRole("listitem").nth(i).element()` over a known-length loop (`items.length` / `itemsWithIndicator.length`), matching the pattern already used in the last test and in sibling files.

packages/react/src/components/timeline/timeline.test.tsx
```diff
@@ -31,62 +31,73 @@ describe("<Timeline />", () => {
...
+  test("does not render numbers in indicators without `number` variant", async () => {
+    await render(<Timeline.Root items={items} />)
+    const listItems = document.querySelectorAll('[role="listitem"]')
 
-    listItems.forEach((item) => {
-      const indicator = item.children[0]?.children[0]
-      expect(indicator).not.toHaveTextContent(/\d/)
+    listItems.forEach((itemEl) => {
+      const indicator = itemEl.children[0]?.children[0]
+      expect(indicator?.textContent ?? "").not.toMatch(/\d/)
```
claude-for-yamada-ui: Issue: The assertion changed from `expect(indicator).not.toHaveTextContent(/\d/)` to `expect(indicator?.textContent ?? "").not.toMatch(/\d/)`. If `indicator` is missing, the expression collapses to `expect("").not.toMatch(/\d/)`, which passes trivially — masking the case where the indicator node disappeared. Line 76 still uses `toHaveTextContent` on the same kind of node, proving the matcher works here.

Fix: Revert to `expect(indicator).not.toHaveTextContent(/\d/)` (or assert `indicator` is defined first) so a missing indicator remains a test failure.

packages/react/src/components/timeline/timeline.test.tsx
```diff
@@ -31,62 +31,73 @@ describe("<Timeline />", () => {
...
       { indicator: "A", title: "Step 1" },
       { title: "Step 2" },
     ]
 
-    render(<Timeline.Root variant="number" items={itemsWithIndicator} />)
-    const listItems = screen.getAllByRole("listitem")
-    const firstIndicator = listItems[0]?.children[0]?.children[0]
-    const secondIndicator = listItems[1]?.children[0]?.children[0]
+    await render(<Timeline.Root variant="number" items={itemsWithIndicator} />)
+    const firstIndicator = page.getByRole("listitem").nth(0).element()
```
claude-for-yamada-ui: Issue: `page.getByRole("listitem").nth(0)` here, but `page.getByRole("listitem").first()` is used on lines 36 and 57 within this same file for the exact same semantic operation. Mixing `.first()` and `.nth(0)` in one file is inconsistent.

Fix: Use `.first()` for consistency — `page.getByRole("listitem").first().element().children[0]?.children[0]`.

claude-for-yamada-ui (changes requested):
The migration is largely mechanical, but three spots regressed vs. the jsdom originals or diverge from surrounding patterns: two tests fall back to `document.querySelectorAll` instead of the `page` locator API that the rest of this file and sibling browser-mode migrations use, the indicator-text-content assertion was downgraded to a `.textContent ?? ""` string check that no longer fails when the indicator is missing, and the final test mixes `.nth(0)` with `.first()` for the same operation already present elsewhere in this file.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Issue: These assertions regressed. `document.querySelectorAll('[role="listitem"]')` at lines 72 and 82 silently returns an empty NodeList — the `forEach` loop does nothing if no listitems exist, so the test passes vacuously. Similarly, `indicator?.textContent ?? ""` on this line collapses to asserting `""` has no digits when the indicator itself is missing. The original `getAllByRole` would have thrown on an empty collection, and `toHaveTextContent` would have thrown on `undefined` — both failure modes are now gone.

Fix: Keep browser-mode queries but preserve the fail-on-missing semantics. Locate listitems with `page.getByRole("listitem").nth(i)` (or assert the count first), and keep the element-level matcher on the indicator — e.g. `expect(page.getByRole("listitem").nth(i).element().children[0]?.children[0]).not.toHaveTextContent(/\d/)`.

codex-for-yamada-ui (changes requested):
The migrated assertions are weaker than the jsdom originals: `document.querySelectorAll` returns an empty `NodeList` silently, and `indicator?.textContent ?? ""` hides a missing indicator node.

packages/react/src/components/timeline/timeline.test.tsx
```diff
@@ -31,62 +31,73 @@ describe("<Timeline />", () => {
...
       expect(indicator).toHaveTextContent(String(index + 1))
     })
   })
 
-  test("does not render numbers in indicators without `number` variant", () => {
-    render(<Timeline.Root items={items} />)
-    const listItems = screen.getAllByRole("listitem")
+  test("does not render numbers in indicators without `number` variant", async () => {
+    await render(<Timeline.Root items={items} />)
+    const listItems = document.querySelectorAll('[role="listitem"]')
```
composer-for-yamada-ui: Issue: Both this loop and the one at line 72 still use `document.querySelectorAll('[role="listitem"]')` while the rest of the file uses the `page` locator API (`page.getByRole("listitem")`, `.first()`, `.nth(...)`). Sibling migrated browser tests such as `list.test.tsx` and `tag.test.tsx` stay on `page` for queries, so this mixes global DOM queries with locator-based queries in the same file and also drops the "non-zero items" check that `getAllByRole` implicitly gave us.

Fix: Iterate with `page.getByRole("listitem")` and `.nth(i)` using the known item count (`items.length` / `itemsWithIndicator.length`), deriving `itemEl` from `.element()` so the list-item walks match the surrounding tests.

packages/react/src/components/timeline/timeline.test.tsx
composer-for-yamada-ui: Issue: The assertion became `expect(indicator?.textContent ?? "").not.toMatch(/\d/)`. If `indicator` is missing, this collapses to asserting `""` has no digits, which passes trivially and no longer flags the case where the indicator node disappeared — unlike the previous direct `.not.toHaveTextContent(/\d/)` on the element.

Fix: Keep a strict expectation on the element (e.g. assert `indicator` is defined first, or `expect(indicator!).not.toHaveTextContent(/\d/)`) so a missing indicator remains a test failure.

packages/react/src/components/timeline/timeline.test.tsx
```diff
@@ -31,62 +31,73 @@ describe("<Timeline />", () => {
...
 
-    render(<Timeline.Root variant="number" items={itemsWithIndicator} />)
-    const listItems = screen.getAllByRole("listitem")
-    const firstIndicator = listItems[0]?.children[0]?.children[0]
-    const secondIndicator = listItems[1]?.children[0]?.children[0]
+    await render(<Timeline.Root variant="number" items={itemsWithIndicator} />)
+    const firstIndicator = page.getByRole("listitem").nth(0).element()
+      .children[0]?.children[0]
+    const secondIndicator = page.getByRole("listitem").nth(1).element()
+      .children[0]?.children[0]
```
composer-for-yamada-ui: Issue: `firstIndicator` / `secondIndicator` are computed with a split chain — `page.getByRole("listitem").nth(...).element()` on one line and `.children[0]?.children[0]` on the next. Elsewhere in this same file the pattern is to assign `.element()` to a named local (`itemEl`) first and then read `children`.

Fix: Introduce locals such as `const firstItem = page.getByRole("listitem").nth(0).element()` (and `secondItem`), then derive `firstItem.children[0]?.children[0]` — mirroring the `item` / `itemEl` pattern used above in the same file.

composer-for-yamada-ui (changes requested):
Three concerns: two tests swap from `page` locators to raw `document.querySelectorAll`, the indicator assertion silently passes when the node is missing, and the `firstIndicator`/`secondIndicator` split chain diverges from the file's named-local convention.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/timeline/timeline.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6650, opened 2026-04-13
test(bleed): migrate to Vitest Browser Mode

### LINKED ISSUE #6469: test(bleed): migrate to Vitest Browser Mode
### Description

Migrate the `bleed` component tests to Vitest Browser Mode.

- [bleed.test.tsx](packages/react/src/components/bleed/bleed.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/bleed/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6469

## AI used

## Description

Migrate `bleed` component tests from jsdom (`#test`) to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

`packages/react/src/components/bleed/bleed.test.tsx` used `screen` assertions from `#test`, including style assertions that depended on jsdom behavior.

## New behavior

- Converted queries/assertions to Browser Mode style (`page.getBy*`, `await expect.element(...)`)
- Stabilized full-bleed style tests by rendering in a fixed-width wrapper and asserting computed inline margins are non-zero where expected

No

## Additional Information

- Verified with: `pnpm react test:browser --run src/components/bleed/` (pass)
- Part of parent issue: #6460


packages/react/src/components/bleed/bleed.test.tsx
```diff
@@ -10,34 +10,54 @@ describe("<Bleed />", () => {
...
+    await render(
+      <div style={{ width: "200px" }}>
+        <Bleed inline="full">Box</Bleed>
+      </div>,
+    )
+
+    const style = getComputedStyle(page.getByText("Box").element())
+
+    expect(style.marginInlineEnd).not.toBe("0px")
+    expect(style.marginInlineStart).not.toBe("0px")
```
codex-for-yamada-ui: **Issue:** The migrated assertions only check `not.toBe("0px")`, which is too weak for the `inline="full"` contract. `Bleed` is implemented to apply `calc(50% - 50vw)` for full inline bleed, but this test would still pass for incorrect non-zero values (for example `-1px`), so regression coverage was reduced compared to the previous exact-behavior check.

**Fix:** Assert the resolved computed value against the expected full-bleed amount instead of only non-zero. For example, derive the expected value from container width and viewport width, then compare with `toBeCloseTo` (or equivalent numeric tolerance) for `marginInlineStart`/`marginInlineEnd`. Apply the same stronger expectation pattern to the `inlineStart='full'` and `inlineEnd='full'` cases so the active side verifies full bleed, not just non-zero.

codex-for-yamada-ui (changes requested):
- テスト: The migrated assertions only check `not.toBe("0px")`, which is too weak for the `inline="full"` contract. `Bleed` is implemented to apply `calc(50% - 50vw)` for full inline bleed, but this test would still pass for incorrect non-zero values (for example `-1px`), so regression coverage was reduced compared to the previous exact-behavior check.

claude-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/bleed/bleed.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

PR #6664, opened 2026-04-13
test(separator): migrate to Vitest Browser Mode

### LINKED ISSUE #6546: test(separator): migrate to Vitest Browser Mode
### Description

Migrate the `separator` component tests to Vitest Browser Mode.

- [separator.test.tsx](packages/react/src/components/separator/separator.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/separator/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6546

## AI used

## Description

Migrate `separator` component tests from jsdom (`#test`) to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

`packages/react/src/components/separator/separator.test.tsx` used `screen` assertions from `#test`.

## New behavior

- Switched test utilities to `#test/browser` (`page`, `render`)

No

## Additional Information

- Verified with: `pnpm react test:browser --run src/components/separator/` (pass)
- Part of parent issue: #6460


claude-for-yamada-ui (dismissed):

packages/react/src/components/separator/separator.test.tsx
```diff
@@ -1,22 +1,26 @@
-import { render, screen } from "#test"
+import { page, render } from "#test/browser"
 import { Separator } from "./separator"
 
 describe("<Separator />", () => {
-  test("renders component correctly", () => {
-    render(<Separator />)
+  test("renders component correctly", async () => {
+    await render(<Separator />)
```
codex-for-yamada-ui: **Issue**: This test only calls `await render(<Separator />)` with no assertion, so it is a no-op that duplicates what the subsequent `className` / tag-name tests already verify. The migration keeps a redundant case in the suite.

**Fix**: Either remove this test, or replace its body with `await a11y(<Separator />)` (and extend the `#test/browser` import on line 1 to include `a11y`) so the `renders component correctly` slot carries the same accessibility baseline already adopted in `kbd`, `text`, and `box` browser-mode tests.

codex-for-yamada-ui (changes requested):
The migration itself is correct, but the `renders component correctly` case only awaits `render(<Separator />)` with no assertion and is now redundant with the class-name / tag-name tests that follow. Either drop this test or, matching the established browser-mode convention (see e.g. `kbd`, `text`, `box`), make it an accessibility baseline with `await a11y(<Separator />)` so this slot carries real coverage.

packages/react/src/components/separator/separator.test.tsx
composer-for-yamada-ui: **Issue**: The `renders component correctly` test body is just `await render(<Separator />)` with no assertion and no accessibility audit, even though the pattern used for this test name in other already-migrated browser-mode suites (e.g. `kbd`, `text`, `box`) is `await a11y(<Component />)`. As written, this test adds no coverage beyond what the sibling tests already provide.

**Fix**: Import `a11y` from `#test/browser` on line 1 and replace the test body with `await a11y(<Separator />)`, matching the established browser-mode convention.

composer-for-yamada-ui (changes requested):
The migration lines up with the project's browser-mode pattern, but the `renders component correctly` test still has no assertion after migration. Other components already migrated to Vitest Browser Mode (for example `kbd`, `text`, and `box`) use `await a11y(...)` for this exact test slot, so this file drops that baseline a11y coverage. Please adopt the same pattern here.

packages/react/src/components/separator/separator.test.tsx
hirotomoyamada: Fixed.

hirotomoyamada: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6666, opened 2026-04-13
test(image): migrate to Vitest Browser Mode

### LINKED ISSUE #6513: test(image): migrate to Vitest Browser Mode
### Description

Migrate the `image` component tests to Vitest Browser Mode.

- [image.test.tsx](packages/react/src/components/image/image.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/image/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6513

## AI used

## Description

Migrate `image` component tests from jsdom (`#test`) to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

`packages/react/src/components/image/image.test.tsx` used `screen` assertions from `#test`.

## New behavior

No

## Additional Information

- Verified with: `pnpm react test:browser --run src/components/image/` (pass)
- Part of parent issue: #6460


packages/react/src/components/image/image.test.tsx
```diff
@@ -12,19 +12,21 @@ describe("<Image />", () => {
...
+    await expect.element(page.getByRole("img")).toHaveClass("ui-image")
   })
 
-  test("renders HTML tag correctly", () => {
-    render(<Image src={src} />)
-    expect(screen.getByRole("img").tagName).toBe("IMG")
+  test("renders HTML tag correctly", async () => {
+    await render(<Image src={src} />)
+
+    expect(page.getByRole("img").element().tagName).toBe("IMG")
```
codex-for-yamada-ui: **Issue**: This drops Vitest browser mode's built-in retry behavior because `.element()` is read synchronously and passed to plain `expect`. If the DOM commit lags slightly, the test can fail intermittently even when the component is correct.

**Fix**: Keep the assertion locator-based or runtime-narrowed: `const img = page.getByRole("img").element(); expect(img).toBeInstanceOf(HTMLImageElement)`.

codex-for-yamada-ui (changes requested):
In `image.test.tsx`, the `renders HTML tag correctly` assertion unwraps the locator with `.element()` before passing to plain `expect`. That bypasses Vitest browser mode's retry semantics, so the test can become flaky if the node is not attached at the instant `.element()` runs.

packages/react/src/components/image/image.test.tsx
composer-for-yamada-ui: **Issue**: The bare `expect(page.getByRole("img").element().tagName)` chain is inconsistent with the surrounding `await expect.element(page.getByRole("img"))` assertions and bypasses the locator's auto-wait/retry behavior.

**Fix**: Express the tag check via runtime narrowing instead of unwrapping into a sync read: `const img = page.getByRole("img").element(); expect(img).toBeInstanceOf(HTMLImageElement)`.

composer-for-yamada-ui (changes requested):
The `renders HTML tag correctly` case in `image.test.tsx` is the only DOM assertion that doesn't go through `await expect.element(page.getByRole("img"))`. Mixing the bare `.element().tagName` chain with the rest of the file's `expect.element(...)` style undermines the migration's stated goal and may bypass the auto-waiting behavior used elsewhere.

claude-for-yamada-ui (dismissed):

packages/react/src/components/image/image.test.tsx
hirotomoyamada: Fixed.

hirotomoyamada: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6673, opened 2026-04-13
test(card): migrate to Vitest Browser Mode

### LINKED ISSUE #6475: test(card): migrate to Vitest Browser Mode
### Description

Migrate the `card` component tests to Vitest Browser Mode.

- [card.test.tsx](packages/react/src/components/card/card.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/card/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6475

## AI used

## Description

Migrate `card` component tests from jsdom (`#test`) to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

`packages/react/src/components/card/card.test.tsx` used `screen` assertions from `#test`.

## New behavior

- Updated assertions to Browser Mode style using locators and async element checks
- Replaced role-based header/footer tag checks with direct text-element tag assertions for browser-stable coverage

No

## Additional Information

- Verified with: `pnpm react test:browser --run src/components/card/` (pass)
- Part of parent issue: #6460


packages/react/src/components/card/card.test.tsx
```diff
@@ -13,33 +13,39 @@ describe("<Card />", () => {
...
         <Card.Footer>Card Footer</Card.Footer>
       </Card.Root>,
     )
 
-    expect(screen.getByTestId("root").tagName).toBe("ARTICLE")
-    expect(screen.getByRole("banner").tagName).toBe("HEADER")
-    expect(screen.getByText("Card Body").tagName).toBe("DIV")
-    expect(screen.getByRole("contentinfo").tagName).toBe("FOOTER")
+    expect(page.getByTestId("root").element().tagName).toBe("ARTICLE")
+    expect(page.getByText("Card Header").element().tagName).toBe("HEADER")
```
codex-for-yamada-ui: **Issue**: `getByText("Card Header").element().tagName` targets the text node's containing element, not necessarily the `Card.Header` slot. Any harmless wrapper around the text would break this assertion even though the component still emits a `<header>`.

**Fix**: Add `data-testid` to each slot (`Card.Header data-testid="header"`, etc.) and assert that element's `tagName`, or assert the landmark roles directly (`page.getByRole("banner")` for the header, `getByRole("contentinfo")` for the footer).

codex-for-yamada-ui (changes requested):
In `card.test.tsx`, the migration uses `getByText(...)` to infer tag names for `Card.Header`, `Card.Body`, and `Card.Footer`. That locator binds to the text-bearing descendant rather than the slot element itself, so the test becomes brittle and can fail on harmless markup changes (e.g. wrapping the text in a span) while the component still renders the correct `header`/`div`/`footer`.

packages/react/src/components/card/card.test.tsx
```diff
@@ -13,33 +13,39 @@ describe("<Card />", () => {
...
-  test("renders HTML tag correctly", () => {
-    render(
+  test("renders HTML tag correctly", async () => {
+    await render(
       <Card.Root data-testid="root">
         <Card.Header>Card Header</Card.Header>
         <Card.Body>Card Body</Card.Body>
         <Card.Footer>Card Footer</Card.Footer>
       </Card.Root>,
     )
```
composer-for-yamada-ui, line 44: **Issue**: `getByText("Card Header")` / `getByText("Card Footer")` lose the previous landmark-role coverage (`getByRole("banner")` / `getByRole("contentinfo")`). A regression that drops the `<header>` / `<footer>` role would still pass.

**Fix**: Restore role-based queries for the header/footer, e.g. `await expect.element(page.getByRole("banner")).toBeInTheDocument()` and `await expect.element(page.getByRole("contentinfo")).toBeInTheDocument()`, alongside the tag-name checks.

composer-for-yamada-ui (changes requested):
The migrated `renders HTML tag correctly` assertions no longer query landmark roles. Previously the test used `getByRole("banner")` and `getByRole("contentinfo")`, verifying the header/footer are exposed as those landmarks (not just that some node with matching text renders as `HEADER`/`FOOTER`). The new `page.getByText(...)` checks only tag names via text and can pass even if landmark roles regress.

claude-for-yamada-ui (dismissed):

packages/react/src/components/card/card.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/card/card.test.tsx
hirotomoyamada: Fixed.

claude-for-yamada-ui (dismissed):

codex-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/card/card.test.tsx
hirotomoyamada: Follow-up: the `getByRole("banner")` / `getByRole("contentinfo")` queries you suggested fail in browser mode because `<header>` / `<footer>` only expose the `banner` / `contentinfo` landmark roles when they are direct children of `<body>`. Inside `<Card.Root>` (which renders `<article>`), the header/footer are not landmarks per the ARIA spec. Kept the per-slot `data-testid` + `tagName` poll pattern and removed the role-based assertions accordingly.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6682, opened 2026-04-13
test(fade): migrate tests to browser mode

### LINKED ISSUE #6496: test(fade): migrate to Vitest Browser Mode
### Description

Migrate the `fade` component tests to Vitest Browser Mode.

- [fade.test.tsx](packages/react/src/components/fade/fade.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/fade/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
## Description
- migrate `fade.test.tsx` from `#test` to `#test/browser`
- update interactions/assertions to browser-mode APIs with async `render` and `page`
- adapt visibility toggle assertions to computed opacity checks in real-browser rendering

## Related Issue
Closes #6496

## Breaking Change
No


claude-for-yamada-ui (dismissed):

packages/react/src/components/fade/fade.test.tsx
```diff
@@ -60,15 +68,17 @@ describe("<Fade />", () => {
...
+    expect(container.textContent).not.toContain("Fade")
 
     await user.click(button)
-    await waitFor(() => expect(screen.getByText("Fade")).toBeVisible())
+    await expect.element(page.getByText("Fade")).toBeVisible()
 
     await user.click(button)
-    await waitFor(() => expect(screen.queryByText("Fade")).toBeNull())
+    await vi.waitFor(() => {
+      expect(container.textContent).not.toContain("Fade")
```
codex-for-yamada-ui: **Issue:** The browser-mode rewrite no longer verifies the `unmountOnExit` contract directly. Checking `container.textContent` only proves that the literal text disappeared; it would still pass if `Fade` left an empty or otherwise textless subtree mounted after exit, which is exactly the regression this test is supposed to catch.

**Fix:** Replace the text-content check with a direct DOM-removal assertion, such as verifying `page.getByText("Fade").query()` or `document.querySelector(".ui-fade")` is `null` before opening and again after the close animation.

codex-for-yamada-ui (changes requested):
The browser-mode rewrite preserves the surface behaviour of the previous suite, but the `unmountOnExit` test loses fidelity: it only verifies that the literal text "Fade" disappears, which would still pass if Fade left an empty wrapper mounted after exit — the very regression this case exists to catch. Switch to a direct DOM-removal assertion to keep the contract under test.

packages/react/src/components/fade/fade.test.tsx
```diff
@@ -33,17 +36,22 @@ describe("<Fade />", () => {
...
     }
 
-    const { user } = render(<TestComponent />)
+    const { user } = await render(<TestComponent />)
 
-    const button = await screen.findByRole("button", { name: /button/i })
-    const collapse = await screen.findByText("Fade")
-    expect(collapse).not.toBeVisible()
+    const button = page.getByRole("button", { name: /button/i })
+    const collapse = page.getByText("Fade")
```
composer-for-yamada-ui: **Issue:** The locator is still bound to a variable named `collapse` even though this suite exercises `<Fade />`, not `<Collapse />`. The same block was edited for browser APIs, so this naming now reads like a mistaken copy from collapse/collapse-style tests and diverges from the clearer pattern used elsewhere (for example `fade-scale.test.tsx` uses `fadeScale` for the animated node).

**Fix:** Rename `collapse` to `fade` (or another Fade-specific name) and update `getOpacity` to close over that locator.

packages/react/src/components/fade/fade.test.tsx
```diff
@@ -1,5 +1,6 @@
-import { a11y, render, screen, waitFor } from "#test"
+import { a11y, page, render } from "#test/browser"
 import { useState } from "react"
+import { vi } from "vitest"
```
composer-for-yamada-ui: **Issue:** The browser Vitest project enables `globals: true` (see `packages/react/vitest.config.ts`), and other nearby `#test/browser` component suites such as `action-bar.test.tsx` and `accordion.test.tsx` call `vi.waitFor` without importing `vi`. Adding a standalone `import { vi } from "vitest"` is inconsistent with that local convention.

**Fix:** Remove the `vitest` import and rely on the configured Vitest globals for `vi`, matching the surrounding browser tests (unless a project rule explicitly requires explicit imports everywhere).

composer-for-yamada-ui (changes requested):
Two consistency issues stand out against the surrounding browser-mode tests in this directory. The locator variable retains the `collapse` name from a different component, and the explicit `import { vi } from "vitest"` is unnecessary because the browser Vitest project sets `globals: true` and adjacent suites such as `action-bar.test.tsx` rely on the global. Aligning with the local convention will keep the suite uniform.

packages/react/src/components/fade/fade.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/fade/fade.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/fade/fade.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6683, opened 2026-04-13
test(container): migrate tests to browser mode

### LINKED ISSUE #6487: test(container): migrate to Vitest Browser Mode
### Description

Migrate the `container` component tests to Vitest Browser Mode.

- [container.test.tsx](packages/react/src/components/container/container.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/container/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
## Description
- migrate `container.test.tsx` from `#test` to `#test/browser`
- update render usage to async browser-mode APIs with `page`-based assertions
- preserve existing coverage for class names and semantic tags across container slots

## Related Issue
Closes #6487

## Breaking Change
No


packages/react/src/components/container/container.test.tsx
```diff
@@ -75,9 +83,9 @@ describe("<Container />", () => {
       </Container.Root>,
     )
 
-    expect(screen.getByTestId("root").tagName).toBe("SECTION")
-    expect(screen.getByTestId("header").tagName).toBe("HEADER")
-    expect(screen.getByTestId("body").tagName).toBe("DIV")
-    expect(screen.getByTestId("footer").tagName).toBe("FOOTER")
+    expect(page.getByTestId("root").element().tagName).toBe("SECTION")
```
claude-for-yamada-ui: **Issue**: These four `expect(page.getByTestId(x).element().tagName).toBe(...)` reads bypass the `await expect.element(...)` retry path used by the sibling `className` assertions. If the locator is not yet attached when `.element()` runs, the test can fail for a reason unrelated to the rendered tag.

**Fix**: Wrap each tag check in a polling assertion, e.g. `await expect.poll(() => page.getByTestId("root").element().tagName).toBe("SECTION")`, and apply the same pattern to `header`, `body`, and `footer`.

claude-for-yamada-ui (changes requested):
In `container.test.tsx`, the `renders HTML tag correctly` test reads `tagName` via sync `.element()` access while every other DOM check in the file uses the `await expect.element(...)` retrying pattern. This is internally inconsistent and bypasses the auto-wait semantics that browser mode provides for stability.

packages/react/src/components/container/container.test.tsx
codex-for-yamada-ui: **Issue**: `.element()` is a non-retrying escape hatch in vitest browser mode, so wrapping it in a synchronous `expect(...)` drops the auto-wait behavior other tests in this file rely on. A late-attached locator would fail the assertion for a reason unrelated to the rendered tag.

**Fix**: Use a retriable path: `await expect.poll(() => page.getByTestId("root").element().tagName).toBe("SECTION")` and apply the same to `header`, `body`, and `footer`.

codex-for-yamada-ui (changes requested):
The semantic-tag checks in `container.test.tsx` use `page.getByTestId(...).element().tagName`, but `.element()` is documented as a strict escape hatch rather than a retriable assertion path. In browser mode this can fail immediately if the locator is not ready yet, turning these assertions flaky even when the rendered tag is correct. Keep them on the retriable path (e.g. `expect.poll`).

composer-for-yamada-ui (dismissed):

packages/react/src/components/container/container.test.tsx
hirotomoyamada: Fixed.

hirotomoyamada: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6713, opened 2026-04-13
feat(skills/create-issue): skip interactive collection when sufficient context exists

### LINKED ISSUE #6712: feat(skills/create-issue): skip interactive collection when sufficient context exists
### Description

The `create-issue` skill currently always goes through interactive steps — confirming issue type and collecting information field by field — even when the user has already provided all the necessary details in the command arguments or prior conversation messages.

When a user invokes `/create-issue` with a detailed argument, the skill still asks redundant questions, creating friction and slowing down the issue creation workflow. For example, if the argument already describes a bug with reproduction steps, expected/actual behavior, and environment, asking "Which template fits your need?" is unnecessary.

### Proposed Solution or API

Add a context-assessment step at the start of the workflow. Before entering the interactive collection loop, the skill evaluates whether:

- Issue type can be determined from the arguments/conversation
- A clear title and description can be constructed
- PR intent can be inferred

If all conditions are met, skip the interactive steps (template confirmation and field-by-field collection) and proceed directly to the duplicate-search and plan-presentation steps.

Inference rules for PR intent when not explicitly stated:
- User says "I'll fix it" / "I'll submit a PR" → `Yes`
- No mention → default to `No`

### Alternatives

None considered.

No

### AI used

bmthd:
Closes #6712

## AI used

## Description

Add a context-assessment step to the `create-issue` skill that detects when sufficient information is already available in the conversation and skips unnecessary interactive collection steps.

## Current behavior (updates)

The skill always goes through interactive steps — confirming issue type (step 3) and collecting information field by field (step 4) — even when the user has already provided all the necessary details in the command arguments or prior conversation messages.

## New behavior

A new step 3 evaluates whether the current conversation contains enough context to draft the issue without asking:

- Issue type can be determined from the arguments/conversation
- A clear title and description can be constructed
- PR intent can be inferred (explicit statement → use it; no mention → default to `No`)

If all conditions are met, the skill skips the interactive collection steps and proceeds directly to the duplicate-search and plan-presentation steps. If context is insufficient, it falls back to the original interactive flow.

No

## Additional Information

The original interactive steps (3a and 4) remain unchanged and are used as fallback when context is insufficient.


nano-nano (approved):
LGTM👍

.agents/skills/create-issue/SKILL.md
```diff
@@ -18,11 +18,20 @@ Use tools to interact with the user throughout the process.
...
+3. **Assess available context**
+   - Evaluate whether the current conversation (arguments, prior messages) already contains enough information to draft the issue without asking the user:
+     - Issue type can be determined (e.g., bug, feature, docs)
+     - A clear title and description can be constructed
+     - PR intent can be inferred (e.g., the user said "I'll fix it" → `Yes`; no mention → default to `No`)
+   - **If sufficient context exists:** skip steps 3a and 4, proceed directly to step 5 using the inferred values. Do not ask the user for information that is already present.
+   - **If context is insufficient:** follow steps 3a and 4 to collect the missing information interactively.
 
-4. **Collect information**
+3a. **Confirm issue type** _(skip if context is sufficient)_
```
odendayoko: @bmthd 
I think this should be a normal numbered step instead of `3a.`.

`3a` is a bit unusual as a Markdown workflow step, and references like `steps 3a and 4` make the numbering scheme less consistent for an agent to follow. Since this is a regular fallback step after `Assess available context`, it would be clearer to make it `4. Confirm issue type` and renumber the following steps accordingly.

odendayoko (changes requested):

.agents/skills/create-issue/SKILL.md
bmthd: 対応しました！

nano-nano (approved):

odendayoko (approved):

chida09 (approved):

hirotomoyamada (approved):

PR #6715, opened 2026-04-13
test(input): migrate tests to browser mode

### LINKED ISSUE #6516: test(input): migrate to Vitest Browser Mode
### Description

Migrate the `input` component tests to Vitest Browser Mode.

- [input.test.tsx](packages/react/src/components/input/input.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/input/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
## Description
- migrate `input.test.tsx` from `#test` to `#test/browser`
- convert assertions to browser-mode APIs (`page`, async `render`, and `expect.element`)
- keep existing a11y coverage while disabling browser-mode `color-contrast` rule for known contrast mismatch in test theme

## Related Issue
Closes #6516

## Breaking Change
No


packages/react/src/components/input/input.test.tsx
```diff
@@ -12,6 +12,7 @@ describe("<Input />", () => {
       {
         axeOptions: {
           rules: {
+            "color-contrast": { enabled: false },
```
codex-for-yamada-ui, line 18: **Issue:** The PR newly disables axe's `color-contrast` rule for the Input a11y test, which weakens accessibility regression coverage for this component without any in-file justification or linked constraint specific to Input.

**Fix:** Remove `"color-contrast": { enabled: false }` and keep contrast auditing enabled; if there is a known unavoidable theme-level false positive, document the reason explicitly in the test and scope the exception as narrowly as possible.

codex-for-yamada-ui (changes requested):
- Accessibility: The PR newly disables axe's `color-contrast` rule for the Input a11y test, which weakens accessibility regression coverage for this component without any in-file justification or linked constraint specific to Input.

packages/react/src/components/input/input.test.tsx
claude-for-yamada-ui, line 18: **Issue:** Disabling axe's `color-contrast` rule weakens a11y coverage for the Input component without an in-file justification. The existing precedent in `link.test.tsx` documents the same exception with a comment explaining it is a theme-level audit item, not a test bug.

**Fix:** Add a brief inline comment in the same style as `link.test.tsx` (lines 6-8) explaining that the contrast mismatch originates from the test theme rather than from the Input component itself, so future readers understand why the rule is intentionally disabled.

claude-for-yamada-ui (changes requested):
- Accessibility: Disabling axe's `color-contrast` rule weakens a11y coverage for the Input component without an in-file justification. The existing precedent in `link.test.tsx` documents the same exception with a comment explaining it is a theme-level audit item, not a test bug.

composer-for-yamada-ui (dismissed):

packages/react/src/components/input/input.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6716, opened 2026-04-13
test(skeleton): migrate tests to browser mode

### LINKED ISSUE #6550: test(skeleton): migrate to Vitest Browser Mode
### Description

Migrate the `skeleton` component tests to Vitest Browser Mode.

- [skeleton.test.tsx](packages/react/src/components/skeleton/skeleton.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/skeleton/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6550

## AI used

## Description

Migrate `skeleton` component tests from `#test` to `#test/browser` and adjust assertions to browser runtime behavior.

## Current behavior (updates)

`skeleton` tests relied on jsdom-style assumptions and did not run in browser mode.

## New behavior

`skeleton` tests run in browser mode and assert DOM/CSS variable behavior using browser-compatible checks.

No


claude-for-yamada-ui (dismissed):

packages/react/src/components/skeleton/skeleton.test.tsx
```diff
@@ -18,111 +18,123 @@ describe("<Skeleton />", () => {
...
       <Skeleton data-testid="skeleton" fitContent={false}>
         <span>Content</span>
       </Skeleton>,
     )
-    const el = screen.getByTestId("skeleton")
-    expect(el.style.getPropertyValue("--width")).toBe("")
-    expect(el.style.getPropertyValue("--height")).toBe("")
+    const styles = getComputedStyle(page.getByTestId("skeleton").element())
+    expect(styles.getPropertyValue("--width").trim()).not.toBe("fit-content")
+    expect(styles.getPropertyValue("--height").trim()).not.toBe("fit-content")
```
codex-for-yamada-ui: **Issue:** The negative `fitContent` assertions were weakened from checking that `--width`/`--height` are unset to only checking they are not `"fit-content"`. This would still pass if those variables are incorrectly set to some other value, so regressions in the `fitContent={false}` behavior can slip through.

**Fix:** Restore strict assertions that the custom properties are empty/unset (for both tests), e.g. `expect(styles.getPropertyValue("--width").trim()).toBe("")` and `expect(styles.getPropertyValue("--height").trim()).toBe("")`.

codex-for-yamada-ui (changes requested):
- Testing: The negative `fitContent` assertions were weakened from checking that `--width`/`--height` are unset to only checking they are not `"fit-content"`. This would still pass if those variables are incorrectly set to some other value, so regressions in the `fitContent={false}` behavior can slip through.

packages/react/src/components/skeleton/skeleton.test.tsx
```diff
@@ -18,111 +18,123 @@ describe("<Skeleton />", () => {
...
   })
```
composer-for-yamada-ui, line 139: **Issue:** The two negative `fitContent` cases no longer assert that `--width` and `--height` are unset on the skeleton element. Previously the tests used `el.style.getPropertyValue(...)` and expected `""`, matching `skeleton.tsx` where `fitContent` is false yields `undefined` for those custom properties (no specified value on that element). The PR now uses `getComputedStyle`, which resolves custom properties with inheritance, and replaces the strict empty-string check with `not.toBe("fit-content")`, which still passes if the element incorrectly gets other non-`fit-content` values (for example accidental pixel sizes).

**Fix:** Restore the original semantics: after `await render(...)`, assert on the element’s own specified styles, e.g. `page.getByTestId("skeleton").element().style.getPropertyValue("--width")` and the same for `--height`, expecting `""` (or the exact string your runtime emits when the prop is omitted), instead of `getComputedStyle` plus a negated `fit-content` check.

packages/react/src/components/skeleton/skeleton.test.tsx
```diff
@@ -18,111 +18,123 @@ describe("<Skeleton />", () => {
...
-    expect(screen.getByTestId("skeletonText").children[0]).toHaveClass(
+    await expect
+      .element(page.getByTestId("skeleton"))
+      .toHaveClass("ui-skeleton")
+    await expect
+      .element(page.getByTestId("skeletonCircle"))
+      .toHaveClass("ui-skeleton")
+    expect(page.getByTestId("skeletonText").element().children[0]).toHaveClass(
       "ui-skeleton",
     )
```
composer-for-yamada-ui: **Issue:** In the `sets className correctly` test, the first two class assertions use `await expect.element(page.getByTestId(...)).toHaveClass(...)`, but the `SkeletonText` line uses a synchronous `expect` on `page.getByTestId("skeletonText").element().children[0]`. That diverges from the Vitest browser pattern used immediately above and from other `#test/browser` tests that rely on `expect.element` with `page` locators.

**Fix:** Express the first skeleton line as a locator and assert with the same API, e.g. `await expect.element(page.getByTestId("skeletonText").locator("> div").first()).toHaveClass("ui-skeleton")` (or an equivalent stable child locator), so all three class checks use `await expect.element`.

composer-for-yamada-ui (changes requested):
- テスト: The two negative `fitContent` cases no longer assert that `--width` and `--height` are unset on the skeleton element. Previously the tests used `el.style.getPropertyValue(...)` and expected `""`, matching `skeleton.tsx` where `fitContent` is false yields `undefined` for those custom properties (no specified value on that element). The PR now uses `getComputedStyle`, which resolves custom properties with inheritance, and replaces the strict empty-string check with `not.toBe("fit-content")`, which still passes if the element incorrectly gets other non-`fit-content` values (for example accidental pixel sizes).
- 一貫性: In the `sets className correctly` test, the first two class assertions use `await expect.element(page.getByTestId(...)).toHaveClass(...)`, but the `SkeletonText` line uses a synchronous `expect` on `page.getByTestId("skeletonText").element().children[0]`. That diverges from the Vitest browser pattern used immediately above and from other `#test/browser` tests that rely on `expect.element` with `page` locators.

packages/react/src/components/skeleton/skeleton.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/skeleton/skeleton.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/skeleton/skeleton.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6723, opened 2026-04-13
test(dropzone): migrate tests to browser mode

### LINKED ISSUE #6491: test(dropzone): migrate to Vitest Browser Mode
### Description

Migrate the `dropzone` component tests to Vitest Browser Mode.

- [dropzone.test.tsx](packages/react/src/components/dropzone/dropzone.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/dropzone/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6491

## AI used

## Description

Migrate `dropzone` component tests from `#test` to `#test/browser`.

## Current behavior (updates)

`dropzone` tests were still using jsdom-based `#test` utilities.

## New behavior

Tests now run in browser mode with async `render`, `page`-based assertions, and browser-compatible drag/drop event dispatching.

No


packages/react/src/components/dropzone/dropzone.test.tsx
```diff
@@ -10,6 +10,13 @@ describe("<Dropzone />", () => {
       <Dropzone.Root>
         <Text>Drag file here or click to select file</Text>
       </Dropzone.Root>,
+      {
+        axeOptions: {
+          rules: {
+            "color-contrast": { enabled: false },
```
codex-for-yamada-ui: **Issue:** This change disables axe's `color-contrast` rule in the Dropzone accessibility test, which means the test no longer detects contrast regressions for this component.

**Fix:** Keep `color-contrast` enabled in `a11y(...)` and address the underlying contrast problem in the rendered Dropzone state (or adjust the fixture/theme setup so the component passes without suppressing the rule).

codex-for-yamada-ui (changes requested):
- Accessibility: This change disables axe's `color-contrast` rule in the Dropzone accessibility test, which means the test no longer detects contrast regressions for this component.

packages/react/src/components/dropzone/dropzone.test.tsx
composer-for-yamada-ui: **Issue:** The `a11y` call disables the axe `color-contrast` rule. A search of `packages/react` `*.test.tsx` shows this is the only place that turns off that rule, so this test no longer enforces the same color-contrast guarantee as the default `a11y` pattern used by other components, and the change does not record why that exception is required.

**Fix:** First try `await a11y(...)` with default options and fix an actual token/theme or fixture issue if a violation is reported. If a real or environment-specific false positive forces the exception, add the minimum JSDoc on the `a11y` call that the project allows, stating why `color-contrast` must be disabled; avoid broad rule suppression if a narrower `axeOptions` (as in `input.test.tsx` for a different rule) is enough.

composer-for-yamada-ui (changes requested):
- Accessibility: The `a11y` call disables the axe `color-contrast` rule. A search of `packages/react` `*.test.tsx` shows this is the only place that turns off that rule, so this test no longer enforces the same color-contrast guarantee as the default `a11y` pattern used by other components, and the change does not record why that exception is required.

claude-for-yamada-ui (dismissed):

packages/react/src/components/dropzone/dropzone.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

PR #6724, opened 2026-04-13
test(native-accordion): migrate tests to browser mode

### LINKED ISSUE #6524: test(native-accordion): migrate to Vitest Browser Mode
### Description

Migrate the `native-accordion` component tests to Vitest Browser Mode.

- [native-accordion.test.tsx](packages/react/src/components/native-accordion/native-accordion.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/native-accordion/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6524

## AI used

## Description

Migrate `native-accordion` component tests from `#test` to `#test/browser`.

## Current behavior (updates)

`native-accordion` tests were still using jsdom-based `#test` utilities.

## New behavior

Tests now run in browser mode with async `render` and `page`-based assertions. The multiple-open behavior test now explicitly enables `multiple` to match native details behavior in real browsers.

No


packages/react/src/components/native-accordion/native-accordion.test.tsx
```diff
@@ -100,24 +106,29 @@ describe("<NativeAccordion />", () => {
...
+    const item1 = button1.closest("details")
+    const item2 = button2.closest("details")
 
-    await user.click(screen.getByText(/Accordion Label 1/i))
-    await user.click(screen.getByText(/Accordion Label 2/i))
+    await user.click(button1)
+    await user.click(button2)
 
     expect(item1).toHaveAttribute("open")
     expect(item2).toHaveAttribute("open")
```
codex-for-yamada-ui, line 123: **Issue:** This hunk replaces the default-mode interaction check with a `multiple`-only check. `NativeAccordion` defaults to `multiple=false` (single-open behavior via shared `name`), but there is now no interaction assertion that default mode actually keeps only one item open, so regressions in the default behavior would not be detected.

**Fix:** Keep this `multiple` test and add a separate default-mode interaction test (render without `multiple`, click item 1 then item 2, assert item 1 closes and item 2 remains open).

codex-for-yamada-ui (changes requested):
- Test: This hunk replaces the default-mode interaction check with a `multiple`-only check. `NativeAccordion` defaults to `multiple=false` (single-open behavior via shared `name`), but there is now no interaction assertion that default mode actually keeps only one item open, so regressions in the default behavior would not be detected.

claude-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/native-accordion/native-accordion.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6728, opened 2026-04-14
test(notice): migrate tests to browser mode

### LINKED ISSUE #6528: test(notice): migrate to Vitest Browser Mode
### Description

Migrate the `notice` component tests to Vitest Browser Mode.

- [notice.test.tsx](packages/react/src/components/notice/notice.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/notice/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6528

## AI used

## Description

Migrate `notice` tests from `#test` to `#test/browser`.

## Current behavior (updates)

`notice` hook tests were still based on jsdom utilities (`act`, `screen`, `waitFor`).

## New behavior

Tests now run in browser mode with async `renderHook`, `page` interactions, and `expect.poll` assertions against rendered notice state.

No


packages/react/src/components/notice/notice.test.tsx
```diff
@@ -1,325 +1,241 @@
...
     })
 
-    act(() => {
-      screen.getByRole("button", { name: /close/i }).click()
-    })
+    await expectNoticeVisible("Closeable Button Notice")
+    await expect
+      .poll(() => !!document.querySelector("[data-close-button]"))
+      .toBe(true)
+    ;(page.getByRole("button").element() as HTMLElement).click()
```
claude-for-yamada-ui: **Issue:** The previous test resolved the close button via `getByRole("button", { name: /close/i })`, asserting both presence and accessible name (`aria-label="Close"`) of the `CloseButton`. The new code clicks `page.getByRole("button").element()` without a `name` filter and uses `[data-close-button]` (which is set on `NoticeContent`, not on the close button) as the only structural assertion. This drops the regression guard on the close button's accessible name.

**Fix:** Locate and click the close control via its accessible role+name, e.g. `(page.getByRole("button", { name: /close/i }).element() as HTMLElement).click()`, and assert that locator's presence/absence instead of the wrapper's `data-close-button` attribute.

packages/react/src/components/notice/notice.test.tsx
```diff
@@ -1,325 +1,241 @@
...
+      closable: false,
+      closeStrategy: "button",
+      duration: null,
+      title: "Non-closable Notice",
     })
+
+    await expectNoticeVisible("Non-closable Notice")
+    await expect
+      .poll(() => !!document.querySelector("[data-close-button]"))
+      .toBe(false)
```
claude-for-yamada-ui: **Issue:** The `closable: false` regression test previously asserted absence via `queryByRole("button", { name: /close/i })`. It now asserts only `!document.querySelector("[data-close-button]")`, which is a markup detail on `NoticeContent` rather than an absence-of-accessible-close-control check. A future change that keeps the data attribute but removes the named close button (or vice versa) would silently regress accessibility.

**Fix:** Restore the user-centric absence assertion, for example `await expect.poll(() => page.getByRole("button", { name: /close/i }).query()).toBeNull()`, optionally keeping the data-attribute check as a supplemental signal.

packages/react/src/components/notice/notice.test.tsx
```diff
@@ -1,325 +1,241 @@
-import { act, renderHook, screen, waitFor } from "#test"
+import { page, renderHook } from "#test/browser"
 import { useNotice } from "./use-notice"
 
+async function expectNoticeVisible(text: string) {
+  await expect.poll(() => document.body.textContent.includes(text)).toBe(true)
+}
+
+async function expectNoticeHidden(text: string) {
+  await expect.poll(() => document.body.textContent.includes(text)).toBe(false)
```
claude-for-yamada-ui: **Issue:** `expectNoticeVisible/expectNoticeHidden` rely on `document.body.textContent.includes(text)`, a global substring check across the entire document. This weakens what the assertion proves: a string elsewhere on the page (other notices, layout chrome, debug output) can satisfy or break the check independently of the notice element under test. It also breaks if `document.body.textContent` is `null`.

**Fix:** Anchor the assertion to the rendered notice node, e.g. `await expect.element(page.getByText(text)).toBeInTheDocument()` for presence and `await expect.poll(() => page.getByText(text).query()).toBeNull()` for absence; this keeps coverage element-scoped and null-safe.

claude-for-yamada-ui (changes requested):
- アクセシビリティ: The previous test resolved the close button via `getByRole("button", { name: /close/i })`, asserting both presence and accessible name (`aria-label="Close"`) of the `CloseButton`. The new code clicks `page.getByRole("button").element()` without a `name` filter and uses `[data-close-button]` (which is set on `NoticeContent`, not on the close button) as the only structural assertion. This drops the regression guard on the close button's accessible name.
- アクセシビリティ: The `closable: false` regression test previously asserted absence via `queryByRole("button", { name: /close/i })`. It now asserts only `!document.querySelector("[data-close-button]")`, which is a markup detail on `NoticeContent` rather than an absence-of-accessible-close-control check. A future change that keeps the data attribute but removes the named close button (or vice versa) would silently regress accessibility.
- テスト: `expectNoticeVisible/expectNoticeHidden` rely on `document.body.textContent.includes(text)`, a global substring check across the entire document. This weakens what the assertion proves: a string elsewhere on the page (other notices, layout chrome, debug output) can satisfy or break the check independently of the notice element under test. It also breaks if `document.body.textContent` is `null`.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: **Issue:** `expectNoticeVisible/expectNoticeHidden` now use `document.body.textContent.includes(...)`, which is a global substring check. This weakens the test contract from element-level verification to "the text exists somewhere in the page", so it can pass without proving that the notice node itself appeared/disappeared.

**Fix:** Use locator-based assertions tied to the rendered element (for example, `await expect.element(page.getByText(text)).toBeInTheDocument()` for presence and `await expect.poll(() => page.getByText(text).query()).toBeNull()` for absence).

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: **Issue:** The close-button test no longer verifies the accessible close control. It checks `[data-close-button]` (set on `NoticeContent`, not the button) and then clicks `page.getByRole("button")` without a name filter, so it stops guarding the `CloseButton` accessibility contract (`aria-label="Close"`).

**Fix:** Assert and interact with the close button via accessible selectors, e.g. `const close = page.getByRole("button", { name: /close/i })`, then verify and click that locator.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: **Issue:** The `closable: false` case now checks only `[data-close-button]` on the content wrapper, which does not directly prove that no close button is rendered. This removed the previous regression guard on close-button absence.

**Fix:** Restore a direct absence assertion for the close control, e.g. `expect(page.getByRole("button", { name: /close/i }).query()).toBeNull()` (you can keep the data-attribute check as supplemental).

codex-for-yamada-ui (changes requested):
- テスト: `expectNoticeVisible/expectNoticeHidden` now use `document.body.textContent.includes(...)`, which is a global substring check. This weakens the test contract from element-level verification to "the text exists somewhere in the page", so it can pass without proving that the notice node itself appeared/disappeared.
- アクセシビリティ: The close-button test no longer verifies the accessible close control. It checks `[data-close-button]` (set on `NoticeContent`, not the button) and then clicks `page.getByRole("button")` without a name filter, so it stops guarding the `CloseButton` accessibility contract (`aria-label="Close"`).
- テスト: The `closable: false` case now checks only `[data-close-button]` on the content wrapper, which does not directly prove that no close button is rendered. This removed the previous regression guard on close-button absence.

packages/react/src/components/notice/notice.test.tsx
composer-for-yamada-ui: **Issue:** The previous test (removed in this diff) closed the notice via `getByRole("button", { name: /close/i })`, so the test ensured the action targeted the control exposed with a "close" accessible name. The new line uses `page.getByRole("button")` with no `name` option, so that accessible-name check is no longer part of the test.

**Fix:** Use the browser `page` API with a role+name filter matching the old behavior, e.g. `page.getByRole("button", { name: /close/i })`, before calling `.click()` (adjust to whatever the Vitest `page` wrapper exposes, if the method shape differs).

packages/react/src/components/notice/notice.test.tsx
composer-for-yamada-ui: **Issue:** The prior test asserted that no close control with an accessible `close` name was present using `queryByRole("button", { name: /close/i })`. The new code replaces that with `document.querySelector("[data-close-button]")`. In the implementation, `data-close-button` is applied on the `NoticeContent` wrapper with `dataAttr(closable && closeStrategies.includes("button"))` (see `packages/react/src/components/notice/notice.tsx` around the `NoticeContent` props), not on the close button, so the test now keys off a markup detail instead of the same user-facing absence of a named close button.

**Fix:** Assert absence via the same accessible surface as before (e.g. that no `button` with `name` matching `/close/i` is in the document / tree), and drop or supplement the `data-close-button` check so regression coverage stays user-centric unless you have a specific reason to pin the `data-` attribute.

composer-for-yamada-ui (changes requested):
- Accessibility: The previous test (removed in this diff) closed the notice via `getByRole("button", { name: /close/i })`, so the test ensured the action targeted the control exposed with a "close" accessible name. The new line uses `page.getByRole("button")` with no `name` option, so that accessible-name check is no longer part of the test.
- Accessibility: The prior test asserted that no close control with an accessible `close` name was present using `queryByRole("button", { name: /close/i })`. The new code replaces that with `document.querySelector("[data-close-button]")`. In the implementation, `data-close-button` is applied on the `NoticeContent` wrapper with `dataAttr(closable && closeStrategies.includes("button"))` (see `packages/react/src/components/notice/notice.tsx` around the `NoticeContent` props), not on the close button, so the test now keys off a markup detail instead of the same user-facing absence of a named close button.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/notice/notice.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6731, opened 2026-04-14
test(pagination): migrate tests to browser mode

### LINKED ISSUE #6529: test(pagination): migrate to Vitest Browser Mode
### Description

Migrate the `pagination` component tests to Vitest Browser Mode.

- [pagination.test.tsx](packages/react/src/components/pagination/pagination.test.tsx)
- [use-pagination.test.ts](packages/react/src/components/pagination/use-pagination.test.ts)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/pagination/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6529

## AI used

## Description

Migrate `pagination` component and hook tests to Vitest Browser Mode.

## Current behavior (updates)

`pagination` tests were still written against jsdom `#test` helpers.

## New behavior

The tests now use browser-mode utilities (`#test/browser`, `page`, async `render`, and browser-compatible assertions/interactions) and pass in browser project runs.

No

## Additional Information

N/A


packages/react/src/components/pagination/pagination.test.tsx
```diff
@@ -163,127 +177,133 @@ describe("<Pagination />", () => {
...
 
-  test("all items are disabled when disabled prop is true", () => {
-    render(<Pagination.Root disabled total={10} />)
+  test("all items are disabled when disabled prop is true", async () => {
+    await render(<Pagination.Root disabled total={10} />)
 
-    const buttons = screen.getAllByRole("button")
+    const buttons = page.getByRole("button").elements()
     buttons.forEach((button) => {
       expect(button).toBeDisabled()
```
codex-for-yamada-ui, line 224: **Issue:** `page.getByRole("button").elements()` returns an empty array when nothing matches, so this test can pass without asserting anything if buttons are accidentally not rendered. The previous `getAllByRole("button")` behavior failed in that case, so this migration weakens regression detection.

**Fix:** Keep an explicit non-empty assertion before iterating (e.g. `expect(buttons.length).toBeGreaterThan(0)`), or switch to locator-based checks with `all()` plus `await expect.element(...)` for each button so the test still fails when no buttons are present.

codex-for-yamada-ui (changes requested):
- Test: `page.getByRole("button").elements()` returns an empty array when nothing matches, so this test can pass without asserting anything if buttons are accidentally not rendered. The previous `getAllByRole("button")` behavior failed in that case, so this migration weakens regression detection.

packages/react/src/components/pagination/pagination.test.tsx
```diff
@@ -163,127 +177,133 @@ describe("<Pagination />", () => {
...
     })
```
claude-for-yamada-ui, line 225: **Issue:** `page.getByRole("button").elements()` returns an empty array when no buttons are found, so `forEach` becomes a no-op and the assertion silently passes. Under the previous `screen.getAllByRole("button")` API the call would throw if no buttons were rendered, so this migration weakens regression detection for the "all items disabled" check.

**Fix:** Add a sanity assertion before iterating, e.g. `expect(buttons.length).toBeGreaterThan(0)`, or restructure with `await expect.element(button).toBeDisabled()` per locator using `.all()` so the test still fails when the disabled prop accidentally suppresses the entire pagination.

claude-for-yamada-ui (changes requested):
- Test: `page.getByRole("button").elements()` returns an empty array when no buttons are found, so `forEach` becomes a no-op and the assertion silently passes. Under the previous `screen.getAllByRole("button")` API the call would throw if no buttons were rendered, so this migration weakens regression detection for the "all items disabled" check.

composer-for-yamada-ui (dismissed):

packages/react/src/components/pagination/pagination.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/pagination.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6732, opened 2026-04-14
test(stack): migrate tests to browser mode

### LINKED ISSUE #6556: test(stack): migrate to Vitest Browser Mode
### Description

Migrate the `stack` component tests to Vitest Browser Mode.

- [h-stack.test.tsx](packages/react/src/components/stack/h-stack.test.tsx)
- [stack.test.tsx](packages/react/src/components/stack/stack.test.tsx)
- [v-stack.test.tsx](packages/react/src/components/stack/v-stack.test.tsx)

> Note: `z-stack.test.tsx` is already migrated and is out of scope for this issue.

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/stack/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6556

## AI used

## Description

Migrate stack-related tests to Vitest Browser Mode.

## Current behavior (updates)

`h-stack`, `stack`, and `v-stack` tests were still using jsdom `#test` utilities.

## New behavior

These tests now use browser-mode testing utilities (`#test/browser`, `page`, async `render`, and browser-safe interaction/assertion patterns).

No

## Additional Information

`z-stack.test.tsx` is already migrated and remains unchanged.


packages/react/src/components/stack/stack.test.tsx
```diff
@@ -84,16 +84,15 @@ describe("<Stack />", () => {
...
       <Stack>
         {data.map(({ name }) => (
           <Component key={name} name={name} />
         ))}
       </Stack>,
     )
 
-    const items = await screen.findAllByTestId("character")
-
+    const items = page.getByTestId("character").elements()
```
claude-for-yamada-ui: **Issue**: `await screen.findAllByTestId("character")` is replaced with the synchronous `page.getByTestId("character").elements()`. `findAllBy*` retries until matches appear; `.elements()` takes a one-shot snapshot, so `expect(items).toHaveLength(6)` may observe the DOM before all 6 children mount and race.

**Fix**: Use a retrying path, e.g. `await expect.poll(() => page.getByTestId("character").elements().length).toBe(6)` and then read `.elements()` if you need the array, or `await expect.element(page.getByTestId("character").nth(5)).toBeInTheDocument()` before snapshotting.

packages/react/src/components/stack/stack.test.tsx
```diff
@@ -121,20 +120,16 @@ describe("<Stack />", () => {
       )
     }
 
-    render(<Wrapper data={data} />)
+    const { user } = await render(<Wrapper data={data} />)
 
-    const items = await screen.findAllByTestId("character")
+    const items = page.getByTestId("character").elements()
```
claude-for-yamada-ui: **Issue**: Same race as line 95 — `items` is snapshotted immediately after `await render(...)` via `.elements()`, so `toHaveLength(6)` can run before all six characters have been mounted.

**Fix**: Poll the length or await a specific element's presence before calling `.elements()`, mirroring the fix at line 95.

packages/react/src/components/stack/stack.test.tsx
```diff
@@ -121,20 +120,16 @@ describe("<Stack />", () => {
...
-    fireEvent.click(onClick)
-
-    await waitFor(() => {
-      expect(unMountMock).toHaveBeenCalledExactlyOnceWith("孫悟空")
-    })
+    await user.click(page.getByTestId("delete-button"))
+    await expect.poll(() => unMountMock.mock.calls.length).toBe(1)
+    expect(unMountMock).toHaveBeenCalledExactlyOnceWith("孫悟空")
 
     expect(unMountMock).toHaveBeenCalledTimes(1)
```
claude-for-yamada-ui: **Issue**: `toHaveBeenCalledTimes(1)` here is redundant — the preceding `toHaveBeenCalledExactlyOnceWith("孫悟空")` on line 132 already asserts both the call count (exactly once) and the argument, so this line duplicates the count check for no extra coverage.

**Fix**: Drop line 134.

claude-for-yamada-ui (changes requested):
Two related concerns on `stack.test.tsx`. First, the migration replaces `await screen.findAllByTestId(...)` with a synchronous `page.getByTestId(...).elements()` snapshot in two places (lines 95 and 125). `findAllBy*` implicitly retries until matches appear, whereas `.elements()` captures the current DOM state right after `render` resolves, so the subsequent `toHaveLength(6)` can race if children mount in a later microtask. Please restore a retrying wait. Second, after `toHaveBeenCalledExactlyOnceWith("孫悟空")` on line 132 the trailing `toHaveBeenCalledTimes(1)` on line 134 is redundant — the exactly-once matcher already covers count plus argument.

packages/react/src/components/stack/stack.test.tsx
codex-for-yamada-ui: **Issue**: `page.getByTestId("character").elements()` takes a non-retrying snapshot in Vitest Browser Mode, so this migration drops the wait semantics that `screen.findAllByTestId("character")` previously provided. That makes the migrated test flaky if the list is not attached at the exact moment `.elements()` runs. The same regression is repeated at line 125 in the sibling test.

**Fix**: Keep the assertion on a retrying path — `await expect.poll(() => page.getByTestId("character").elements().length).toBe(6)` is one option, or use a locator assertion (`await expect.element(page.getByTestId("character").nth(5)).toBeInTheDocument()`) before snapshotting. Apply the same change at both occurrences.

codex-for-yamada-ui (changes requested):
In `stack.test.tsx`, the migration swaps `await screen.findAllByTestId("character")` for `page.getByTestId("character").elements()` at lines 95 and 125. `findAllBy*` retries until the list is present, whereas `.elements()` returns a non-retrying one-shot snapshot in Vitest Browser Mode — so the `toHaveLength(6)` assertion can observe the DOM before the children finish mounting and become flaky. Restore retry semantics at both call sites (e.g. `await expect.poll(() => page.getByTestId("character").elements().length).toBe(6)` or a locator assertion like `await expect.element(page.getByTestId("character").nth(5)).toBeInTheDocument()`).

packages/react/src/components/stack/stack.test.tsx
composer-for-yamada-ui: **Issue**: `toHaveBeenCalledTimes(1)` is redundant immediately after `toHaveBeenCalledExactlyOnceWith("孫悟空")` on line 132 — the exactly-once matcher already implies a call count of exactly one, so this line adds no coverage.

**Fix**: Delete line 134.

packages/react/src/components/stack/stack.test.tsx
composer-for-yamada-ui: **Issue**: `page.getByTestId("character").elements()` snapshots the current DOM immediately after `await render(...)`; the previous `await screen.findAllByTestId(...)` retried until the nodes were present. Without that retry, `toHaveLength(6)` can fail intermittently if child rendering is not yet settled on the tick that `.elements()` runs. The same pattern repeats at line 125.

**Fix**: Use a retrying assertion — `await expect.poll(() => page.getByTestId("character").elements().length).toBe(6)` is the closest direct equivalent, or await a specific locator (`expect.element(page.getByTestId("character").nth(5)).toBeInTheDocument()`) before snapshotting. Apply the change to both occurrences.

composer-for-yamada-ui (changes requested):
Two concerns on `stack.test.tsx`. The redundant `toHaveBeenCalledTimes(1)` on line 134 should be dropped because the preceding `toHaveBeenCalledExactlyOnceWith("孫悟空")` already asserts both count and argument. And `page.getByTestId("character").elements()` at lines 95 and 125 is a one-shot snapshot, not a retrying wait like the previous `findAllByTestId`, so the `toHaveLength(6)` check can race if the children mount after `render` resolves. Please poll or otherwise wait until six matches are attached before taking the snapshot.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

packages/react/src/components/stack/stack.test.tsx
hirotomoyamada: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6734, opened 2026-04-14
test(pin-input): migrate tests to browser mode

### LINKED ISSUE #6531: test(pin-input): migrate to Vitest Browser Mode
### Description

Migrate the `pin-input` component tests to Vitest Browser Mode.

- [pin-input.test.tsx](packages/react/src/components/pin-input/pin-input.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/pin-input/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6531

## AI used

## Description

Migrate `pin-input` tests to Vitest Browser Mode.

## Current behavior (updates)

`pin-input.test.tsx` depended on jsdom-oriented helpers (`#test`, `screen`, `waitFor`, `act`).

## New behavior

The test file now uses browser-mode helpers (`#test/browser`, `page`, async `render`, and `expect.poll`) while preserving existing behavioral coverage for typing, focus movement, and paste-like input updates.

No

## Additional Information

N/A


packages/react/src/components/pin-input/pin-input.test.tsx
```diff
@@ -136,186 +138,134 @@ describe("<PinInput />", () => {
...
-    })
-
-    await waitFor(() => {
-      inputs.forEach((input, index) => {
-        expect(input).toHaveValue(testValue[index])
-      })
-    })
+    await render(<PinInput.Root />)
+    const [firstInput] = getInputs()
+    inputText(firstInput, testValue)
```
codex-for-yamada-ui: **Issue:** The paste-focused tests now bypass `user.paste(...)` and directly mutate DOM state (`input.value` + `dispatchEvent("input")`). That no longer validates the real clipboard-driven user interaction path, so paste regressions can be missed while these tests still pass.

**Fix:** Use `user.paste(...)` on the focused input in the paste tests, then assert the resulting values.

packages/react/src/components/pin-input/pin-input.test.tsx
```diff
@@ -136,186 +138,134 @@ describe("<PinInput />", () => {
...
+
+    await expect
+      .poll(() =>
+        getInputs()
+          .map((input) => input.value)
+          .join(""),
+      )
```
codex-for-yamada-ui: **Issue:** The `> 2 characters` case now checks only the concatenated value of all fields. This weakens coverage because wrong per-field distribution (for example, multiple characters in one field) can still satisfy the joined-string assertion.

**Fix:** Restore per-field assertions by index (each field has the expected character), not only the aggregate joined string.

codex-for-yamada-ui (changes requested):
- Testing: The paste-focused tests now bypass `user.paste(...)` and directly mutate DOM state (`input.value` + `dispatchEvent("input")`). That no longer validates the real clipboard-driven user interaction path, so paste regressions can be missed while these tests still pass.
- Testing: The `> 2 characters` case now checks only the concatenated value of all fields. This weakens coverage because wrong per-field distribution (for example, multiple characters in one field) can still satisfy the joined-string assertion.

packages/react/src/components/pin-input/pin-input.test.tsx
```diff
@@ -136,186 +138,134 @@ describe("<PinInput />", () => {
...
+      .toBe(testValue)
```
composer-for-yamada-ui: **Issue:** The assertion only checks the concatenation of all field values, so it no longer guarantees each `input` holds `testValue[index]`. A bug that keeps all characters in the first field would still produce joined `"1234"`, which the old per-index `toHaveValue` checks would have caught.

**Fix:** Inside `expect.poll` (or equivalent), assert for each index that `getInputs()[index]?.value` equals `testValue[index]` (or compare arrays), restoring the stricter per-cell guarantee.

packages/react/src/components/pin-input/pin-input.test.tsx
```diff
@@ -136,186 +138,134 @@ describe("<PinInput />", () => {
...
     const lastInput = inputs[3]
 
-    await act(async () => {
-      await user.click(lastInput!)
-      await user.keyboard("[Backspace]")
-    })
+    lastInput?.focus()
+    lastInput?.dispatchEvent(
+      new KeyboardEvent("keydown", { key: "Backspace", bubbles: true }),
+    )
```
composer-for-yamada-ui: **Issue:** `browser-testing.md` for `#test/browser` requires driving keyboard behavior through the `user` API instead of `focus()` plus raw `new KeyboardEvent("keydown", { key: "Backspace" })`, which bypasses the real input pipeline. The same pattern is repeated in this diff for the `manageFocus is false` case around lines 182-185, so it diverges from the documented test style and from other tests in this file that use `user`.

**Fix:** Focus with `user` (e.g. `await user.click(lastInput!)`) and send Backspace with `await user.keyboard("{Backspace}")` (or the equivalent supported sequence), and apply the same approach to the `manageFocus={false` block.

packages/react/src/components/pin-input/pin-input.test.tsx
```diff
@@ -1,6 +1,15 @@
...
+import { a11y, page, render } from "#test/browser"
 import { PinInput } from "."
 
+const getInputs = () =>
+  page.getByRole("textbox").elements() as HTMLInputElement[]
+const inputText = (input: HTMLInputElement | undefined, value: string) => {
+  if (!input) return
+
+  input.value = value
+  input.dispatchEvent(new Event("input", { bubbles: true }))
```
composer-for-yamada-ui: **Issue:** `inputText` mutates `input.value` and dispatches a single synthetic `input` event, which does not follow the “interactions through `user` / resemble real users” direction in `browser-testing.md`, and is weaker than the previous `user.paste` / `user.type` flows for exercising real browser input behavior.

**Fix:** Use `const { user } = await render(...)` and drive the same scenarios with `user` (e.g. `user.paste` on the first field, or `user.type` as appropriate) instead of the manual DOM + event shortcut.

composer-for-yamada-ui (changes requested):
- 正確性: The assertion only checks the concatenation of all field values, so it no longer guarantees each `input` holds `testValue[index]`. A bug that keeps all characters in the first field would still produce joined `"1234"`, which the old per-index `toHaveValue` checks would have caught.
- 一貫性: `browser-testing.md` for `#test/browser` requires driving keyboard behavior through the `user` API instead of `focus()` plus raw `new KeyboardEvent("keydown", { key: "Backspace" })`, which bypasses the real input pipeline. The same pattern is repeated in this diff for the `manageFocus is false` case around lines 182-185, so it diverges from the documented test style and from other tests in this file that use `user`.
- テスト: `inputText` mutates `input.value` and dispatches a single synthetic `input` event, which does not follow the “interactions through `user` / resemble real users” direction in `browser-testing.md`, and is weaker than the previous `user.paste` / `user.type` flows for exercising real browser input behavior.

packages/react/src/components/pin-input/pin-input.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/pin-input/pin-input.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/pin-input/pin-input.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/pin-input/pin-input.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/pin-input/pin-input.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

PR #6743, opened 2026-04-14
test(native-popover): migrate tests to browser mode

### LINKED ISSUE #6525: test(native-popover): migrate to Vitest Browser Mode
### Description

Migrate the `native-popover` component tests to Vitest Browser Mode.

- [native-popover.test.tsx](packages/react/src/components/native-popover/native-popover.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/native-popover/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6525

## AI used

## Description

Migrate `native-popover` tests to Vitest browser mode.

## Current behavior (updates)

`native-popover.test.tsx` used jsdom-style test imports and sync render patterns.

## New behavior

- Switch to `#test/browser` render and a11y setup.
- Update tests to async render flow for browser mode.
- Stabilize popover mode/target-action assertions by checking latest rendered instance.

No.

## Additional Information

- Ran: `pnpm react test:browser --run src/components/native-popover/`


packages/react/src/components/native-popover/native-popover.test.tsx
```diff
@@ -137,34 +138,35 @@ describe("<NativePopover />", () => {
...
-    rerender(<Component popoverTargetAction="hide" />)
+    await render(<Component popoverTargetAction="hide" />)
 
-    expect(screen.getByTestId("trigger")).toHaveAttribute(
-      "popoverTargetAction",
-      "hide",
-    )
+    const triggers = screen.getAllByTestId("trigger")
+
+    expect(triggers.at(-1)).toHaveAttribute("popoverTargetAction", "hide")
```
codex-for-yamada-ui: **Issue:** Both prop-variation tests replaced `rerender` with a second `render` call and then assert `getAllByTestId(...).at(-1)`. This no longer verifies prop updates on the same mounted instance; it mounts an additional tree and tolerates duplicate test IDs, which weakens regression detection compared with the original test intent.

**Fix:** Keep a single mounted instance per test by using `const { rerender } = await render(...)` and then `await rerender(...)`, and assert with `screen.getByTestId(...)` after rerender.

codex-for-yamada-ui (changes requested):
- Test: Both prop-variation tests replaced `rerender` with a second `render` call and then assert `getAllByTestId(...).at(-1)`. This no longer verifies prop updates on the same mounted instance; it mounts an additional tree and tolerates duplicate test IDs, which weakens regression detection compared with the original test intent.

packages/react/src/components/native-popover/native-popover.test.tsx
```diff
@@ -1,4 +1,5 @@
-import { a11y, fireEvent, render, screen } from "#test"
+import { a11y, render } from "#test/browser"
+import { fireEvent, screen } from "@testing-library/react"
```
composer-for-yamada-ui: **Issue:** This file is the only browser test under `packages/react` that imports `screen` (and the interaction helpers) from `@testing-library/react` instead of using `page` from `#test/browser`. Other component browser tests in this package (e.g. `number-input.test.tsx`, `wrap.test.tsx`) use `import { a11y, page, render } from "#test/browser"` and `page` locators with `await expect.element(...)` for assertions, so the new split import diverges from the established browser test pattern in the same monorepo.

**Fix:** Import `page` from `#test/browser` alongside `a11y` and `render`, remove the `@testing-library/react` import, replace `screen.getByTestId` / `getAllByTestId` with the corresponding `page` queries, and use `await expect.element(...).toHaveClass` / `toHaveAttribute` (and other matchers) the same way as in other `*.test.tsx` files that use `#test/browser`, so the file matches the project’s browser testing conventions.

packages/react/src/components/native-popover/native-popover.test.tsx
composer-for-yamada-ui: **Issue:** The two tests for changing `popover` and `popoverTargetAction` no longer use `rerender` on a single instance; they call `await render` twice in the same test and then assert on `getAllByTestId(...).at(-1)`, which leaves two full trees mounted and encodes a dependency on document order. Elsewhere, browser tests that need prop updates on the same mounted UI use the async `render` result: `const { rerender } = await render(...)` and `await rerender(<... />)` (e.g. `number-input.test.tsx:558-568`), which keeps one tree and a clearer, closer equivalent to the previous jsdom `rerender` behavior.

**Fix:** In those two tests, use `const { rerender } = await render(<Component ...first props />)`, keep querying the single node with `getByTestId` (or `page` if aligned with Finding 1), then `await rerender(<Component ...second props />)` and assert on the same locator without `getAllByTestId` and `.at(-1)`.

composer-for-yamada-ui (changes requested):
- Consistency, Readability & Maintainability: This file is the only browser test under `packages/react` that imports `screen` (and the interaction helpers) from `@testing-library/react` instead of using `page` from `#test/browser`. Other component browser tests in this package (e.g. `number-input.test.tsx`, `wrap.test.tsx`) use `import { a11y, page, render } from "#test/browser"` and `page` locators with `await expect.element(...)` for assertions, so the new split import diverges from the established browser test pattern in the same monorepo.
- Consistency, Correctness, Readability & Maintainability: The two tests for changing `popover` and `popoverTargetAction` no longer use `rerender` on a single instance; they call `await render` twice in the same test and then assert on `getAllByTestId(...).at(-1)`, which leaves two full trees mounted and encodes a dependency on document order. Elsewhere, browser tests that need prop updates on the same mounted UI use the async `render` result: `const { rerender } = await render(...)` and `await rerender(<... />)` (e.g. `number-input.test.tsx:558-568`), which keeps one tree and a clearer, closer equivalent to the previous jsdom `rerender` behavior.

claude-for-yamada-ui (dismissed):

packages/react/src/components/native-popover/native-popover.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/native-popover/native-popover.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/native-popover/native-popover.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6750, opened 2026-04-14
test(button): migrate tests to browser mode

### LINKED ISSUE #6473: test(button): migrate to Vitest Browser Mode
### Description

Migrate the `button` component tests to Vitest Browser Mode.

- [button-group.test.tsx](packages/react/src/components/button/button-group.test.tsx)
- [button.test.tsx](packages/react/src/components/button/button.test.tsx)
- [icon-button.test.tsx](packages/react/src/components/button/icon-button.test.tsx)

### Proposed Solution or API

3. Verify the migrated tests pass with `pnpm react test:browser --run src/components/button/`.

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #5285

hirotomoyamada:
Closes #6473

## AI used

## Description

Migrate button-related tests from jsdom (`#test`) to Vitest Browser Mode (`#test/browser`).

## Current behavior (updates)

- The following test files still used jsdom test utilities:
  - `button.test.tsx`
  - `button-group.test.tsx`
  - `icon-button.test.tsx`

## New behavior

- Updated all three files to Browser Mode imports and async `render` flow.
- Stabilized Browser Mode assertions by using `screen` queries for DOM element checks and by preserving test IDs where rerender assertions depend on them.
- Verified with:
  - `pnpm react test:browser --run src/components/button/`

No

## Additional Information

N/A


packages/react/src/components/button/button.test.tsx
```diff
@@ -11,26 +12,28 @@ describe("<Button />", () => {
...
     expect(screen.getByText("test").tagName).toBe("BUTTON")
   })
 
-  test("renders with icon", () => {
-    const { getByText, rerender } = render(<Button endIcon={<>end icon</>} />)
+  test("renders with icon", async () => {
+    const { getByText, rerender } = await render(
+      <Button endIcon={<>end icon</>} />,
+    )
     expect(getByText("end icon")).toBeTruthy()
```
claude-for-yamada-ui: **Issue:** `getByText` here is destructured from `vitest-browser-react`'s `render()` return value, which implements `LocatorSelectors` and returns a Playwright-style `Locator` object — not a DOM element like `@testing-library/react`'s `getByText`. A `Locator` is always a truthy object whether or not the element actually exists in the DOM, so `expect(getByText("end icon")).toBeTruthy()` on line 29 and `expect(getByText("start icon")).toBeTruthy()` on line 32 never fail and no longer verify that the icon rendered. The same regression exists at line 30 (`end icon`) and was introduced by the migration away from jsdom's element-returning API. The author's `pnpm react test:browser --run src/components/button/` smoke pass does not catch this because the assertion is a tautology.

**Fix:** Use an assertion that actually queries the DOM for Browser Mode locators, e.g. `await expect.element(getByText("end icon")).toBeInTheDocument()` (the same idiom used in `kbd.test.tsx`, `card.test.tsx`, `separator.test.tsx`), or switch this test to `screen.getByText(...)` to stay consistent with the rest of `button.test.tsx`.

packages/react/src/components/button/button.test.tsx
```diff
@@ -1,4 +1,5 @@
-import { a11y, render, screen } from "#test"
+import { a11y, render } from "#test/browser"
+import { screen } from "@testing-library/react"
```
claude-for-yamada-ui: **Issue:** Every other browser-mode test already migrated in this codebase (`packages/react/src/components/kbd/kbd.test.tsx`, `card/card.test.tsx`, `progress/progress.test.tsx`, `separator/separator.test.tsx`, and the just-merged `color-selector/color-selector.test.tsx` from PR #6917) uses `page` from `#test/browser` together with `await expect.element(...)`. This PR instead introduces `import { screen } from "@testing-library/react"` and sprinkles `screen.getByRole/getByTestId/getByText` through all three button test files, establishing a second, parallel pattern in the button package with no rationale in the PR description. Browser Mode's idiomatic API is `page` locators + `expect.element`; `screen` from `@testing-library/react` keeps the jsdom-style synchronous-DOM mental model and bypasses Browser Mode's retry/wait semantics, which is the whole point of the migration.

**Fix:** Drop the `@testing-library/react` import and rewrite the queries with `page.getByRole(...)` / `page.getByTestId(...)` / `page.getByText(...)` plus `await expect.element(...)`, matching the already-migrated files listed above. Apply the same change in `button-group.test.tsx:1-3` and `icon-button.test.tsx:1-3`.

packages/react/src/components/button/button.test.tsx
```diff
@@ -69,61 +72,65 @@ describe("<Button />", () => {
...
       </Button>,
     )
-    expect(getByTestId("btn")).not.toHaveAttribute("type")
+    expect(screen.getByTestId("btn")).not.toHaveAttribute("type")
   })
 
-  test("should be disabled", () => {
-    const { getByRole, getByTestId, rerender } = render(
+  test("should be disabled", async () => {
+    const { getByRole, getByTestId, rerender } = await render(
```
claude-for-yamada-ui: **Issue:** In "should be disabled" (line 124-141) the test still destructures `{ getByRole, getByTestId, rerender }` from the Browser Mode `render()` return value. As in Finding 1, these are Locators, not DOM elements. `expect(button).toBeDisabled()` (line 131) and `expect(buttonAsDiv).toHaveAttribute("disabled")` (line 140) only pass because Browser Mode extends `expect` for locators — but `const button = getByRole("button")` is captured once before the rerender (line 130) and then the rerender swaps the element to `as="div"`. Keeping the stale locator around and mixing it with `getByTestId` after the rerender is fragile and also inconsistent with the other tests in this same file which switched to `screen.getBy*` precisely to requery after each rerender (see lines 76-97 and 100-113).

**Fix:** Requery the element after `rerender` via `page.getByTestId("btn")` / `page.getByRole(...)` (or `screen.getBy*` if you keep that API), and use `await expect.element(...)` for the disabled/attribute assertions. Do not reuse a Locator captured before a rerender across subsequent assertions.

packages/react/src/components/button/button.test.tsx
claude-for-yamada-ui: **Issue:** The import block differs between the three files touched by this PR. `button-group.test.tsx:1-3` and `icon-button.test.tsx:1-3` place `@testing-library/react` first, then an empty line, then `#test/browser`; `button.test.tsx:1-2` has them in the opposite order with no blank line. All three files use the same import grouping, so the ordering should be identical.

**Fix:** Pick one ordering (matching whatever `oxlint` / the already-migrated browser-mode tests produce) and apply it consistently across the three files.

claude-for-yamada-ui (changes requested):
- 正確性: `getByText` here is destructured from `vitest-browser-react`'s `render()` return value, which implements `LocatorSelectors` and returns a Playwright-style `Locator` object — not a DOM element like `@testing-library/react`'s `getByText`. A `Locator` is always a truthy object whether or not the element actually exists in the DOM, so `expect(getByText("end icon")).toBeTruthy()` on line 29 and `expect(getByText("start icon")).toBeTruthy()` on line 32 never fail and no longer verify that the icon rendered. The same regression exists at line 30 (`end icon`) and was introduced by the migration away from jsdom's element-returning API. The author's `pnpm react test:browser --run src/components/button/` smoke pass does not catch this because the assertion is a tautology.
- 一貫性: Every other browser-mode test already migrated in this codebase (`packages/react/src/components/kbd/kbd.test.tsx`, `card/card.test.tsx`, `progress/progress.test.tsx`, `separator/separator.test.tsx`, and the just-merged `color-selector/color-selector.test.tsx` from PR #6917) uses `page` from `#test/browser` together with `await expect.element(...)`. This PR instead introduces `import { screen } from "@testing-library/react"` and sprinkles `screen.getByRole/getByTestId/getByText` through all three button test files, establishing a second, parallel pattern in the button package with no rationale in the PR description. Browser Mode's idiomatic API is `page` locators + `expect.element`; `screen` from `@testing-library/react` keeps the jsdom-style synchronous-DOM mental model and bypasses Browser Mode's retry/wait semantics, which is the whole point of the migration.
- 正確性: In "should be disabled" (line 124-141) the test still destructures `{ getByRole, getByTestId, rerender }` from the Browser Mode `render()` return value. As in Finding 1, these are Locators, not DOM elements. `expect(button).toBeDisabled()` (line 131) and `expect(buttonAsDiv).toHaveAttribute("disabled")` (line 140) only pass because Browser Mode extends `expect` for locators — but `const button = getByRole("button")` is captured once before the rerender (line 130) and then the rerender swaps the element to `as="div"`. Keeping the stale locator around and mixing it with `getByTestId` after the rerender is fragile and also inconsistent with the other tests in this same file which switched to `screen.getBy*` precisely to requery after each rerender (see lines 76-97 and 100-113).
- 一貫性: The import block differs between the three files touched by this PR. `button-group.test.tsx:1-3` and `icon-button.test.tsx:1-3` place `@testing-library/react` first, then an empty line, then `#test/browser`; `button.test.tsx:1-2` has them in the opposite order with no blank line. All three files use the same import grouping, so the ordering should be identical.

packages/react/src/components/button/button.test.tsx
codex-for-yamada-ui: **Issue:** The import groups are in the wrong order for the repo’s enforced `perfectionist/sort-imports` rule (`value-external` must come before `value-internal`). `#test/browser` is placed before `@testing-library/react`, unlike the other updated button test files, so this will fail lint in CI.

**Fix:** Move `import { screen } from "@testing-library/react"` above `import { a11y, render } from "#test/browser"` and keep the group separation consistent.

codex-for-yamada-ui (changes requested):
- Scope and Conventions: The import groups are in the wrong order for the repo’s enforced `perfectionist/sort-imports` rule (`value-external` must come before `value-internal`). `#test/browser` is placed before `@testing-library/react`, unlike the other updated button test files, so this will fail lint in CI.

packages/react/src/components/button/button.test.tsx
composer-for-yamada-ui: **Issue:** This PR is the only place under `packages/react` where `screen` is imported directly from `@testing-library/react` in `*.test.tsx` files, while the rest of the browser-migrated component tests in this package use `page` from `#test/browser` and `await expect.element(...)` / `locator.element()` (e.g. `src/components/flex/flex.test.tsx`, `src/components/badge/badge.test.tsx`), so the same PR introduces a second, divergent query/assertion style in the same package.

**Fix:** Refactor all three changed files to import and use `page` from `#test/browser`, drop the `screen` import from `@testing-library/react`, and port assertions to the same `page` + `expect.element` pattern as other browser component tests, including `queryBy*` cases (e.g. loading text) using the appropriate `page`/Vitest browser absence assertions, and `tagName` checks via the locator’s `.element()` if required by the runner API, so these tests follow one convention end-to-end.

composer-for-yamada-ui (changes requested):
- 一貫性: This PR is the only place under `packages/react` where `screen` is imported directly from `@testing-library/react` in `*.test.tsx` files, while the rest of the browser-migrated component tests in this package use `page` from `#test/browser` and `await expect.element(...)` / `locator.element()` (e.g. `src/components/flex/flex.test.tsx`, `src/components/badge/badge.test.tsx`), so the same PR introduces a second, divergent query/assertion style in the same package.

packages/react/src/components/button/button.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/button/button.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/button/button.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6793, opened 2026-04-17
Security: Catch-all fallback re-executes module code with `node-eval`

tomaioo:
## Summary

Security: Catch-all fallback re-executes module code with `node-eval`

## Problem

**Severity**: `Medium` | **File**: `packages/cli/src/utils/module.ts:L22`

Any error during the `vm` execution path triggers a catch block that re-evaluates the same code using `node-eval`. This means malicious or faulty code can execute once in the VM path and then execute again in the fallback path, amplifying side effects and making exploit behavior more reliable.

## Solution

Do not re-evaluate on generic errors. Only fallback for narrowly scoped, validated error types (if absolutely necessary), and never execute the same code twice. Prefer failing closed with explicit error reporting.

## Changes

- `packages/cli/src/utils/module.ts` (modified)

### DIFF
--- a/packages/cli/src/utils/module.ts
+++ b/packages/cli/src/utils/module.ts
@@ -1,6 +1,5 @@
 import { isObject } from "@yamada-ui/utils"
 import { build } from "esbuild"
-import nodeEval from "node-eval"
 import { realpathSync } from "node:fs"
 import { Script } from "node:vm"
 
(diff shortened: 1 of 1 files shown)


hirotomoyamada:
Hi, we are closing this PR because we have determined that it violates our AI usage policy.

Also, regarding security, this process is acceptable because the source referenced by this script is its own source.

PR #6840, opened 2026-04-18
feat(oxfmt): migrate to oxfmt

### LINKED ISSUE #6803: Migrate project's Prettier to oxfmt
## Description

Migrate the project's own formatter from Prettier to oxfmt (Oxc). This is about Yamada UI's internal dev tooling, not the `@yamada-ui/cli` end-user integration tracked in #5993/#5994.

- Large format-time speedups via the Rust-based Oxc toolchain.
- oxfmt Beta (released 2026-02) claims 100% Prettier JS/TS compatibility, and this repo uses zero Prettier plugins, keeping the migration surface small.
- Aligns internal dev tooling with the Oxc direction already scoped for the CLI (#5993, #5994).

## Proposed Solution or API

- Replace `.prettierrc.js` and `packages/workspace/src/prettier/config.js` with an `oxfmt` config.
- Update `lefthook.yaml`, `.github/workflows/quality.yml`, and `.vscode/settings.json` to use oxfmt for format-on-save / CI.
- Remove `eslint-config-prettier` once Prettier is fully dropped.

## Known constraints and concerns (from preliminary investigation)

- oxfmt is still Beta; whole-repo reformat diffs must be reviewed carefully before landing.
- Current `printWidth: 80` must be set explicitly (oxfmt default is 100).
- Compatibility of `jsxSingleQuote` and `bracketSpacing` options with oxfmt needs verification.
- Markdown / MDX / YAML / JSON / CSS are all covered by oxfmt Beta, so no coverage gap for this repo.
- Rollback path: oxfmt → Prettier may require a second reformat PR.

## Alternatives

- Stay on Prettier.
- Hybrid: format TS/TSX with oxfmt, keep Prettier for other file types during the transition.

## AI used

Related #5994
Related #5993

taroj1205:
Closes #6803

## AI used

## Description

Migrate Yamada UI's internal formatting workflow from Prettier to oxfmt.

## Current behavior (updates)

Formatting checks and write flows were based on Prettier across package scripts, hooks, and CI.

## New behavior

- Add and use root `.oxfmtrc.json`.
- Replace formatter scripts in packages/workspaces/playgrounds with oxfmt-based wiring.
- Update `lefthook` formatter step to run oxfmt on staged files.
- Update Quality workflow formatter job for oxfmt file matching and execution.
- Remove `.prettierrc.js` and `.prettierignore` from root.
- Update related workspace/turbo inputs and catalog entries for oxfmt.

No.

## Additional Information

This migration is scoped to repository internal tooling (not CLI integration scope tracked separately).


hirotomoyamada:
@coderabbitai resolve

packages/workspace/src/oxfmt/config.ts
```diff
@@ -0,0 +1,23 @@
...
+    ".turbo",
+    ".next",
+    ".velite",
+    "dist",
+    "trace",
+    "node_modules",
+    "coverage",
+    "storybook-static",
+    "routeTree.gen.ts",
+  ],
```
claude-for-yamada-ui, line 15: **Issue:** The `ignorePatterns` list omits `pnpm-lock.yaml`, which was present in the removed `.prettierignore`. Because the root `format` script is `oxfmt --check` and the repo contains `*.yaml`/`*.yml` within scope, oxfmt will now be able to pick up `pnpm-lock.yaml` — a machine-generated file that was explicitly excluded before. At minimum, this risks format-check noise or wasted CI time on a multi-megabyte lockfile on every dependency update.

**Fix:** Add `"pnpm-lock.yaml"` (and any other auto-generated lockfiles) back to `ignorePatterns` to match the previous `.prettierignore` set.

.agents/references/pr-merge-guide.md
```diff
@@ -19,7 +19,7 @@ The following branch protection rules are enabled on the `main` and `v*` branche
 ### Status Checks
 
 - **Require status checks to pass**: All of the following required status checks must be green.
-  - Prettier
+  - OxFmt
```
claude-for-yamada-ui: **Issue:** The required status-check entry is renamed to `OxFmt`, but the actual GitHub Actions job in `.github/workflows/quality.yml` is declared with `name: Format`. GitHub surfaces required checks by the job `name:` value, so the documentation no longer identifies the real check.

**Fix:** Change the bullet to `- Format` to match `name: Format` in `.github/workflows/quality.yml`, or alternatively rename the workflow job to `OxFmt` so both line up.

packages/workspace/src/oxfmt/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  const { code } = await originalFormat(
+    path.split("/").at(-1)!,
+    typeof content === "string" ? content : JSON.stringify(content),
+    { ...config, ...options },
+  )
+
+  await writeFile(path, code)
+}
+
+export default config
```
claude-for-yamada-ui: **Issue:** `export default config` re-exports the imported config from this module, but no consumer uses the default export. The root `oxfmt.config.ts` imports directly from the config module, and other call sites use named imports (`format`, `writeFileWithFormat`, `FormatConfigWithParser`). This default export is dead code carried over without a caller.

**Fix:** Drop `export default config` from `packages/workspace/src/oxfmt/index.ts`; callers that actually need the config object can continue to import it from `./config.ts`.

packages/workspace/src/oxfmt/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  options?: FormatConfig,
+) {
+  const { code } = await originalFormat(
+    path.split("/").at(-1)!,
+    typeof content === "string" ? content : JSON.stringify(content),
+    { ...config, ...options },
+  )
+
+  await writeFile(path, code)
+}
```
claude-for-yamada-ui, line 40: **Issue:** `writeFileWithFormat` no longer wraps the `originalFormat` call in try/catch. The prior Prettier-based implementation delegated to `format()`, which swallowed parse errors and fell back to the original content, so the file was always written. Now, if oxfmt cannot parse the input (e.g., intermediate codegen output that is syntactically invalid before a subsequent lint-fix pass), `writeFileWithFormat` will throw and leave the destination file unwritten, breaking generator pipelines that previously tolerated a transient unformatted write.

**Fix:** Wrap the `originalFormat` call in a try/catch (mirroring the `format()` helper just above) and fall back to writing the raw `content` on failure, preserving the previous behavior.

claude-for-yamada-ui (dismissed):
Thanks for migrating from Prettier to oxfmt. A few concerns block approval: (1) the new oxfmt ignore list drops `pnpm-lock.yaml`, which the previous `.prettierignore` excluded and would reintroduce noise into formatter scope on every dependency update; (2) `.agents/references/pr-merge-guide.md` now names the required check `OxFmt`, but the actual Quality workflow job is `name: Format`, so the doc points at a non-existent status check; (3) the new `writeFileWithFormat` no longer catches formatter errors, reverting the pre-existing contract whereby the helper fell back to writing the unformatted string on parse failure; (4) the default export on the new oxfmt workspace module has no caller in the repo and looks like carry-over.

packages/workspace/src/oxfmt/config.ts
codex-for-yamada-ui, line 15: **Issue:** The new formatter config dropped the old `pnpm-lock.yaml` exclusion. Because the root `format` script now runs `oxfmt --check` across the repository, dependency updates will start pulling the lockfile into formatter scope and create lockfile-only churn that was explicitly avoided by the previous `.prettierignore`.

**Fix:** Add `pnpm-lock.yaml` back to the ignore list for oxfmt, either in `ignorePatterns` here or via an explicit ignore file passed to the formatter.

codex-for-yamada-ui (dismissed):
The oxfmt migration looks workable overall, but the new formatter config drops the `pnpm-lock.yaml` exclusion that the removed `.prettierignore` carried. With the root `format` script now running `oxfmt --check` across the repo, dependency updates can pull the lockfile into formatter scope and create lockfile-only churn that was explicitly avoided before. Please restore an equivalent exclusion before merging.

.agents/references/pr-merge-guide.md
composer-for-yamada-ui: **Issue:** The merge guide lists the required Quality formatter check as `OxFmt`, but `.github/workflows/quality.yml` defines that job with `name: Format`, so GitHub will expose the check as **Format**. Branch protection and contributor docs should use the exact check title shown in the UI.

**Fix:** Change the bullet to `Format` (or rename the workflow job's `name` to `OxFmt` and keep the doc aligned with that choice).

packages/workspace/src/oxfmt/index.ts
```diff
@@ -0,0 +1,39 @@
...
+  content: any,
+  options?: FormatConfig,
+) {
+  const { code } = await originalFormat(
+    path.split("/").at(-1)!,
+    typeof content === "string" ? content : JSON.stringify(content),
+    { ...config, ...options },
+  )
+
+  await writeFile(path, code)
```
composer-for-yamada-ui: **Issue:** `writeFileWithFormat` calls `originalFormat` with no `try/catch`. The removed Prettier implementation routed through `format()`, which caught failures and still wrote the original string, so codegen could complete with unformatted output on formatter errors. This version propagates failures and never writes on error, which is a behavioral regression for the same helper API.

**Fix:** Either document and accept the stricter behavior across all call sites, or mirror the old semantics (e.g. wrap in `try/catch` and fall back to writing the unformatted string, matching the previous `writeFileWithFormat` contract).

CONTRIBUTING.md
```diff
@@ -112,10 +112,6 @@ the following types:
    will be used to generate a changelog when we publish an update.
    [Learn more about Changeset](https://github.com/atlassian/changesets/tree/master/packages/cli).
 
-> If you made minor changes like CI config, prettier, etc, you can run
-> `pnpm changeset add --empty` to generate an empty changeset file to document
-> your changes.
-
 5. Next, commit your changes following the [commit convention](#commit-convention).
```
composer-for-yamada-ui, line 115: **Issue:** The diff deletes the note that minor/tooling-only edits can use `pnpm changeset add --empty` to record the change. That guidance was not inherently Prettier-specific; removing it without a replacement drops useful documentation (this right-side line 115 is the nearest stable anchor after the removed block).

**Fix:** Restore equivalent guidance in generic terms (e.g. CI, formatting config, or other repo-wide tooling), or replace it with a short pointer to when an empty changeset is still appropriate.

composer-for-yamada-ui (dismissed):
Thanks for the oxfmt migration. A few issues remain before this can merge: (1) the merge-guide references the formatter check as `OxFmt`, but the actual Quality workflow job is `name: Format`, so the doc does not match the GitHub check title; (2) the new `writeFileWithFormat` helper propagates formatter errors where the previous Prettier-based implementation fell back to writing the unformatted string — the same helper API now has a stricter contract; (3) the changeset-guidance note about `pnpm changeset add --empty` for CI/config/tooling-only edits was dropped without a replacement, which removes useful generic documentation.

hirotomoyamada (approved):

PR #6597, opened 2026-04-11
fix(color-selector): merge user props via mergeProps

hirotomoyamada:
## Description

Closes #6443

Replace unsafe prop spreads in `ColorSelector` slider and color-swatch sub-components with `mergeProps` so that user-provided \`className\`, \`style\`, \`css\`, \`ref\`, and \`on*\` handlers from top-level child props and their context-level prop-bags (\`saturationSliderProps\`, \`hueSliderProps\`, \`alphaSliderProps\`, \`colorSwatchGroupProps\`, \`colorSwatchGroupLabelProps\`, \`colorSwatchItemProps\`) are correctly merged instead of being silently overwritten.

### Breaking changes

- [ ] Yes
- [x] No

### Testing

- \`pnpm react test:jsdom --run src/components/color-selector/\` 41/41 passing
- \`pnpm react format\` / \`pnpm react lint\` / \`pnpm react typecheck\` clean

### AI used

- [ ] I did not use AI.
- [x] I checked the generated content before submitting.

PR #6596, opened 2026-04-11
fix(color-picker): merge user props via mergeProps

hirotomoyamada:
## Description

Closes #6442

Replace unsafe prop spreads in `ColorPicker` with `mergeProps` so that user-provided \`className\`, \`style\`, \`css\`, \`ref\`, and \`on*\` handlers from top-level props, \`rootProps\`, \`elementProps\`, and \`startElementProps\`/\`endElementProps\` are correctly merged with context/getter props instead of being silently overwritten.

### Changes

- Root \`<InputGroup.Root>\` uses \`mergeProps({ className, css, colorScheme }, getRootProps(groupItemProps), rootProps)()\` so top-level and rootProps are merged rather than clobbered.
- \`<InputGroup.Element>\` (color-swatch and eye-dropper variants) merges \`elementProps\` with \`startElementProps\`/\`endElementProps\` via \`mergeProps\`.
- The \`useInputPropsContext()\` mapper uses \`mergeProps(context, props)()\`.
- Added regression tests covering top-level className + \`rootProps.className\` merging and \`InputPropsContext\` className merging.

The \`getFieldProps({ ...varProps, ...fieldProps })\` call on line 214 was intentionally left as-is: \`varProps\` only contains CSS custom properties and carries no mergeable keys, so the spread is already safe (same rule applied to the false positives closed in #6585 / #6437).

### Breaking changes

- [ ] Yes
- [x] No

### Testing

- \`pnpm react test:jsdom --run src/components/color-picker/\` 28/28 passing
- \`pnpm react format\` / \`pnpm react lint\` / \`pnpm react typecheck\` clean

### AI used

- [ ] I did not use AI.
- [x] I checked the generated content before submitting.

PR #6590, opened 2026-04-11
fix(date-picker): merge user props via mergeProps

hirotomoyamada:
## Description

Closes #6441

Replace unsafe prop spreads in `DatePicker` with `mergeProps` so that user-provided `className`, `style`, `css`, `ref`, and `on*` handlers from both top-level props and `rootProps` are correctly merged with context/getter props instead of being silently overwritten. Also updates the `useInputPropsContext()` pass to merge via `mergeProps`.

### Breaking changes

- [ ] Yes
- [x] No

### Testing

- `pnpm react test:jsdom --run src/components/date-picker/` 96/96 passing
- `pnpm react format` / `lint` / `typecheck` clean

### AI used

- [ ] I did not use AI.
- [x] I checked the generated content before submitting.

PR #6588, opened 2026-04-11
fix(native-select): merge user props via mergeProps

hirotomoyamada:
## Description

Closes #6440

Replace unsafe prop spreads in `NativeSelect.Root` with `mergeProps` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers from both top-level props and `rootProps` are correctly merged with context/getter props instead of being silently overwritten.

### Changes

- Wrap the `InputGroup.Root` spread with `mergeProps({ className, css, colorScheme }, getRootProps(groupItemProps), rootProps)()` so user root-level props are merged rather than clobbered.
- Route the `useInputPropsContext()` pass through `mergeProps(context, props)()`.
- Add regression tests asserting className, style, css, and onClick are merged across both \`rootProps\` and the \`InputPropsContext\`.

### Breaking changes

- [ ] Yes
- [x] No

### Testing

- \`pnpm react test:jsdom --run src/components/native-select/\` ✅ 13/13
- \`pnpm react format\` / \`pnpm react lint\` / \`pnpm react typecheck\` ✅

### AI used

- [ ] I did not use AI.
- [x] I checked the generated content before submitting.

PR #6842, opened 2026-04-18
fix(avatar): use mergeProps in PropGetter callbacks

### LINKED ISSUE #6807: bug(avatar): use mergeProps in useAvatarGroup getRootProps

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/avatar/use-avatar-group.ts:25-28 — `getRootProps` in `useAvatarGroup``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
// packages/react/src/components/avatar/use-avatar.ts (getRootProps — already migrated)
  (props) =>
    mergeProps(
      {
        "data-fallback": dataAttr(!!fallbackMessage),
        "data-loaded": dataAttr(loaded),
      },
      rest,
      props,
    )(),
  [loaded, fallbackMessage, rest],
)
```

Note: `useAvatar.getRootProps` in the same component has already been migrated and serves as the in-component reference.

### AI used

hirotomoyamada:
Closes #6807

## AI used

## Description

Migrate the `PropGetter` callbacks in `avatar` from raw `...rest` / `...props` spread to `mergeProps(...)`, so that consumer-supplied `ref`, `className`, `style`, and event handlers merge correctly with the hook's own values.

## Current behavior (updates)

The affected `PropGetter` callback(s) in `avatar` merged the hook's `...rest` with the getter's `...props` via raw spread. This silently overwrote (or failed to merge) consumer-supplied `ref`, `className`, `style`, and event handlers.

## New behavior

Each affected callback now builds its output via `mergeProps(defaults, rest, props[, internalHandlers])()`, which merges:
- `ref` via `mergeRefs`
- `className` via `cx`
- `style` via object merge
- event handlers via `handlerAll`

No. Behavior converges to what consumers already expect when customizing the affected element(s).

## Additional Information

Part of a broader cleanup tracked across issues #6805-#6839 that replaces raw spread merges in `PropGetter` callbacks across `packages/react` with `mergeProps`.


packages/react/src/components/avatar/use-avatar.ts
```diff
@@ -92,12 +92,15 @@ export const useAvatar = ({
...
 
   const getRootProps: PropGetter = useCallback(
-    (props) => ({
-      ...rest,
-      ...props,
-      "data-fallback": dataAttr(!!fallbackMessage),
-      "data-loaded": dataAttr(loaded),
-    }),
+    (props) =>
+      mergeProps(
```
codex-for-yamada-ui: **Issue**: `mergeProps` is called with the computed `data-fallback` / `data-loaded` object first, so any caller-provided `data-loaded` or `data-fallback` in `rest` or `props` now overrides the hook's actual state. This can desynchronize the root's `_loaded` styling from the real image-loading state because `avatar.style.ts` keys off `data-loaded`.

**Fix**: Merge `rest` and `props` first, then apply the computed `data-*` attributes last:
```ts
mergeProps(rest, props, {
  "data-fallback": dataAttr(!!fallbackMessage),
  "data-loaded": dataAttr(loaded),
})()
```

composer-for-yamada-ui: **Issue**: `mergeProps` applies later arguments over earlier ones for ordinary keys. With `mergeProps({ "data-fallback": ..., "data-loaded": ... }, rest, props)()`, both `rest` and `props` can overwrite the hook-managed `data-fallback` and `data-loaded` when those keys are present, reversing the previous `{ ...rest, ...props, "data-fallback": ..., "data-loaded": ... }` behavior.

**Fix**: Put internal attributes last: `mergeProps(rest, props, { "data-fallback": ..., "data-loaded": ... })()`.

packages/react/src/components/avatar/use-avatar.ts
```diff
@@ -92,12 +92,15 @@ export const useAvatar = ({
   const getGroupProps: PropGetter = useCallback((props) => ({ ...props }), [])
 
   const getRootProps: PropGetter = useCallback(
```
codex-for-yamada-ui, line 94: **Issue**: This bug fix changes how the avatar root merges props, but the PR adds no regression test covering the new merge semantics.

**Fix**: Add an avatar regression test that exercises the root prop merge path and asserts merged `className`/`style`/handler behavior while the hook's computed `data-loaded` attribute still reflects the actual load state.

composer-for-yamada-ui, line 94: **Issue**: The diff changes how root props are composed (including merge semantics for overlapping keys), but no tests were added or updated to lock in the intended precedence for `getRootProps`.

**Fix**: Add tests (e.g. in `avatar.test.tsx`) that cover merged `className` / `style` / handlers on the root and assert hook-owned `data-*` stay authoritative when overlapping props are passed through `rest` or the getter argument.

codex-for-yamada-ui (dismissed):
Two concerns on the root prop merge refactor: (1) the mergeProps argument order lets caller-provided data-* attrs override the hook's computed state, and (2) the behavioral change has no regression test coverage.

composer-for-yamada-ui (dismissed):
Two concerns overlap with Codex: mergeProps argument order reverses the previous internal-attrs-last semantics, and no tests were added to lock in the intended merge behavior.

packages/react/src/components/avatar/use-avatar.ts
```diff
@@ -92,12 +92,15 @@ export const useAvatar = ({
...
+      mergeProps(
+        {
+          "data-fallback": dataAttr(!!fallbackMessage),
+          "data-loaded": dataAttr(loaded),
+        },
+        rest,
+        props,
+      )(),
     [loaded, fallbackMessage, rest],
   )
```
claude-for-yamada-ui, line 101: **Issue:** The PR claims to fix a bug where consumer-supplied `ref`/`className`/`style`/event handlers were silently overwritten in `getRootProps`, but no test was added to exercise the new merge behavior. CLAUDE.md mandates tests for bug fixes, and all preceding PRs in this cleanup series (for example #6597 color-selector, #6596 color-picker, #6590 date-picker, #6588 native-select) added dedicated merge tests in the corresponding `*.test.tsx`. Without tests, the fix is unverified and can silently regress.

**Fix:** Add tests to `packages/react/src/components/avatar/avatar.test.tsx` covering the four contract points: (a) `className` from `rest` (e.g. via `<Avatar className="foo" />`) is concatenated with consumer `className` (e.g. via `getRootProps({ className: "bar" })` in a custom harness, or via the `Avatar` component itself using class-merging), (b) `style` objects merge, (c) both `onClick` handlers fire, (d) both refs resolve. Follow the structure used by the previously merged `color-selector`/`color-picker` tests.

packages/react/src/components/avatar/use-avatar.ts
```diff
@@ -92,12 +92,15 @@ export const useAvatar = ({
...
-    }),
+    (props) =>
+      mergeProps(
+        {
+          "data-fallback": dataAttr(!!fallbackMessage),
+          "data-loaded": dataAttr(loaded),
+        },
+        rest,
+        props,
+      )(),
```
claude-for-yamada-ui: **Issue:** The argument order passed to `mergeProps` reverses the priority of `data-fallback`/`data-loaded` compared with the previous implementation. Originally these data attributes were spread last (`{ ...rest, ...props, "data-fallback": ..., "data-loaded": ... }`), so the hook's values always won. With `mergeProps(dataAttrs, rest, props)()`, because the generic fallback inside `mergeProps` is `result[key] = !isUndefined(props[key]) ? props[key] : result[key]` (see `packages/react/src/core/components/props.ts:123`), a consumer (or a prop accidentally forwarded via `...rest` to `useAvatar`) passing `data-fallback`/`data-loaded` will now silently override the hook's internal state. The PR description explicitly states behavior should "converge to what consumers already expect" — overridable internal state markers are the opposite of that.

**Fix:** Put the hook-owned data attributes last so they retain their previous priority, e.g. `mergeProps(rest, props, { "data-fallback": dataAttr(!!fallbackMessage), "data-loaded": dataAttr(loaded) })()`. Add a regression test that passes conflicting `data-fallback` via props and asserts the hook's value still wins.

claude-for-yamada-ui (changes requested):
- Tests: The PR claims to fix a bug where consumer-supplied `ref`/`className`/`style`/event handlers were silently overwritten in `getRootProps`, but no test was added to exercise the new merge behavior. CLAUDE.md mandates tests for bug fixes, and all preceding PRs in this cleanup series (for example #6597 color-selector, #6596 color-picker, #6590 date-picker, #6588 native-select) added dedicated merge tests in the corresponding `*.test.tsx`. Without tests, the fix is unverified and can silently regress.
- Correctness: The argument order passed to `mergeProps` reverses the priority of `data-fallback`/`data-loaded` compared with the previous implementation. Originally these data attributes were spread last (`{ ...rest, ...props, "data-fallback": ..., "data-loaded": ... }`), so the hook's values always won. With `mergeProps(dataAttrs, rest, props)()`, because the generic fallback inside `mergeProps` is `result[key] = !isUndefined(props[key]) ? props[key] : result[key]` (see `packages/react/src/core/components/props.ts:123`), a consumer (or a prop accidentally forwarded via `...rest` to `useAvatar`) passing `data-fallback`/`data-loaded` will now silently override the hook's internal state. The PR description explicitly states behavior should "converge to what consumers already expect" — overridable internal state markers are the opposite of that.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: **Issue:** This bug fix changes `getRootProps` merge behavior (switching from spread overwrite to `mergeProps` composition), but there is no accompanying test in the PR diff to verify the newly intended behavior (merged handlers/refs/className/style). The current avatar tests only cover rendering/fallback/group behavior, so this regression path remains unguarded.

**Fix:** Add focused tests for the changed path (e.g., exercise `getRootProps` behavior via `useAvatar` or component-level integration) that assert callback composition and prop merging semantics (`ref`, `className`, `style`, and event handlers) so the fix is enforced going forward.

codex-for-yamada-ui (changes requested):
- Testing: This bug fix changes `getRootProps` merge behavior (switching from spread overwrite to `mergeProps` composition), but there is no accompanying test in the PR diff to verify the newly intended behavior (merged handlers/refs/className/style). The current avatar tests only cover rendering/fallback/group behavior, so this regression path remains unguarded.

packages/react/src/components/avatar/use-avatar.ts
composer-for-yamada-ui: **Issue:** `mergeProps` applies arguments left-to-right; for keys already in `result`, a defined value from a later argument overwrites (`props.ts` の `result[key] = !isUndefined(props[key]) ? props[key] : result[key]`). Putting the `data-fallback` / `data-loaded` object first means values from `rest` (and then `props`) can replace the hook’s `dataAttr(!!fallbackMessage)` and `dataAttr(loaded)`. The previous object spread always set those two keys last, so the hook’s state always won. This can desync internal styling/state from the real loaded/fallback state when consumers pass those `data-*` attributes (they are valid `HTMLProps` on the root).

**Fix:** Merge consumer props first, then apply hook data attributes last, e.g. `mergeProps(rest, props, { "data-fallback": dataAttr(!!fallbackMessage), "data-loaded": dataAttr(loaded) })()`, so ref/className/style/events still compose across `rest` and `props` while the hook-owned `data-*` values keep final precedence.

packages/react/src/components/avatar/use-avatar.ts
```diff
@@ -92,12 +92,15 @@ export const useAvatar = ({
...
     [loaded, fallbackMessage, rest],
```
composer-for-yamada-ui, line 100: **Issue:** The diff changes `getRootProps` behavior (the stated goal is correct merging of `ref`, `className`, `style`, and event handlers), but `avatar.test.tsx` in this package still only covers rendering and group behavior; nothing asserts merged refs/handlers/styles or the intended precedence of hook `data-*` vs spread props. Repository rules expect tests for bug fixes on changed code paths.

**Fix:** Add focused tests (e.g. render `Avatar` with `className`, `style`, `onClick`, and `ref` on the root plus `getRootProps`-style usage if applicable) asserting composition rather than last-writer wins, and a case that hook `data-loaded` / `data-fallback` stay authoritative when overlapping keys appear on the root.

composer-for-yamada-ui (changes requested):
- 正確性: `mergeProps` applies arguments left-to-right; for keys already in `result`, a defined value from a later argument overwrites (`props.ts` の `result[key] = !isUndefined(props[key]) ? props[key] : result[key]`). Putting the `data-fallback` / `data-loaded` object first means values from `rest` (and then `props`) can replace the hook’s `dataAttr(!!fallbackMessage)` and `dataAttr(loaded)`. The previous object spread always set those two keys last, so the hook’s state always won. This can desync internal styling/state from the real loaded/fallback state when consumers pass those `data-*` attributes (they are valid `HTMLProps` on the root).
- テスト: The diff changes `getRootProps` behavior (the stated goal is correct merging of `ref`, `className`, `style`, and event handlers), but `avatar.test.tsx` in this package still only covers rendering and group behavior; nothing asserts merged refs/handlers/styles or the intended precedence of hook `data-*` vs spread props. Repository rules expect tests for bug fixes on changed code paths.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/avatar/use-avatar.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6848, opened 2026-04-18
Action bar root props now merge user className and handlers with hook defaults

### LINKED ISSUE #6805: bug(action-bar): use mergeProps in PropGetter callbacks

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/action-bar/use-action-bar.ts:57 — `getRootProps``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
// packages/react/src/components/accordion/use-accordion.ts (getItemProps — already migrated)
const getItemProps: PropGetter = useCallback(
  (props) => mergeProps({ "data-expanded": dataAttr(open) }, rest, props)(),
  [open, rest],
)
```

### AI used

hirotomoyamada:
Closes #6805

## AI used

## Description

Updates `getRootProps` in `useActionBar` to use `mergeProps(rest, props)()` so `className`, `style`, `ref`, and event handlers from consumers merge with hook-owned props instead of relying on a plain object spread.

## Current behavior (updates)

Plain `{ ...rest, ...props }` did not merge `className`, `style`, or `ref` the way the rest of the library does.

## New behavior

`mergeProps` applies the same merging rules as other migrated `PropGetter` callbacks.

No

## Additional Information

Tests: `pnpm react test:jsdom --run src/components/action-bar/use-action-bar.test.ts`


packages/react/src/components/action-bar/use-action-bar.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("bar-root")
```
composer-for-yamada-ui, line 46: **Test**

The new test only checks `id`, `className`, and `data-testid`, while the PR description and changeset claim correct merging for `style`, `ref`, and event handlers. Those merge paths are therefore untested and could regress without detection.

**Fix:** Add cases that assert merged `style` (hook + user fragments combined), composed `ref` behavior (e.g. two refs both invoked or attached), and event handler composition (e.g. `onClick` on both `rest` and `getRootProps` arguments and verify both run), aligned with how `mergeProps` is specified in `mergeProps` / `handlerAll`.

codex-for-yamada-ui, line 46: **Testing**

The new regression test only verifies `id`, `className`, and a passthrough data attribute. It would still pass if `getRootProps()` stopped composing `style`, `ref`, or `on*` callbacks, even though those are the behaviors this PR explicitly claims to fix.

**Fix:** Expand this test to cover at least one merged event handler, one merged style object, and a merged callback ref passed from both the hook options and `getRootProps()` so the new behavior is actually protected.

composer-for-yamada-ui (changes requested):
The new test asserts only `id`, `className`, and `data-testid`, while the PR description and changeset promise correct merging for `style`, `ref`, and event handlers — those merge paths are therefore untested and could regress without detection.

codex-for-yamada-ui (changes requested):
The added regression test does not cover the `style`, `ref`, and event-handler merge paths that this PR explicitly claims to fix — those behaviors can regress independently while the test still passes.

claude-for-yamada-ui (dismissed):

packages/react/src/components/action-bar/use-action-bar.test.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

codex-for-yamada-ui (dismissed):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6849, opened 2026-04-18
Accordion root props now merge user values with hook defaults using mergeProps

### LINKED ISSUE #6806: bug(accordion): use mergeProps in getRootProps

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/accordion/use-accordion.ts:116-123 — `getRootProps` in `useAccordion``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
// packages/react/src/components/accordion/use-accordion.ts (getItemProps — already migrated)
const getItemProps: PropGetter = useCallback(
  (props) => mergeProps({ "data-expanded": dataAttr(open) }, rest, props)(),
  [open, rest],
)
```

Note: `getItemProps` in the same file has already been migrated and serves as the in-file reference.

### AI used

hirotomoyamada:
Closes #6806

## AI used

## Description

Replaces the manual spread and `mergeRefs` handling in `useAccordion` `getRootProps` with `mergeProps(rest, props)()`, matching `getItemProps` and ensuring `className`, `style`, `ref`, and events merge correctly.

## Current behavior (updates)

`getRootProps` used `{ ...props, ...rest }` with a manual `ref` merge, which did not match the library’s `mergeProps` behavior for other fields.

## New behavior

`mergeProps` merges consumer props with hook `rest` consistently with other accordion getters.

No

## Additional Information

Tests: `pnpm react test:jsdom --run src/components/accordion/use-accordion-root.test.ts`


packages/react/src/components/accordion/use-accordion.ts
```diff
@@ -114,11 +115,7 @@ export const useAccordion = ({
   })
 
   const getRootProps: PropGetter = useCallback(
-    ({ ref, ...props } = {}) => ({
-      ...props,
-      ...rest,
-      ref: mergeRefs(ref, rest.ref),
-    }),
+    (props = {}) => mergeProps(rest, props)(),
```
composer-for-yamada-ui, line 118: **API semantics and backward compatibility**

Replacing `{ ...props, ...rest, ref: mergeRefs(...) }` with `mergeProps(rest, props)()` reverses precedence for overlapping keys supplied to `getRootProps` versus hook options: previously `rest` from `useAccordion` overwrote consumer props from the getter call (except `ref`, which was merged); now later `props` from `getRootProps` win, which can change behavior for any headless usage that relied on hook values overriding getter arguments.

**Fix:** Document this precedence explicitly (e.g. in hook JSDoc or the changeset) and treat it as an intentional bug fix for #6806, or — if hook precedence must be preserved for those keys — adjust the `mergeProps` argument order so behavior matches the intended contract, and encode the chosen rule in a regression test.

packages/react/src/components/accordion/use-accordion-root.test.ts
```diff
@@ -0,0 +1,24 @@
...
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("acc-root")
+  })
```
claude-for-yamada-ui: **Test**

The new test only asserts `id` precedence plus `className` / `data-testid` propagation. It never verifies that `ref`, `style`, or event handlers are actually merged by the new `mergeProps(rest, props)()` call — the exact fields this bug fix targets. A regression that dropped ref or event merging (e.g. swapping the mergeProps argument order) would not be caught here.

**Fix:** Extend the test to also assert ref merging (both refs receive the node), event-handler chaining (both handlers fire on the merged result), and style merging for `getRootProps`, mirroring the behavioral guarantees claimed in the changeset.

packages/react/src/components/accordion/use-accordion.ts
composer-for-yamada-ui, line 118: **Internal consistency**

`getRootProps` now follows `mergeProps` semantics (consumer `getRootProps` props override hook `rest` on conflicts, with merged `className` / `style` / `ref` / events), while `getItemProps` in the same module still uses `...props, ...rest` so hook `rest` still overrides consumer props for overlapping keys. Headless composition therefore behaves differently on the root than on items.

**Fix:** Either document that root and item intentionally differ (and why), or align `getItemProps` with the same merging strategy in a follow-up change.

packages/react/src/components/accordion/use-accordion.ts
```diff
@@ -114,11 +115,7 @@ export const useAccordion = ({
...
     [rest],
   )
```
claude-for-yamada-ui, line 120: **Correctness**

Swapping `{ ...props, ...rest }` for `mergeProps(rest, props)()` flips precedence for non-merged scalar fields (`id`, `role`, `aria-*`, etc.) from "rest wins" to "props wins". Any consumer that currently passes those attributes into `useAccordion(props)` expecting them to override values supplied through `getRootProps({...})` will silently observe a different attribute on the rendered root. The PR is marked as "No" for breaking change and the changeset uses `patch`, but this is a user-observable behavior flip on a publicly exported hook.

**Fix:** Either call the precedence flip out explicitly in the changeset (and reconsider the bump), or preserve the original ordering with `mergeProps(props, rest)()` so that `rest` still wins for scalar props while ref / className / style / events merge. Confirm which ordering the codebase intends and align with the other migrated getters.

packages/react/src/components/accordion/use-accordion-root.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("acc-root")
```
composer-for-yamada-ui: **Testing and changelog alignment**

The new test only asserts `id`, `className`, and `data-testid`, while the changeset states that `style`, `ref`, and event handlers also combine correctly with hook defaults — those behaviors are not exercised by the test.

**Fix:** Extend the test with cases that assert merged `style`, merged `ref` behavior, and composed event handlers (e.g. `onClick`) between hook options and `getRootProps`, or narrow the changeset wording to match what is actually verified.

packages/react/src/components/accordion/use-accordion.ts
claude-for-yamada-ui, line 118: **Consistency**

The PR description justifies the change as "matching `getItemProps`", but `getItemProps` in the same file (around lines 234-242) still uses manual `{ ...props, ...rest }` + `mergeRefs`, and `getButtonProps` / `getPanelProps` are likewise unmigrated. The "consistency with `getItemProps`" rationale does not hold against the actual file.

**Fix:** Correct the PR description (the change does not match `getItemProps` today). Per the repository rule "do not bundle multiple fixes", sibling-getter migration should be a separate PR.

composer-for-yamada-ui (changes requested):
The switch to `mergeProps(rest, props)()` changes precedence semantics compared to the previous `{ ...props, ...rest }`, and the sibling getter `getItemProps` still uses the old pattern — so root and item getters now merge differently. The added test also does not cover the `style`, `ref`, and event-handler merging behaviors the changeset explicitly promises.

claude-for-yamada-ui (changes requested):
Flagging three issues: the new test does not exercise the ref / style / event-handler merge paths the changeset explicitly lists, the switch from `{ ...props, ...rest }` to `mergeProps(rest, props)()` flips precedence for scalar fields (consumer now wins over hook rest, e.g. `id`) which is user-visible on a publicly exported hook, and the PR description claims consistency with `getItemProps` even though `getItemProps` in the same file still uses the manual spread + `mergeRefs`.

codex-for-yamada-ui (dismissed):

packages/react/src/components/accordion/use-accordion-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/accordion/use-accordion-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/accordion/use-accordion.ts
codex-for-yamada-ui: Scoped this PR to `getRootProps` and documented the root-only semantics in code; sibling getter migration will be a separate PR per the repository's no-bundled-fixes rule.

codex-for-yamada-ui: Fixed.

packages/react/src/components/accordion/use-accordion.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/accordion/use-accordion.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui: The code comment documenting root-only semantics is a good step, but the PR description still reads 'matching `getItemProps`' in the Description section. In this revision `getItemProps` (same file, ~line 236) continues to use manual `{ ...props, ...rest }` + `mergeRefs`, so the claim is still inaccurate and the original concern stands. Please update the PR body to drop the 'matching `getItemProps`' phrasing (e.g. describe this as a root-only migration aligned with the library's `mergeProps` convention) so the rationale matches the code.

composer-for-yamada-ui: Re-checked at 6bb60b5758636ef07895761c82602d67eb4a9ce3: `getRootProps` is documented and tested, but `getItemProps` still spreads `...props` then `...rest`, so hook `rest` still wins on scalar conflicts—the opposite of `mergeProps(rest, props)()` on the root. That cross-getter inconsistency is not yet explained for consumers. Please add a brief note (e.g. JSDoc on `useAccordion` / `getRootProps` or next to `getItemProps`) that root vs item precedence intentionally differs and why, or track item-getter alignment in a separate PR if that is the preferred fix.

codex-for-yamada-ui: I can't edit the PR body in this flow. The code is now self-documented via JSDoc, and the `matching getItemProps` wording should be corrected by the author as a docs-only follow-up.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (dismissed):

packages/react/src/components/accordion/use-accordion.ts
claude-for-yamada-ui: Code is sound: `getRootProps` now carries JSDoc documenting the root-only migration and its intentional divergence from sibling getters, and the new test exercises ref/style/event merging. Remaining nit is purely PR-body wording (the 'matching getItemProps' phrase is no longer accurate since sibling getters stay on `{ ...props, ...rest }`). Resolving this thread; please update the PR description at merge time to describe this as a root-only migration aligned with the library's `mergeProps` convention.

composer-for-yamada-ui (dismissed):

claude-for-yamada-ui (dismissed):

packages/react/src/components/accordion/use-accordion-root.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      backgroundColor: "blue",
+      color: "red",
+    })
+    expect(hookRef.current).toBe(node)
+    expect(userRef.current).toBe(node)
+    expect(hookOnClick).toHaveBeenCalledWith(event)
+    expect(userOnClick).toHaveBeenCalledWith(event)
+    expect(merged["data-testid"]).toBe("acc-root")
+  })
+})
```
claude-for-yamada-ui: **Issue:** The test file is named `use-accordion-root.test.ts`, but it tests the `useAccordion` hook exported from `./use-accordion`. The established codebase convention for hook unit tests is `use-<module>.test.ts` mirroring the hook module name (e.g., `use-pagination.test.ts` tests `use-pagination.ts`, `use-format-number.test.ts` tests `use-format-number.ts`). There is no `useAccordionRoot` hook — the `-root` suffix implies a separate module that does not exist, making this test file discoverable in a way that diverges from siblings.

**Fix:** Rename the file to `use-accordion.test.ts` to match the hook module it exercises. If future tests for `useAccordionItem` are planned, keep them in the same file under a separate `describe("useAccordionItem", ...)` block, or create `use-accordion.test.ts` now as the canonical home.

packages/react/src/components/accordion/use-accordion.ts
claude-for-yamada-ui, line 120: **Issue:** The PR description states this change is "matching `getItemProps`," but `getItemProps` (lines 234-242, same file, immediately below the diff hunk) still uses the prior `{ ...props, ...rest, ref: mergeRefs(ref, rest.ref) }` spread pattern. After this PR, `getRootProps` and `getItemProps` merge sibling props with divergent semantics (root merges className/style/events and flips scalar precedence to getter-wins; item still silently overwrites className/style/events with `rest`). This contradicts the stated rationale and leaves two getters on the same hook behaving inconsistently.

**Fix:** Either correct the PR description so it no longer claims parity with `getItemProps`, or file a separate follow-up issue for `getItemProps` (and `getButtonProps`, which also merges manually via `handlerAll`/`mergeRefs`) so the inconsistency is explicitly tracked rather than papered over.

claude-for-yamada-ui (changes requested):
- 一貫性: The test file is named `use-accordion-root.test.ts`, but it tests the `useAccordion` hook exported from `./use-accordion`. The established codebase convention for hook unit tests is `use-<module>.test.ts` mirroring the hook module name (e.g., `use-pagination.test.ts` tests `use-pagination.ts`, `use-format-number.test.ts` tests `use-format-number.ts`). There is no `useAccordionRoot` hook — the `-root` suffix implies a separate module that does not exist, making this test file discoverable in a way that diverges from siblings.
- 正確性: The PR description states this change is "matching `getItemProps`," but `getItemProps` (lines 234-242, same file, immediately below the diff hunk) still uses the prior `{ ...props, ...rest, ref: mergeRefs(ref, rest.ref) }` spread pattern. After this PR, `getRootProps` and `getItemProps` merge sibling props with divergent semantics (root merges className/style/events and flips scalar precedence to getter-wins; item still silently overwrites className/style/events with `rest`). This contradicts the stated rationale and leaves two getters on the same hook behaving inconsistently.

codex-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/accordion/use-accordion-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/accordion/use-accordion.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6854, opened 2026-04-18
Carousel root props now merge user styles and refs with mergeProps

### LINKED ISSUE #6810: bug(carousel): use mergeProps in remaining PropGetter callbacks

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/carousel/use-carousel.ts:417 — `getRootProps``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
  (props) =>
    mergeProps(
      {
        id,
      },
      rest,
      props,
    )(),
)
```

Note: `getRootProps` in the same file has already been migrated and serves as the in-file reference.

### AI used

hirotomoyamada:
Closes #6810

## AI used

## Description

useCarousel getRootProps uses mergeProps for id, ARIA, rest, and user props.

No

## Additional Information

Tests: pnpm react test:jsdom --run src/components/carousel/use-carousel-root.test.ts


packages/react/src/components/carousel/use-carousel.ts
```diff
@@ -410,16 +411,21 @@ export const useCarousel = ({
...
+      mergeProps(
+        {
+          id,
+          "aria-roledescription": "carousel",
+          "data-orientation": orientation,
+        },
+        rest,
+        props,
+        {
+          ref: mergeRefs(ref, rest.ref),
```
claude-for-yamada-ui: **Issue:** Pre-computing `mergeRefs(ref, rest.ref)` as the 4th arg to `mergeProps` causes `rest.ref` to be invoked twice per mount/unmount. Trace: arg2 seeds `result.ref = rest.ref`; arg3 (props without ref) triggers the default merge path and wraps it into `mergeRefs(rest.ref, undefined)`; arg4's pre-merged ref then wraps again, yielding `mergeRefs(<calls rest.ref>, <calls ref then rest.ref>)`. When the callback fires, `rest.ref` receives the node twice. The referenced fix in issue #6810 avoided this by leaving ref handling to `mergeProps` itself (not destructuring `ref` out of `props`, or passing only the bare user `ref` as the last arg). The PR description explicitly claims `ref` now "combines correctly," but this implementation regresses that case.

**Fix:** Let `mergeProps` do ref merging by dropping the pre-computed `mergeRefs` call, e.g. `mergeProps({id, "aria-roledescription": "carousel", "data-orientation": orientation}, rest, { ref, ...props }, { onMouseEnter, onMouseLeave })()`, matching the pattern suggested in the linked issue.

packages/react/src/components/carousel/use-carousel-root.test.ts
```diff
@@ -0,0 +1,25 @@
...
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("carousel-root")
+    expect(merged["aria-roledescription"]).toBe("carousel")
+  })
+})
```
claude-for-yamada-ui: **Issue:** The PR summary and changeset assert that `className`, `style`, `ref`, and event handlers now combine with hook defaults, but the only test case validates `id`, `className`, `data-testid`, and `aria-roledescription`. `ref` merging (critical given Finding 1), `style` merging, `data-orientation` defaulting, and `onMouseEnter`/`onMouseLeave` chaining with hook defaults (the whole reason for the final `mergeProps` arg) are all untested. This leaves the behavior the PR claims to fix uncovered and would not catch the double-ref invocation.

**Fix:** Add assertions that (a) user-provided `ref` and `rest.ref` callbacks each receive the node exactly once, (b) user-provided `style` objects are merged with hook defaults rather than overwritten, (c) `data-orientation` reflects `orientation`, and (d) user `onMouseEnter`/`onMouseLeave` handlers run alongside the hook's hover handlers (e.g., verify hook hover state toggles while user handler also fires).

claude-for-yamada-ui (changes requested):
- 正確性: Pre-computing `mergeRefs(ref, rest.ref)` as the 4th arg to `mergeProps` causes `rest.ref` to be invoked twice per mount/unmount. Trace: arg2 seeds `result.ref = rest.ref`; arg3 (props without ref) triggers the default merge path and wraps it into `mergeRefs(rest.ref, undefined)`; arg4's pre-merged ref then wraps again, yielding `mergeRefs(<calls rest.ref>, <calls ref then rest.ref>)`. When the callback fires, `rest.ref` receives the node twice. The referenced fix in issue #6810 avoided this by leaving ref handling to `mergeProps` itself (not destructuring `ref` out of `props`, or passing only the bare user `ref` as the last arg). The PR description explicitly claims `ref` now "combines correctly," but this implementation regresses that case.
- テスト: The PR summary and changeset assert that `className`, `style`, `ref`, and event handlers now combine with hook defaults, but the only test case validates `id`, `className`, `data-testid`, and `aria-roledescription`. `ref` merging (critical given Finding 1), `style` merging, `data-orientation` defaulting, and `onMouseEnter`/`onMouseLeave` chaining with hook defaults (the whole reason for the final `mergeProps` arg) are all untested. This leaves the behavior the PR claims to fix uncovered and would not catch the double-ref invocation.

packages/react/src/components/carousel/use-carousel.ts
codex-for-yamada-ui: **Issue:** `mergeProps` already merges `ref` keys. Since `rest` contributes `ref`, adding `ref: mergeRefs(ref, rest.ref)` as a later argument creates nested merging (`mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`), which calls `rest.ref` twice per assignment.

**Fix:** Prevent automatic `ref` re-merge here: either remove `ref` from the `rest` object passed into `mergeProps`, or call `mergeProps(... )({ mergeRef: false })` and set `ref: mergeRefs(ref, rest.ref)` exactly once.

packages/react/src/components/carousel/use-carousel-root.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("carousel-root")
+    expect(merged["aria-roledescription"]).toBe("carousel")
```
codex-for-yamada-ui, line 142: **Issue:** The new test does not verify `style`, `ref`, or event-handler composition even though the PR claims those are fixed. This gap allowed the ref regression above to pass.

**Fix:** Expand this test to cover `style` merge behavior, composed `onMouseEnter`/`onMouseLeave`, and ref-callback invocation counts (both refs called once).

codex-for-yamada-ui (changes requested):
- Correctness: `mergeProps` already merges `ref` keys. Since `rest` contributes `ref`, adding `ref: mergeRefs(ref, rest.ref)` as a later argument creates nested merging (`mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`), which calls `rest.ref` twice per assignment.
- Test: The new test does not verify `style`, `ref`, or event-handler composition even though the PR claims those are fixed. This gap allowed the ref regression above to pass.

packages/react/src/components/carousel/use-carousel-root.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("carousel-root")
```
composer-for-yamada-ui, line 141: **Issue:** The new test exercises `id`, `className` / `class`, and `data-*` / `aria-roledescription`, but the linked changeset text at `.changeset/carousel-mergeprops-6810.md:5` also promises correct merging for `style`, `ref`, and event handlers. Those `mergeProps` code paths in `getRootProps` are not asserted, so a regression in `style` merging, `ref` composition, or `onMouseEnter` / `onMouseLeave` composition would not be caught by this PR as written, and the release note is not substantiated by tests.

**Fix:** Add focused assertions in this test (or extend coverage elsewhere) for at least one of: merged `style`, merged refs (e.g. `ref` on `useCarousel` plus `ref` in `getRootProps` / callback refs), and chained pointer handlers, **or** narrow the changeset line so it only claims what the tests and intended behavior both verify.

composer-for-yamada-ui (changes requested):
- Tests: The new test exercises `id`, `className` / `class`, and `data-*` / `aria-roledescription`, but the linked changeset text at `.changeset/carousel-mergeprops-6810.md:5` also promises correct merging for `style`, `ref`, and event handlers. Those `mergeProps` code paths in `getRootProps` are not asserted, so a regression in `style` merging, `ref` composition, or `onMouseEnter` / `onMouseLeave` composition would not be caught by this PR as written, and the release note is not substantiated by tests.

packages/react/src/components/carousel/use-carousel.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/components/carousel/use-carousel-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/carousel/use-carousel-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/carousel/use-carousel-root.test.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

PR #6856, opened 2026-04-18
Checkbox and checkbox group root props now merge with mergeProps

### LINKED ISSUE #6812: bug(checkbox): use mergeProps in PropGetter callbacks

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/checkbox/use-checkbox.ts:137-150 — `useCheckbox.getRootProps``
- `packages/react/src/components/checkbox/use-checkbox-group.ts:103-119 — `useCheckboxGroup.getRootProps``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
  (props) =>
    mergeProps(
      {
        id,
      },
      rest,
      props,
    )(),
)
```

### AI used

hirotomoyamada:
Closes #6812

## AI used

## Description

useCheckbox and useCheckboxGroup getRootProps use mergeProps.

No

## Additional Information

Tests: pnpm react test:jsdom --run src/components/checkbox/checkbox.test.tsx


packages/react/src/components/checkbox/use-checkbox.ts
```diff
@@ -136,15 +137,17 @@ export const useCheckbox = <Y extends string = string>({
...
   const getRootProps: PropGetter<"label"> = useCallback(
     (props = {}) => {
-      const sharedProps = {
-        ...dataProps,
-        "data-checked": dataAttr(checked),
-        "data-indeterminate": dataAttr(indeterminate),
-        ...rest,
-        ...props,
-      }
+      const merged = mergeProps(
```
claude-for-yamada-ui, line 140: **Issue:** The PR changes the merging semantics of `getRootProps` in both `useCheckbox` and `useCheckboxGroup` — consumer-supplied `className`, `style`, `ref`, and event handlers are now composed (via `cx` / `merge` / `mergeRefs` / `handlerAll`) instead of being plainly overwritten. This is precisely the behavior fix advertised in #6812 and in the changeset, but no tests are added to prove it. The existing `checkbox.test.tsx` only asserts internal class names, and never passes a custom `className` / `style` / `ref` / event handler through `rootProps` (for `Checkbox`) or through `getRootProps({...})` (for `useCheckboxGroup`). CLAUDE.md explicitly requires tests for bug fixes, and without them any future regression of the `mergeProps` plumbing will go undetected.

**Fix:** Add regression tests that specifically exercise the fixed paths. At minimum: (a) render `<Checkbox rootProps={{ className: "custom", style: { color: "red" }, onClick: handler }}>` and assert the rendered `<label>` keeps both the internal `ui-checkbox__root` class and `"custom"`, keeps the merged style, and invokes the user handler; (b) do the same for `CheckboxGroup` (or call `useCheckboxGroup` via `renderHook` and invoke `getRootProps({ className, ref, onBlur })`) and assert merging rather than replacement; (c) assert that a consumer-supplied `ref` on either getter receives the DOM node in addition to the hook's internal ref.

packages/react/src/components/checkbox/use-checkbox-group.ts
```diff
@@ -106,15 +107,20 @@ export const useCheckboxGroup = <Y extends string = string>(
...
+        dataProps,
+        {
+          "aria-describedby": cx(ariaDescribedbyProp, ariaDescribedby),
+          "aria-labelledby": cx(labelId, ariaLabelledby),
+          role: "group",
+        },
+        rest,
+        props,
+        {
+          ref: mergeRefs(ref, rest.ref),
```
claude-for-yamada-ui: **Issue:** `rest` is already passed to `mergeProps`, so `rest.ref` is picked up there and composed with any subsequent `ref`. Adding `{ ref: mergeRefs(ref, rest.ref) }` as the last arg re-introduces `rest.ref` a second time: after processing this final object, `result.ref` becomes `mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`, which means `rest.ref` is assigned the mounted node twice on every mount/unmount. This is not a correctness bug for current callers (`assignRef` is idempotent for ref objects and harmless-but-wasteful for function refs), but it is an easy-to-misread footgun that diverges from how the same merge is done elsewhere in the codebase and can cause double-invocation for function refs with side effects.

**Fix:** Drop the duplicated reference to `rest.ref`. Either pass only the getter-supplied `ref` in the final object (`{ ref }`) and rely on `mergeProps`'s built-in `mergeRef` handling to combine it with `rest.ref`, or inline the reconstructed props with `mergeProps(dataProps, { ... }, rest, { ref, ...props })()` so `ref` is merged exactly once.

claude-for-yamada-ui (changes requested):
- テスト: The PR changes the merging semantics of `getRootProps` in both `useCheckbox` and `useCheckboxGroup` — consumer-supplied `className`, `style`, `ref`, and event handlers are now composed (via `cx` / `merge` / `mergeRefs` / `handlerAll`) instead of being plainly overwritten. This is precisely the behavior fix advertised in #6812 and in the changeset, but no tests are added to prove it. The existing `checkbox.test.tsx` only asserts internal class names, and never passes a custom `className` / `style` / `ref` / event handler through `rootProps` (for `Checkbox`) or through `getRootProps({...})` (for `useCheckboxGroup`). CLAUDE.md explicitly requires tests for bug fixes, and without them any future regression of the `mergeProps` plumbing will go undetected.
- 正確性: `rest` is already passed to `mergeProps`, so `rest.ref` is picked up there and composed with any subsequent `ref`. Adding `{ ref: mergeRefs(ref, rest.ref) }` as the last arg re-introduces `rest.ref` a second time: after processing this final object, `result.ref` becomes `mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`, which means `rest.ref` is assigned the mounted node twice on every mount/unmount. This is not a correctness bug for current callers (`assignRef` is idempotent for ref objects and harmless-but-wasteful for function refs), but it is an easy-to-misread footgun that diverges from how the same merge is done elsewhere in the codebase and can cause double-invocation for function refs with side effects.

packages/react/src/components/checkbox/use-checkbox-group.ts
```diff
@@ -106,15 +107,20 @@ export const useCheckboxGroup = <Y extends string = string>(
...
+        },
+      )(),
```
codex-for-yamada-ui, line 122: **Issue:** `getRootProps` now merges `rest` (which may already contain `rest.ref`) via `mergeProps`, and then also appends `{ ref: mergeRefs(ref, rest.ref) }`. Since `mergeProps` auto-merges `ref`, `rest.ref` can be composed twice, so callback refs may be invoked twice on mount/unmount. The previous implementation invoked each ref once.

**Fix:** Ensure `rest.ref` is merged only once. For example, exclude `ref` from `rest` before `mergeProps` and set `ref: mergeRefs(ref, restRef)` once, or disable ref auto-merge in `mergeProps` and assign the final `ref` afterward.

packages/react/src/components/checkbox/use-checkbox.ts
```diff
@@ -136,15 +137,17 @@ export const useCheckbox = <Y extends string = string>({
...
+        dataProps,
+        {
+          "data-checked": dataAttr(checked),
+          "data-indeterminate": dataAttr(indeterminate),
+        },
+        rest,
+        props,
+      )()
```
codex-for-yamada-ui, line 148: **Issue:** This PR changes root prop merge semantics (`useCheckbox` and `useCheckboxGroup`) but adds no regression tests for the new merge behavior. Existing checkbox tests do not cover root-level merge paths (handler/class/ref composition), so regressions here are currently unguarded.

**Fix:** Add focused tests in `packages/react/src/components/checkbox/checkbox.test.tsx` to verify root prop merging behavior, including merged handlers/className and a case ensuring group root ref callbacks are invoked exactly once.

codex-for-yamada-ui (changes requested):
- 正確性: `getRootProps` now merges `rest` (which may already contain `rest.ref`) via `mergeProps`, and then also appends `{ ref: mergeRefs(ref, rest.ref) }`. Since `mergeProps` auto-merges `ref`, `rest.ref` can be composed twice, so callback refs may be invoked twice on mount/unmount. The previous implementation invoked each ref once.
- テスト: This PR changes root prop merge semantics (`useCheckbox` and `useCheckboxGroup`) but adds no regression tests for the new merge behavior. Existing checkbox tests do not cover root-level merge paths (handler/class/ref composition), so regressions here are currently unguarded.

packages/react/src/components/checkbox/use-checkbox.ts
composer-for-yamada-ui, line 148: **Issue:** This change introduces `mergeProps` in `getRootProps`, which alters how user props (event handlers, `className`, `style`, `ref`, and other merged keys) combine with `dataProps` and `rest` compared to plain object spread. The diff does not add or update any test that exercises the merged root props path for `useCheckbox` or the analogous `getRootProps` in `use-checkbox-group.ts`, so the new merging behavior and its regression surface are not covered, which conflicts with the project requirement that code changes of this kind include tests.

**Fix:** Add focused tests (for example in `checkbox.test.tsx` using `renderHook` and/or a small test component) that pass overlapping props to `getRootProps` / root components and assert that expected combinations occur—for instance that both a handler originating from the hook/field `rest` path and a user `onClick` are invoked, and that `className` is composed rather than one side being dropped, matching `mergeProps` in `mergeProps` (`packages/react/src/core/components/props.ts`). Add parallel coverage for the checkbox group `getRootProps` `mergeProps` call at the corresponding lines in `use-checkbox-group.ts`.

composer-for-yamada-ui (changes requested):
- テスト: This change introduces `mergeProps` in `getRootProps`, which alters how user props (event handlers, `className`, `style`, `ref`, and other merged keys) combine with `dataProps` and `rest` compared to plain object spread. The diff does not add or update any test that exercises the merged root props path for `useCheckbox` or the analogous `getRootProps` in `use-checkbox-group.ts`, so the new merging behavior and its regression surface are not covered, which conflicts with the project requirement that code changes of this kind include tests.

packages/react/src/components/checkbox/use-checkbox.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/checkbox/use-checkbox.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/checkbox/use-checkbox-group.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/checkbox/use-checkbox.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/checkbox/use-checkbox-group.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #6857, opened 2026-04-18
Color selector root props now merge user styles with data attributes via mergeProps

### LINKED ISSUE #6813: bug(color-selector): use mergeProps in getRootProps

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/color-selector/use-color-selector.ts:166 — `getRootProps``

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
// packages/react/src/components/avatar/use-avatar.ts (getRootProps — already migrated)
  (props) =>
    mergeProps(
      {
        "data-fallback": dataAttr(!!fallbackMessage),
        "data-loaded": dataAttr(loaded),
      },
      rest,
      props,
    )(),
  [loaded, fallbackMessage, rest],
)
```

### AI used

hirotomoyamada:
Closes #6813

## AI used

## Description

useColorSelector getRootProps uses mergeProps.

No

## Additional Information

Tests: pnpm react test:jsdom --run src/components/color-selector/use-color-selector-root.test.ts


packages/react/src/components/color-selector/use-color-selector-root.test.ts
```diff
@@ -0,0 +1,25 @@
+import { renderHook } from "#test"
```
claude-for-yamada-ui, line 1: **Issue:** The new test file is named `use-color-selector-root.test.ts`, which is not consistent with the codebase's convention. Every other hook test in `packages/react/src/components/**` is named after the hook itself (e.g. `use-pagination.test.ts`, `use-autosize.test.tsx`, `use-format-number.test.ts`, `use-infinite-scroll.test.tsx`); no existing test file uses a `-root` (or any per-getter) suffix. This one-off naming makes future tests for the same hook awkward (they will either duplicate files or diverge further) and breaks the pattern that `find use-<hook>.test.*` reliably locates hook tests.

**Fix:** Rename the new file to `use-color-selector.test.ts` (and keep it open for additional `useColorSelector` tests), or fold the new `describe` block into the existing `packages/react/src/components/color-selector/color-selector.test.tsx`. Avoid introducing a new `-root` sub-naming scheme.

packages/react/src/components/color-selector/use-color-selector-root.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("color-root")
+  })
```
claude-for-yamada-ui: **Issue:** The linked issue (#6813) explicitly lists `ref`, `className`, `style`, and event handlers as the props that were being silently overwritten before the fix, but the new test only exercises `id` (a plain-key override), `className` (cx merge) and `data-testid` (pass-through). The special-case branches of `mergeProps` that this fix is meant to enable — `ref` via `mergeRefs`, `style` merging, `css` array merging, and `on*` handler chaining via `handlerAll` — are untested, so a future regression that reverts `getRootProps` to raw spread would still pass this suite for those fields. It also never asserts that `data-disabled` / `data-readonly` from the hook side actually reach the merged output.

**Fix:** Extend the `describe` with cases that (a) pass a `ref` through `getRootProps` and assert both the user's ref and any hook-side ref receive the node (or at minimum that `merged.ref` is a function when a user ref is supplied), (b) pass `style` / `css` / an `onClick` and assert they merge rather than replace the hook-side values, and (c) assert the hook's own `data-disabled="true"` appears in the merged output when `disabled: true` is passed.

claude-for-yamada-ui (changes requested):
- 一貫性: The new test file is named `use-color-selector-root.test.ts`, which is not consistent with the codebase's convention. Every other hook test in `packages/react/src/components/**` is named after the hook itself (e.g. `use-pagination.test.ts`, `use-autosize.test.tsx`, `use-format-number.test.ts`, `use-infinite-scroll.test.tsx`); no existing test file uses a `-root` (or any per-getter) suffix. This one-off naming makes future tests for the same hook awkward (they will either duplicate files or diverge further) and breaks the pattern that `find use-<hook>.test.*` reliably locates hook tests.
- テスト: The linked issue (#6813) explicitly lists `ref`, `className`, `style`, and event handlers as the props that were being silently overwritten before the fix, but the new test only exercises `id` (a plain-key override), `className` (cx merge) and `data-testid` (pass-through). The special-case branches of `mergeProps` that this fix is meant to enable — `ref` via `mergeRefs`, `style` merging, `css` array merging, and `on*` handler chaining via `handlerAll` — are untested, so a future regression that reverts `getRootProps` to raw spread would still pass this suite for those fields. It also never asserts that `data-disabled` / `data-readonly` from the hook side actually reach the merged output.

codex-for-yamada-ui (dismissed):

packages/react/src/components/color-selector/use-color-selector.ts
```diff
@@ -160,12 +161,15 @@ export const useColorSelector = ({
...
-    }),
+    (props = {}) =>
+      mergeProps(
+        {
+          "data-disabled": dataAttr(disabled),
+          "data-readonly": dataAttr(readOnly),
+        },
+        rest,
+        props,
+      )(),
```
composer-for-yamada-ui, line 172: **Issue:** `getRootProps` now returns `mergeProps(... )()`, but `mergeProps` in `core/components/props.ts` is not a drop-in for `{ "data-disabled", "data-readonly", ...rest, ...props }`: it composes `ref` and event handlers and treats many keys when later objects set `value` to `undefined` differently from object spread, while the new test only covers `id`, `className`, and `data-testid` (see `use-color-selector-root.test.ts:20-23`), and the changeset (`.changeset/color-selector-mergeprops-6813.md:5`) only describes generic “correct user prop merging.”

**Fix:** Add tests that assert at least one of the mergeProps-specific paths (e.g. both `useColorSelector` and `getRootProps` pass `onClick` and both run, or `ref` composition), and/or expand the changeset to note how behavior differs from the previous `...rest, ...props` spread for keys beyond simple attributes.

packages/react/src/components/color-selector/use-color-selector-root.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    const merged = result.current.getRootProps({
+      id: "user-id",
+      className: "user",
+    })
+
+    expect(merged.id).toBe("user-id")
+    expect(String(merged.className)).toContain("hook")
+    expect(String(merged.className)).toContain("user")
+    expect(merged["data-testid"]).toBe("color-root")
```
composer-for-yamada-ui: **Issue:** The assertions are limited to `id`, `className`, and `data-testid`, so the switch to `mergeProps` for the root element does not establish coverage for the behavior this helper changes most compared to object spread.

**Fix:** Extend this test (or add an adjacent one) to cover a `mergeProps`-specific case such as `onClick`/`onPointerDown` chaining or `ref` merging so the new `getRootProps` contract is explicit in CI.

composer-for-yamada-ui (changes requested):
- Correctness: `getRootProps` now returns `mergeProps(... )()`, but `mergeProps` in `core/components/props.ts` is not a drop-in for `{ "data-disabled", "data-readonly", ...rest, ...props }`: it composes `ref` and event handlers and treats many keys when later objects set `value` to `undefined` differently from object spread, while the new test only covers `id`, `className`, and `data-testid` (see `use-color-selector-root.test.ts:20-23`), and the changeset (`.changeset/color-selector-mergeprops-6813.md:5`) only describes generic “correct user prop merging.”
- Testing: The assertions are limited to `id`, `className`, and `data-testid`, so the switch to `mergeProps` for the root element does not establish coverage for the behavior this helper changes most compared to object spread.

packages/react/src/components/color-selector/use-color-selector-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/color-selector/use-color-selector-root.test.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/color-selector/use-color-selector.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/color-selector/use-color-selector-root.test.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6586, opened 2026-04-11
fix(autocomplete): merge user props with context props via mergeProps

hirotomoyamada:
Closes #6439

## AI used

- [ ] I did not use AI to create this PR.
- [x] (If there is no check above) I checked the generated content before submitting.

## Description

Replaces unsafe object spread (including JSX double spread) with `mergeProps` in `Autocomplete` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.

Follows the reference pattern established in `sidebar.tsx`.

## Current behavior (updates)

Passing `className` / `style` / `onClick` to `Autocomplete` sub-components would replace context-side values instead of merging.

## New behavior

User-provided props are now merged with context props. Added regression tests.

## Is this a breaking change (Yes/No):

No

## Additional Information

PR #6864, opened 2026-04-19
Pagination navigation root props now merge with mergeProps

### LINKED ISSUE #6820: bug(pagination): use mergeProps in getRootProps

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/pagination/use-pagination.ts:140-148 — `getRootProps` (uses `mergeRefs(ref, rest.ref)`)`

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
  (props) =>
    mergeProps(
      {
        id,
      },
      rest,
      props,
    )(),
)
```

### AI used

hirotomoyamada:
Closes #6820

## AI used

## Description

usePagination getRootProps uses mergeProps for aria-label, role, rest, props, and ref.

No

## Additional Information

Tests: pnpm react test:jsdom --run src/components/pagination/


packages/react/src/components/pagination/use-pagination.ts
```diff
@@ -138,13 +139,18 @@ export const usePagination = ({
...
+    ({ ref, ...props } = {}) =>
+      mergeProps(
+        {
+          "aria-label": t("Pagination"),
+          role: "navigation",
+        },
+        rest,
+        props,
+        {
+          ref: mergeRefs(ref, rest.ref),
```
claude-for-yamada-ui: **Issue:** `rest.ref` is invoked twice per mount after this change. `mergeProps` already merges `rest.ref` into the result at step 2 (because `rest` is passed as an arg and `rest.ref` is kept on it). The final override `{ref: mergeRefs(ref, rest.ref)}` passes `rest.ref` again, and `mergeProps`'s `mergeRef` branch wraps it with the previous ref via `mergeRefs`. Tracing with `rest.ref = R1`, getter `ref = R2`: final callback invokes `R1(node)` (from the wrapped rest.ref), then `R2(node)`, then `R1(node)` again. The previous spread-based code called `R1` exactly once. Consumers passing a ref callback with side effects (e.g. a mount counter, measurement callback) will now observe duplicate invocations — a regression introduced by this fix.

**Fix:** Stop re-passing `rest.ref` in the final override. Either drop `rest.ref` from the final arg (e.g. `{ ref }`) since `rest.ref` is already merged via `rest`, or destructure `rest.ref` out (`const { ref: restRef, ...restProps } = rest`) and use `restRef` only in the final override while passing `restProps` in the middle slot.

packages/react/src/components/pagination/use-pagination.ts
```diff
@@ -138,13 +139,18 @@ export const usePagination = ({
...
+        },
+      )(),
     [rest, t],
   )
```
claude-for-yamada-ui, line 156: **Issue:** The PR changes observable behavior of `getRootProps` (className now cx-merged, style merged, css accumulated, event handlers chained via `handlerAll`, `undefined` no longer overrides, ref composition reworked), but no test covers these new semantics. `use-pagination.test.ts` only tests `range`, and `pagination.test.tsx` has no `getRootProps`/merge assertions. The project rule in CLAUDE.md states "Tests are required: Always write tests when fixing bugs or adding new features", and the precedent PR #6586 (`fix(autocomplete): merge user props with context props via mergeProps`) added 131 lines of tests for equivalent `mergeProps` behavior. Without tests the regression noted in Finding 1 (and future regressions) will not be caught.

**Fix:** Add tests (in `use-pagination.test.ts` or `pagination.test.tsx`) that cover merging of `className`, `style`, `css`, event handlers (e.g. `onClick`), and ref composition (verifying each ref callback is invoked exactly once) when values are supplied via both `usePagination` props (rest) and `getRootProps` props.

.changeset/pagination-mergeprops-6820.md
```diff
@@ -0,0 +1,5 @@
+---
```
claude-for-yamada-ui: **Issue:** The changeset filename `pagination-mergeprops-6820.md` and the summary `Merge \`getRootProps\` in \`usePagination\` with \`mergeProps\`.` deviate from the established pattern used by all other merge-props changesets in this family (`autocomplete-merge-props.md`, `color-picker-merge-props.md`, `color-selector-merge-props.md`, `date-picker-merge-props.md`, `file-input-merge-props.md`, `native-select-merge-props.md`, `select-merge-props.md`, `chart-props-merge-safely.md`). Those use the `<component>-merge-props.md` filename and a user-facing summary (e.g. "Use `mergeProps` in `Autocomplete` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.") that explains the bug impact rather than just restating the refactor.

**Fix:** Rename to `pagination-merge-props.md` and rewrite the summary to describe the user-visible fix (e.g. that consumer-supplied `className`, `style`, `css`, `ref`, and event handlers passed to `usePagination`/`getRootProps` are now correctly merged instead of being overwritten).

claude-for-yamada-ui (changes requested):
- 正確性: `rest.ref` is invoked twice per mount after this change. `mergeProps` already merges `rest.ref` into the result at step 2 (because `rest` is passed as an arg and `rest.ref` is kept on it). The final override `{ref: mergeRefs(ref, rest.ref)}` passes `rest.ref` again, and `mergeProps`'s `mergeRef` branch wraps it with the previous ref via `mergeRefs`. Tracing with `rest.ref = R1`, getter `ref = R2`: final callback invokes `R1(node)` (from the wrapped rest.ref), then `R2(node)`, then `R1(node)` again. The previous spread-based code called `R1` exactly once. Consumers passing a ref callback with side effects (e.g. a mount counter, measurement callback) will now observe duplicate invocations — a regression introduced by this fix.
- テスト: The PR changes observable behavior of `getRootProps` (className now cx-merged, style merged, css accumulated, event handlers chained via `handlerAll`, `undefined` no longer overrides, ref composition reworked), but no test covers these new semantics. `use-pagination.test.ts` only tests `range`, and `pagination.test.tsx` has no `getRootProps`/merge assertions. The project rule in CLAUDE.md states "Tests are required: Always write tests when fixing bugs or adding new features", and the precedent PR #6586 (`fix(autocomplete): merge user props with context props via mergeProps`) added 131 lines of tests for equivalent `mergeProps` behavior. Without tests the regression noted in Finding 1 (and future regressions) will not be caught.
- 一貫性: The changeset filename `pagination-mergeprops-6820.md` and the summary `Merge \`getRootProps\` in \`usePagination\` with \`mergeProps\`.` deviate from the established pattern used by all other merge-props changesets in this family (`autocomplete-merge-props.md`, `color-picker-merge-props.md`, `color-selector-merge-props.md`, `date-picker-merge-props.md`, `file-input-merge-props.md`, `native-select-merge-props.md`, `select-merge-props.md`, `chart-props-merge-safely.md`). Those use the `<component>-merge-props.md` filename and a user-facing summary (e.g. "Use `mergeProps` in `Autocomplete` so that user-provided `className`, `style`, `css`, `ref`, and `on*` event handlers are merged with context/getter props instead of being silently overwritten.") that explains the bug impact rather than just restating the refactor.

packages/react/src/components/pagination/use-pagination.ts
```diff
@@ -138,13 +139,18 @@ export const usePagination = ({
...
+      mergeProps(
+        {
+          "aria-label": t("Pagination"),
+          role: "navigation",
+        },
+        rest,
+        props,
+        {
+          ref: mergeRefs(ref, rest.ref),
+        },
```
codex-for-yamada-ui: **Issue:** `rest` is passed into `mergeProps` and can already contain `rest.ref`, then another `ref` is appended via `{ ref: mergeRefs(ref, rest.ref) }`. Since `mergeProps` auto-merges `ref`, this produces `mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`, so the same callback ref can be invoked twice on mount/unmount.

**Fix:** Split `ref` out of `rest` before calling `mergeProps`, or disable `mergeRef` for this call and assign a single final `ref: mergeRefs(ref, restRef)` once.

packages/react/src/components/pagination/use-pagination.ts
```diff
@@ -138,13 +139,18 @@ export const usePagination = ({
...
+      )(),
```
codex-for-yamada-ui, line 154: **Issue:** The `getRootProps` path now relies on `mergeProps` semantics (merged events/class/style/ref) instead of direct spread, but existing tests do not assert this merge contract for root props (including single-invocation ref behavior), so this regression surface is currently unguarded.

**Fix:** Add regression tests for `getRootProps`/`Pagination.Root` covering root prop merge behavior (event/class/style merge expectations) and callback ref invocation count.

codex-for-yamada-ui (changes requested):
- 正確性: `rest` is passed into `mergeProps` and can already contain `rest.ref`, then another `ref` is appended via `{ ref: mergeRefs(ref, rest.ref) }`. Since `mergeProps` auto-merges `ref`, this produces `mergeRefs(rest.ref, mergeRefs(ref, rest.ref))`, so the same callback ref can be invoked twice on mount/unmount.
- テスト: The `getRootProps` path now relies on `mergeProps` semantics (merged events/class/style/ref) instead of direct spread, but existing tests do not assert this merge contract for root props (including single-invocation ref behavior), so this regression surface is currently unguarded.

packages/react/src/components/pagination/use-pagination.ts
composer-for-yamada-ui, line 154: **Issue:** `mergeProps` combines `ref` with `mergeRefs` when a key already exists in `result` (see `packages/react/src/core/components/props.ts:81-86`). Here `rest` (second arg) is merged before the final object, so if `rest` includes `ref`, the final `ref` becomes `mergeRefs(rest.ref, mergeRefs(getRootPropsRef, rest.ref))` instead of a single `mergeRefs(getRootPropsRef, rest.ref)` as in the previous object spread, where the explicit `ref` property **replaced** the value coming from `...rest`. The resulting callback order and duplicate `assignRef` to `rest.ref` differ from the pre-change behavior and from the intended “getter ref + `rest.ref`” pairing.

**Fix:** Decompose `rest` and omit `ref` from the object passed to `mergeProps` (e.g. `const { ref: refFromRest, ...rootRest } = rest` in the hook body and use `rootRest` as the second argument, keeping `{ ref: mergeRefs(ref, refFromRest) }` last), or pass a `rest` copy without `ref` so the last block is the only place that sets `ref`.

packages/react/src/components/pagination/use-pagination.ts
composer-for-yamada-ui, line 156: **Issue:** The diff only changes `getRootProps`; `packages/react/src/components/pagination/use-pagination.test.ts` still only asserts `range` and does not cover `getRootProps` output, `mergeProps` behavior (e.g. `className` / `onClick` merging), or the `ref` contract for the root.

**Fix:** Add tests that call `getRootProps` with and without a `ref` and with overlapping props from the hook’s `...rest` so the new merge behavior and ref wiring are locked in and cannot regress.

composer-for-yamada-ui (changes requested):
- Correctness: `mergeProps` combines `ref` with `mergeRefs` when a key already exists in `result` (see `packages/react/src/core/components/props.ts:81-86`). Here `rest` (second arg) is merged before the final object, so if `rest` includes `ref`, the final `ref` becomes `mergeRefs(rest.ref, mergeRefs(getRootPropsRef, rest.ref))` instead of a single `mergeRefs(getRootPropsRef, rest.ref)` as in the previous object spread, where the explicit `ref` property **replaced** the value coming from `...rest`. The resulting callback order and duplicate `assignRef` to `rest.ref` differ from the pre-change behavior and from the intended “getter ref + `rest.ref`” pairing.
- Tests: The diff only changes `getRootProps`; `packages/react/src/components/pagination/use-pagination.test.ts` still only asserts `range` and does not cover `getRootProps` output, `mergeProps` behavior (e.g. `className` / `onClick` merging), or the `ref` contract for the root.

.changeset/pagination-mergeprops-6820.md
codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/use-pagination.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/use-pagination.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/use-pagination.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/use-pagination.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/components/pagination/use-pagination.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6876, opened 2026-04-19
fix(scroll-area): use mergeProps for getRootProps

### LINKED ISSUE #6829: bug(scroll-area): use mergeProps in getRootProps

@yamada-ui/react

### Package Version

main (current)

### Description

**Affected location(s):**

- `packages/react/src/components/scroll-area/use-scroll-area.ts:121-158 — `getRootProps` (mixes `...rest`, conditional `safariProps`, and `...props` with `handlerAll` for `onMouseEnter` / `onMouseLeave` / `onScroll`)`

### Link to Reproduction

### Steps to Reproduce

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

Yes

### Additional Information

**Suggested fix**

```ts
  (props) =>
    mergeProps(
      {
        id,
      },
      rest,
      props,
    )(),
)
```

Note: the conditional `safariProps` spread can be merged into the same `mergeProps(...)` chain as another argument, e.g. `mergeProps(defaults, rest, isSafari ? safariProps : undefined, props, { onMouseEnter, onMouseLeave, onScroll })()`.

### AI used

hirotomoyamada:
Closes #6829

## Description

Merges root props with `mergeProps` and keeps explicit `handlerAll` chains for enter/leave/scroll so callback order stays the same.

No


packages/react/src/components/scroll-area/use-scroll-area.ts
```diff
@@ -119,29 +120,36 @@ export const useScrollArea = ({
   )
 
   const getRootProps: PropGetter = useCallback(
```
claude-for-yamada-ui, line 105: **Issue:** The PR fixes a bug in which consumer-supplied `className`, `style`, `ref`, and event handlers passed through `getRootProps` were silently overwritten instead of being merged (per linked issue #6829). However, no tests were added to cover this newly fixed behavior. The existing test suite in `scroll-area.test.tsx` only exercises hover/scroll/data-attrs paths and never calls `getRootProps({ ref, className, style, onMouseEnter, ... })` with consumer-supplied values, so the bug fix is not guarded against future regressions. `CLAUDE.md` mandates tests for bug fixes.

**Fix:** Add tests that render `ScrollArea` with consumer-provided `className`, `style`, `ref`, and `onMouseEnter`/`onMouseLeave`/`onScroll` handlers, then assert: (a) the consumer's className is combined with the internal one rather than replaced; (b) the consumer's style is merged with `{ overflow: "auto" }`; (c) the consumer's ref callback is invoked alongside the internal one (especially in Safari mode where `scrollAreaRef` must still be assigned); (d) the consumer handlers fire before the hook's internal handlers in the documented order (`props → rest → internal`).

claude-for-yamada-ui (changes requested):
The bug fix changes merge semantics for consumer-supplied root props but adds no tests to cover the new behavior.

packages/react/src/components/scroll-area/use-scroll-area.ts
```diff
@@ -119,29 +120,36 @@ export const useScrollArea = ({
...
-        rest.onMouseLeave,
-        onMouseLeave,
-      ),
-      onScroll: handlerAll(props.onScroll, rest.onScroll, onScroll),
-    }),
+    ({ style, ...props } = {}) => {
+      const merged = mergeProps(
+        { ref, style: { overflow: "auto", ...style } },
+        rest,
+        isSafari ? safariProps : {},
```
codex-for-yamada-ui: **Issue:** In the Safari path, this now feeds both the raw `ref` (`{ ref, ... }`) and `safariProps.ref` (`mergeRefs(ref, scrollAreaRef)`) into `mergeProps()`. Because `mergeProps` merges refs by wrapping them with `mergeRefs`, the resulting ref becomes effectively `mergeRefs(ref, mergeRefs(ref, scrollAreaRef))`, so the consumer's ref is assigned twice for the same node on every mount/update in Safari. That is a behavioral regression from the previous code, which only exposed `mergeRefs(ref, scrollAreaRef)` once.

**Fix:** Do not let `mergeProps` merge the same external ref twice. Either omit the base `ref` when `isSafari` is true, or disable ref merging for this call and compose the final ref explicitly so the Safari branch uses a single `mergeRefs(ref, scrollAreaRef, props.ref)` chain.

codex-for-yamada-ui (changes requested):
Safari path double-merges the consumer ref through mergeProps, attaching it twice per mount/update.

packages/react/src/components/scroll-area/use-scroll-area.ts
```diff
@@ -119,29 +120,36 @@ export const useScrollArea = ({
...
-      ),
-      onMouseLeave: handlerAll(
-        props.onMouseLeave,
-        rest.onMouseLeave,
-        onMouseLeave,
-      ),
-      onScroll: handlerAll(props.onScroll, rest.onScroll, onScroll),
-    }),
+    ({ style, ...props } = {}) => {
+      const merged = mergeProps(
```
composer-for-yamada-ui: **Issue:** The diff updates `getRootProps` to use `mergeProps` but does not add or extend tests under `scroll-area`; existing `scroll-area.test.tsx` does not cover merged root props (for example `className`/`style` composition or Safari `ref`/`key` paths), so the new merge semantics are unguarded against regressions.

**Fix:** Add focused tests in `packages/react/src/components/scroll-area/scroll-area.test.tsx` (or a dedicated hook test) that render `ScrollArea` with overlapping `className` and partial `style` on the component and on props passed through the root slot, and assert the expected DOM attributes and handler invocation order where relevant.

composer-for-yamada-ui (changes requested):
The new merge semantics for root props are not covered by any test in scroll-area.

packages/react/src/components/scroll-area/use-scroll-area.ts
claude-for-yamada-ui: HEAD (`9fef382ea279e9b364ce7cefedaaa2e64eb9d5cf`) 不変のため未解消です。`getRootProps` の `mergeProps` 化で消費者由来の `className` / `style` / `ref` / `onMouseEnter` / `onMouseLeave` / `onScroll` がマージされる新挙動に対するテストが `packages/react/src/components/scroll-area/scroll-area.test.tsx` に追加されていないため、CLAUDE.md の「バグ修正にはテスト必須」要件が引き続き満たされていません。(a) 消費者の `className` が内部クラスと結合されること、(b) `style` が `{ overflow: "auto" }` とマージされること、(c) Safari パスでも消費者の ref コールバックと `scrollAreaRef` の両方に node が割り当てられること、(d) ハンドラが `props → rest → internal` の順で発火することを検証するテストを追加してください。

packages/react/src/components/scroll-area/use-scroll-area.ts
composer-for-yamada-ui: There is still no delta: `git diff 9fef382ea279e9b364ce7cefedaaa2e64eb9d5cf...9fef382ea279e9b364ce7cefedaaa2e64eb9d5cf` is empty, so nothing new was pushed. In the current tree, `scroll-area.test.tsx` still does not assert merged root attributes—there are no tests referencing `className`, `style`, or `getRootProps`, so `mergeProps` composition (user `rest` vs internal `overflow: "auto"` / `data-*` / Safari `ref`+`key`) remains uncovered beyond the existing `data-key` check. Please add the focused merge tests outlined in the prior comment.

packages/react/src/components/scroll-area/use-scroll-area.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/scroll-area/use-scroll-area.ts
codex-for-yamada-ui: Fixed.

packages/react/src/components/scroll-area/use-scroll-area.ts
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6898, opened 2026-04-21
Changed ESLint to OxLint

### LINKED ISSUE #6802: Migrate project's ESLint to oxlint
## Description

Migrate the project's own linting toolchain from ESLint to oxlint (Oxc). This is about Yamada UI's internal dev tooling, not the `@yamada-ui/cli` end-user integration tracked in #5993/#5995.

- Large lint-time speedups via the Rust-based Oxc toolchain.
- Aligns internal dev tooling with the Oxc direction already scoped for the CLI (#5993, #5995).
- Reduces plugin maintenance surface over time as oxlint's native coverage grows.

## Rule Classification Summary (oxlint 1.60.0)

| Class | Count | Meaning |
|---|---|---|
| A: native | 169 | Supported natively in oxlint |
| B: native with drift | 4 | Native but behaviour differs from typescript-eslint |
| C: JS Plugins (alpha) | 76 | Covered via oxlint JS Plugins reusing the ESLint plugin |
| D: needs reimplementation | 1 | `use-client-newline` (custom) |
| E: drop | 411 | Already disabled via `eslint-config-prettier` or similar |
| F: blocked on ESLint | 0 | None |

Full rule-level inventory + plugin scoring lives in the working plan; the phases below encode the actionable ordering.

## Decisions

- **Type-aware TS rules (13 rules)**: drop. Avoids blocking on TypeScript 7.0 + `oxlint-tsgolint`. Type-aware rules such as `await-thenable`, `no-unnecessary-condition`, `require-await`, `no-unnecessary-type-assertion`, etc. will no longer be enforced. Trade-off accepted.
- **cspell**: remove the `@cspell/eslint-plugin` integration, keep `pnpm spellcheck` as a standalone CLI step, wired into lefthook and CI.
- **`use-client-newline`**: port to a dedicated oxlint JS Plugin (no native oxlint rule exists for directive padding).
- **JS Plugins (alpha)**: accepted as a long-term dependency for perfectionist, vitest, testing-library, import-alias, unused-imports, and the remaining react / react-hooks gaps.
- **B-class semantic drift (4 rules: `no-array-constructor`, `no-empty-function`, `no-unused-expressions`, `no-unused-vars`)**: accepted.
- **ESLint 9→10 upgrade (#6708)**: unrelated, will be closed separately.
- **Runtime ESLint API consumers** (`packages/cli/src/utils/lint.ts`, `www/libs/eslint-plugin-unused-imports.ts`): left in place for this migration. The `eslint` package stays as a runtime dependency of `packages/cli` and `www` until those consumers are rewritten in a separate follow-up.

## Phased Migration Plan

Each phase is a single PR. Every phase keeps ESLint authoritative until the final cleanup phase.

- [ ] **Phase 0 — Parallel wiring (S)**: install `oxlint`, add root `.oxlintrc.json` stub, add parallel `lint:oxlint` scripts, add non-blocking oxlint job to CI and lefthook. No ESLint removal.
- [ ] **Phase 1 — Next native pilot (S)**: enable 21 native `nextjs/*` rules in `www` and the two Next playgrounds only.
- [ ] **Phase 2 — Native playground baseline (M)**: enable the full native oxlint baseline (core / import / react / react-hooks `rules-of-hooks` + `exhaustive-deps` / jsx-a11y / non-type-aware typescript) in `playgrounds/{vite,react-router/framework,tanstack/router,tanstack/start}`.
- [ ] **Phase 3 — Native www rollout (M)**: expand `www/.oxlintrc.json` to the full native baseline + Next. Promote the `www` oxlint CI check to blocking.
- [ ] **Phase 4 — Native packages rollout (L)**: apply the native baseline to `packages/{react,utils,cli,forge,workspace}`.
- [ ] **Phase 5 — Import leftovers (M)**: JS Plugins alpha for `eslint-plugin-import`, `eslint-plugin-unused-imports`, and `@limegrass/eslint-plugin-import-alias`.
- [ ] **Phase 6 — React gaps (M)**: JS Plugins alpha for `react/jsx-no-leaked-render`, `react/jsx-uses-*`, `react/no-deprecated`, and the remaining `react-hooks/*` compiler-era rules.
- [ ] **Phase 7 — Test stack (L)**: JS Plugins alpha for `@vitest/eslint-plugin` and `eslint-plugin-testing-library` on `**/*.test.{ts,tsx}` globs.
- [ ] **Phase 8 — Spelling split (S)**: remove `@cspell/eslint-plugin` integration, wire `pnpm spellcheck` into lefthook + CI as its own step.
- [ ] **Phase 9 — Type-aware TS drop (M)**: remove the 13 type-aware typescript-eslint rules and accept the 4 B-class drift replacements.
- [ ] **Phase 10 — Perfectionist (L)**: port the full `packages/workspace/src/eslint/perfectionist.ts` config (229 lines, including the 9xs–9xl custom JSX prop ordering) to oxlint JS Plugins.
- [ ] **Phase 11 — Custom rule extraction (M)**: extract `custom-react/use-client-newline` from `packages/workspace/src/eslint/react.ts` into a dedicated oxlint JS Plugin. (Runtime ESLint API consumers are deliberately out of scope per the decision above.)
- [ ] **Phase 12 — Final ESLint removal (M)**: delete all `eslint.config.ts` entries, `packages/workspace/src/eslint/*`, the `./eslint` export from `@yamada-ui/workspace`, per-package `lint` → `oxlint`, remove `@cspell/eslint-plugin` / `eslint-config-prettier` / all plugin devDependencies except the ones required by the runtime consumers. Keep `eslint` + `@typescript-eslint/parser` + `eslint-plugin-unused-imports` in `packages/cli` and `www` devDependencies until the runtime consumers are rewritten (separate follow-up).

## AI used

Related #5995
Related #5993

hirotomoyamada:
Closes #6802

## AI used

## Description

Changed ESLint to OxLint.

No


packages/react/src/components/chart/chart.tsx
```diff
@@ -118,17 +123,21 @@ export function mergeSeries<Y>(
...
-  color: CSSProps["fill"] = "mono",
+  color: ExtractStyleValueWithCondition<CSSProps["fill"]> = "mono",
   strategy: GradientStrategy = "invert",
 ): (Y & { fill: CSSProps["fill"] })[] {
   const colors = gradients(data.length, color, strategy)
 
   return data.map((item, index) => ({ ...item, fill: colors[index] }))
 }
 
+export const a: ExtractStyleValueWithCondition<CSSProps["color"]> = {
```
claude-for-yamada-ui: **Issue:** `export const a: ExtractStyleValueWithCondition<CSSProps["color"]> = { base: "mono" }` is leftover debug/scratch code that has been shipped as part of the public module surface. It has no consumer in the codebase and a one-letter identifier strongly suggests it was added while experimenting with the new type helper. It will be transitively importable from `@yamada-ui/react` internals and pollutes bundles.

**Fix:** Delete the `export const a` declaration entirely.

packages/utils/src/function.ts
```diff
@@ -25,7 +25,7 @@ export function handlerAll<Y extends (ev: any, ...args: any[]) => void>(
     fns.some((fn) => {
       fn?.(ev, ...args)
 
-      return ev?.defaultPrevented
+      return ev.defaultPrevented
```
claude-for-yamada-ui: **Issue:** `handlerAll` changed `return ev?.defaultPrevented` to `return ev.defaultPrevented`. `handlerAll` is a widely used helper invoked for many DOM events and custom handlers; call sites that previously passed `undefined`/`null` as `ev` (or event-like objects without `defaultPrevented`) will now throw `TypeError: Cannot read properties of undefined`. The optional chaining was defensive and removing it is a behavioural regression buried inside a migration PR.

**Fix:** Restore `return ev?.defaultPrevented`.

packages/utils/src/dom.test.ts
```diff
@@ -95,6 +95,8 @@ describe("DOM", () => {
 
     test("should return fallback when DOM is not created", () => {
       const origCreateElement = document.createElement
+        .bind(document)
+        .bind(document)
```
claude-for-yamada-ui: **Issue:** `const origCreateElement = document.createElement.bind(document).bind(document)` double-binds the function. The second `.bind(document)` is a no-op and the chained form looks like the output of a broken auto-fix. Besides being dead code it signals that nobody ran this file after the rewrite — the intent was clearly a single `.bind(document)` like the sibling tests at lines 66, 125, 520, 793.

**Fix:** Change to a single `const origCreateElement = document.createElement.bind(document)`.

packages/react/src/components/chart/chart.tsx
```diff
@@ -137,11 +146,11 @@ export function gradients(
...
     if (strategy === "invert") {
       return [
-        `tint(colors.${color}, ${percent})`,
-        `shade(colors.${color}, ${percent})`,
+        `tint(colors.${color as string}, ${percent})`,
+        `shade(colors.${color as string}, ${percent})`,
       ]
     } else {
-      return `${strategy}(colors.${color}, ${percent})`
+      return `${strategy}(colors.${color as string}, ${percent})`
```
claude-for-yamada-ui, line 149: **Issue:** The template literals now contain `${color as string}` casts. Project rules prohibit `as` casts in favour of runtime checks / typed narrowing. Because `color` is now typed as `ExtractStyleValueWithCondition<CSSProps["color"]>` (object or string union), asserting `as string` silently discards the object branch — if a caller passes `{ base: "mono" }` (which the new type allows and the new `export const a` literally demonstrates) the template will emit `"colors.[object Object], 80%"`.

**Fix:** Either keep the parameter type as `string` (aligning the signature with the actual runtime support) or narrow with `typeof color === "string"` before interpolating and handle the object/responsive branch explicitly.

packages/react/src/components/table/table.tsx
```diff
@@ -846,7 +846,7 @@ export const Table = withContext(
             tableProps ?? {},
           )()}
         >
-          <NativeTable.Thead role="rowgroup" {...theadProps}>
+          <NativeTable.Thead {...theadProps}>
```
claude-for-yamada-ui, line 849: **Issue:** Three `role="rowgroup"` attributes were removed from `NativeTable.Thead`/`Tbody`/`Tfoot` (this line plus 946 and 1025) inside a PR titled "changed eslint to oxlint". While native `<thead>`/`<tbody>`/`<tfoot>` do carry an implicit `rowgroup` role, this is still an unrelated semantic change bundled into an ESLint migration, and CLAUDE.md explicitly says "Do not bundle multiple fixes".

**Fix:** Revert the three role deletions, or split them into a separate PR with its own a11y/test rationale.

packages/react/src/hooks/use-local-storage/index.ts
```diff
@@ -48,8 +48,7 @@ export const createStorage = <Y>(type: StorageType, name: string) => {
         if (
           typeof window === "undefined" ||
           !(type in window) ||
-          // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-          window[type] == null ||
+          (window[type] as null | Storage | undefined) == null ||
```
claude-for-yamada-ui, line 51: **Issue:** `(window[type] as null | Storage | undefined) == null` introduces an `as` type assertion that the project rules forbid, and it was added only to silence an oxlint `no-unnecessary-condition` warning that was previously disabled via an eslint-disable comment. `window[type]` is already `Storage` per lib.dom, and the reason to check is that in sandboxed iframes it can be null at runtime despite the typing.

**Fix:** Use a runtime check such as `const storage = window[type]; if (!storage) return (defaultValue ?? "") as Y` or `typeof window[type] === "undefined"`, keeping the runtime check without asserting the type.

packages/react/src/components/accordion/accordion.stories.tsx
```diff
@@ -98,7 +98,7 @@ export const Variant: Story = () => {
       {(_, row, key) => {
         const variantItems = items.map((item) => ({
           ...item,
-          button: `${item.button} (${row})`,
+          button: `${item.button as string} (${row})`,
```
claude-for-yamada-ui, line 101: **Issue:** `${item.button as string}` was added to work around what previously typechecked cleanly. This violates the project rule forbidding `as` casts and hides a type issue rather than fixing it. `item.button` has a known type at this callsite (the mock `items` array above); the cast is unnecessary if the input shape is typed correctly.

**Fix:** Give the local `items` array an explicit narrow type (e.g. `{ button: string; children: string }[]`) so `item.button` is inferred as `string`, and drop the cast.

packages/react/scripts/generate/props/index.ts
```diff
@@ -342,14 +341,14 @@ async function getThemeDefaultValue(statement: Statement, dirPath: string) {
       ? JSON.stringify(defaultProps.size)
           .replace(/"(\w+)":/g, " $1: ")
           .replace(/}$/, " }")
-      : `\"${defaultProps.variant}\"`
+      : `\"${defaultProps.variant as string}\"`
```
claude-for-yamada-ui, line 344: **Issue:** Same `as` cast anti-pattern at lines 344 and 351: `` `"${defaultProps.variant as string}"` `` and `` `"${defaultProps.size as string}"` ``. These were presumably added to silence new oxlint complaints; the original code relied on the `isObject(...) || isArray(...)` guard to narrow the type. The cast short-circuits inference and can serialise `[object Object]` if the guard is ever wrong.

**Fix:** Narrow with `typeof defaultProps.variant === "string"` (or extract the narrowed binding into a const) instead of `as string`.

packages/react/scripts/generate/styles/index.ts
```diff
@@ -624,16 +667,16 @@ function main() {
       }
 
       if (!publish) {
-        await writeFileWithFormat(STYLES_PATH, content)
+        await writeFile(STYLES_PATH, content)
```
claude-for-yamada-ui, line 670: **Issue:** `writeFileWithFormat(STYLES_PATH, content)` was replaced with the raw `writeFile` from `node:fs/promises`, and formatting is now deferred to a post-write `oxfmt --write` subprocess. If the post-write oxfmt step fails the generator exits but leaves an unformatted `styles.ts` on disk. Contrast with `scripts/generate/icons/index.ts` which defensively wraps its exec in `try {} catch {}`.

**Fix:** Either keep `writeFileWithFormat` (single-step, atomic), or mirror the icons script and wrap the post-write exec in `try/catch` so a transient formatter failure does not leave a corrupted checked-in file.

claude-for-yamada-ui (changes requested):
This PR is billed as an ESLint → Oxlint tooling migration, but the diff mixes in several functional changes (varAttr refactors, the `handlerAll` simplification, `colorMix` clamp fix, Table role deletions, a stray `export const a` in chart.tsx) that should each be reviewed on their own merits. Multiple of the new auto-fix rewrites rely on `as` assertions the project rules explicitly forbid in favour of runtime narrowing, and the `handlerAll` change from optional chaining to direct access is a real regression surfaced by the existing tests. Please split the scope, restore the defensive patterns, and replace the `as` casts with typed narrowing before landing.

www/libs/stackblitz.ts
```diff
@@ -33,11 +33,11 @@ const INDEX = `<!doctype html>
 `
 const VITE_CONFIG = `import react from "@vitejs/plugin-react-swc"
 import { defineConfig } from "vite"
-import tsconfigPaths from "vite-tsconfig-paths"
 
 // https://vite.dev/config/
 export default defineConfig({
-  plugins: [react(), tsconfigPaths()],
+  plugins: [react()],
+  resolve: { tsconfigPaths: true },
```
codex-for-yamada-ui: **Issue:** This switches the generated Vite template from `vite-tsconfig-paths` to `resolve.tsconfigPaths`, but the same template still pins `vite: "^7"` later in this file. The StackBlitz project also relies on `@/*` imports (`App` is imported as `@/app.tsx`), so on Vite 7 those aliases stop resolving and the generated example no longer starts.

**Fix:** Either keep `vite-tsconfig-paths` in the StackBlitz template until the template is bumped to Vite 8, or raise the template's `vite` dependency to a version that supports `resolve.tsconfigPaths`.

www/oxlint.config.ts
```diff
@@ -0,0 +1,32 @@
...
+    sharedConfig,
+    nextConfig,
+    reactConfig,
+    createNoRestrictedImportConfig(
+      ["app"],
+      ["components"],
+      ["theme"],
+      ["data"],
+    ),
+  ],
```
codex-for-yamada-ui, line 21: **Issue:** The deleted `www/eslint.config.ts` had a package-specific cspell ignore for `data/**`, but the new oxlint config only inherits `sharedConfig` and never restores that exception. Because `sharedConfig` enables `cspell/spellchecker` and `www` now runs `oxlint . --max-warnings=0`, existing generated files under `www/data/*.ts` are promoted from ignored spellcheck noise to hard lint failures.

**Fix:** Reintroduce the package-specific cspell exclusion in the oxlint config, e.g. by extending `createIgnoreCspellConfig("data")` or adding an override that turns off `cspell/spellchecker` for `data/**`.

codex-for-yamada-ui (changes requested):
Two tooling concerns around the `www` package configuration. The StackBlitz template now emits a `resolve.tsconfigPaths: true` entry that is not available on the template's pinned Vite 7, and the old `www/eslint.config.ts`' cspell ignore for `data/**` did not make it into the new `www/oxlint.config.ts`. Either would cause hard failures for consumers (broken StackBlitz demos / CI lint failures in generated data files) so please address both before landing.

packages/react/src/components/chart/chart.tsx
```diff
@@ -118,17 +123,21 @@ export function mergeSeries<Y>(
...
   strategy: GradientStrategy = "invert",
 ): (Y & { fill: CSSProps["fill"] })[] {
   const colors = gradients(data.length, color, strategy)
 
   return data.map((item, index) => ({ ...item, fill: colors[index] }))
 }
 
+export const a: ExtractStyleValueWithCondition<CSSProps["color"]> = {
+  base: "mono",
+}
```
composer-for-yamada-ui: **Issue:** An `export const a` value was left in the module. It is not used by chart logic, looks like leftover type-experiment code, and should not ship as a public export from this file.

**Fix:** Delete the `export const a` block entirely (keep the `ExtractStyleValueWithCondition` usage only on real APIs such as `mergeData` / `gradients`).

CONTRIBUTING.md
```diff
@@ -50,15 +50,6 @@ To improve our development process, we have set up tools and systems. Yamada UI
...
-### Visual Studio Code
-
-```json
-{
-  "typescript.tsdk": "node_modules/typescript/lib",
-  "eslint.workingDirectories": [{ "mode": "auto" }]
-}
-```
-
 ## AI Usage Policy
```
composer-for-yamada-ui, line 53: **Issue:** The PR removes the whole "Visual Studio Code" section, which drops the `typescript.tsdk` recommendation that is unrelated to migrating from ESLint to Oxlint. The same removal appears in `CONTRIBUTING.ja.md`.

**Fix:** Restore a short VS Code (or editor-agnostic) note that keeps `"typescript.tsdk": "node_modules/typescript/lib"`, and only replace or remove the ESLint-specific `eslint.workingDirectories` guidance.

packages/react/src/core/css/color-mix.ts
```diff
@@ -148,7 +148,7 @@ export function colorMix(
 
       let ratio = parseInt(tone) || 500
 
-      if (ratio < 50 && 950 < ratio) ratio = 500
+      if (ratio < 50 || 950 < ratio) ratio = 500
```
composer-for-yamada-ui, line 151: **Issue:** The `tone` branch fixes a real logic bug (`&&` → `||` for clamping), but the diff adds no unit tests for invalid ratios (e.g. below 50 or above 950 mapping to 500). The same gap applies to other user-visible fixes in this PR (`packages/react/src/components/indicator/indicator.tsx` CSS var handling and `packages/react/src/components/input/use-input-border.ts` `varAttr` usage), which also lack new tests in the diff.

**Fix:** Add focused tests (e.g. for `colorMix` / `tone` behavior and, where practical, assertions that responsive/color-mode values resolve to the expected CSS variable form for Indicator and input borders).

.changeset/shiny-news-ask.md
```diff
@@ -0,0 +1,7 @@
+---
+"@yamada-ui/react": patch
+"@yamada-ui/utils": patch
+"@yamada-ui/cli": patch
+---
+
+Formatted code.
```
composer-for-yamada-ui, line 7: **Issue:** The changeset body is only "Formatted code." while touching `@yamada-ui/react`, `@yamada-ui/utils`, and `@yamada-ui/cli`, which is too vague for consumers reading release notes and does not describe what changed or why.

**Fix:** Replace it with a short, user-facing summary of the lint-tooling migration (ESLint → Oxlint / related tooling), or split into clearer changesets per kind of change if the team expects distinct changelog bullets.

composer-for-yamada-ui (changes requested):
The tooling migration looks solid in shape, but a few items slipped through. A stray `export const a` experiment is being shipped from chart.tsx, the VS Code section deletion in CONTRIBUTING also drops an unrelated `typescript.tsdk` hint, several user-visible fixes (`colorMix`, Indicator, input-border) landed without new tests, and the changeset body ("Formatted code.") does not describe the actual changes in the touched packages. Please address these before landing.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6900, opened 2026-04-21
add mergeAria option for space-separated ARIA attributes

### LINKED ISSUE #6845: feat(mergeProps): add mergeAria option to merge space-separated ARIA attributes
### Description

Add a new `mergeAria` option to `MergePropsOptions` in [packages/react/src/core/components/props.ts](packages/react/src/core/components/props.ts) so that `mergeProps` can merge ARIA attributes whose value type is a space-separated list (IDREF list / token list) by concatenating them instead of overwriting.

`mergeProps` already merges `className`, `style`, `css`, `ref`, and event handlers, but every other prop — including ARIA attributes — is overwritten by the later value. W3C WAI-ARIA 1.2 defines several attributes whose value is a space-separated list, so the "last wins" behavior silently drops earlier IDs/tokens and breaks assistive-technology associations when multiple `PropGetter`s compose props on the same element.

Example of the current broken behavior:

```tsx
mergeProps(
  { 'aria-describedby': 'error-id' },
  { 'aria-describedby': 'helper-id' },
)()
// => { 'aria-describedby': 'helper-id' }  ❌ 'error-id' is lost
```

### Proposed Solution or API

Add `mergeAria?: boolean` to `MergePropsOptions` (default: `true`). When enabled, attributes whose ARIA value type is a space-separated list are concatenated with a single space (deduplicating empty tokens), similar to how `className` is merged today. All other ARIA attributes keep the existing last-wins behavior.

```tsx
mergeProps(
  { 'aria-describedby': 'error-id' },
  { 'aria-describedby': 'helper-id' },
)()
// => { 'aria-describedby': 'error-id helper-id' }  ✅
```

Opt-out:

```tsx
mergeProps(a, b)({ mergeAria: false })
```

**IDREF list (space-separated ID references):**

- `aria-controls`
- `aria-describedby`
- `aria-flowto`
- `aria-labelledby`
- `aria-owns`

### References

- W3C WAI-ARIA 1.2: https://www.w3.org/TR/wai-aria-1.2/#state_prop_def

### AI used

hirotomoyamada:
Closes #6845

## AI used

## Description

Attributes in the list are merged using spaces as delimiters; you can revert to the traditional "last-written-wins" behavior by setting mergeAria: false.

No


packages/react/src/core/components/props.ts
```diff
@@ -102,6 +112,12 @@ export function mergeProps(...args: (Dict | undefined)[]) {
           continue
         }
 
+        if (mergeAria && ariaSpaceSeparatedProps.has(key)) {
+          result[key] = cx(result[key], props[key])
```
codex-for-yamada-ui, line 116: **Issue:** The new `cx`-based ARIA merge path makes empty strings impossible to use as an explicit clear signal. `cx` drops falsy values, so `mergeProps({ "aria-describedby": "hint" }, { "aria-describedby": "" })()` now keeps `"hint"` instead of clearing it. That leaves stale `aria-describedby`/`aria-labelledby` relationships attached when a later prop intentionally removes them.

**Fix:** Only concatenate when both values are non-empty strings. If the incoming ARIA value is `""`, keep the previous last-wins behavior and assign `""` directly; add a test for that clear case.

codex-for-yamada-ui (changes requested):
Thanks for the new `mergeAria` option. One concern about the merging path.

packages/react/src/core/components/props.ts
```diff
@@ -67,6 +76,7 @@ export function mergeProps<Y extends (Dict | undefined)[]>(
 ): (options?: MergePropsOptions) => MergeAll<Y>
 export function mergeProps(...args: (Dict | undefined)[]) {
   return function ({
+    mergeAria = true,
```
composer-for-yamada-ui, line 79: **Issue:** `mergeAria` defaults to `true`, so every existing `mergeProps(...)(options)` call (including those that omit options and rely on defaults) now concatenates overlapping space-separated ARIA list attributes via `cx` instead of following the previous last-defined-wins rule used for ordinary keys. Call sites that intentionally overrode `aria-describedby`, `aria-labelledby`, `aria-controls`, `aria-owns`, or `aria-flowto` on a later object will emit different DOM attribute values without any code change.

**Fix:** Default `mergeAria` to `false` to preserve prior semantics and opt in where concatenation is desired, or treat this as a breaking release (major) and document migration. If the new default is intentional, audit and update all internal/external usages explicitly.

.changeset/tame-worlds-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@yamada-ui/react": minor
+---
+
+Add `mergeAria` option to `mergeProps` for merging space-separated WAI-ARIA list attributes.
```
composer-for-yamada-ui, line 5: **Issue:** The changeset frames the release as adding an option for merging ARIA lists but does not state that the option defaults to enabled or that overlapping values for those attributes will change at runtime for existing callers. That understates the behavioral impact relative to a typical minor-plus-new-option story.

**Fix:** Expand the changeset (and any release notes) to describe the default, which attributes are affected, and the previous vs. new merge behavior — or align the version bump and messaging with a deliberate breaking change.

cspell.json
```diff
@@ -4,7 +4,6 @@
   "words": [
     "activedescendant",
     "Adebayo",
-    "agumon",
     "Adlm",
     "agumon",
```
composer-for-yamada-ui, line 9: **Issue:** The diff removes a duplicate `"agumon"` dictionary entry while also adding `"flowto"`. The duplicate cleanup is unrelated to `mergeAria` / `aria-flowto` support and does not implement the feature; it is incidental churn in the same file.

**Fix:** Drop the unrelated `agumon` deduplication from this PR and keep only the `flowto` addition (or split it into a separate change if cleanup is desired).

composer-for-yamada-ui (changes requested):
Thanks for the `mergeAria` option. A few concerns about defaults, release framing, and scope that I think should land before merge.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6912, opened 2026-04-23
Fixed path handling in `getConfig` tests

hirotomoyamada:
Closes # 

## AI used

## Description

Fixed path handling in `getConfig` tests.

No

### DIFF
--- a/packages/cli/src/utils/config.test.ts
+++ b/packages/cli/src/utils/config.test.ts
@@ -128,7 +128,8 @@ describe("getConfig", () => {
 
     const config = await getConfig(tempDir, "ui.json")
     const section = config.getSection("components")
-    expect(section?.path).toBe(".//custom-components")
+    expect(section?.path).toBe("./custom-components")
+    expect(config.getSectionPath("components")).toBe("./custom-components")
   })
 
   test("getSection should resolve via sectionMap path lookup", async () => {
--- a/packages/cli/src/utils/config.test.ts
+++ b/packages/cli/src/utils/config.test.ts
@@ -128,8 +128,6 @@ describe("getConfig", () => {
 
     const config = await getConfig(tempDir, "ui.json")
     const section = config.getSection("components")
-    // NOTE: source code bug - path starting with "/" produces ".//path"
-    // getSectionPath prepends "./" to "/path" resulting in ".//path"
     expect(section?.path).toBe(".//custom-components")
   })
 
--- a/.changeset/tidy-sections-normalize.md
+++ b/.changeset/tidy-sections-normalize.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/cli": patch
+---
+
+Fixed the `getSectionPath` helper so a leading `/` no longer produces a double-slash (`.//...`) path. Paths now normalize consistently whether or not the user supplies a leading `/`.
--- a/packages/cli/src/utils/config.ts
+++ b/packages/cli/src/utils/config.ts
@@ -93,7 +93,7 @@ export async function getConfig(
     function getSectionPath(section: Section) {
       let path = userConfig[section]?.path ?? DEFAULT_PATH[section]
 
-      if (path.startsWith("/")) path = `./${path}`
+      if (path.startsWith("/")) path = `.${path}`
       if (!path.startsWith("./")) path = `./${path}`
 
       return path


packages/cli/src/utils/config.test.ts
```diff
@@ -128,8 +128,6 @@ describe("getConfig", () => {
 
     const config = await getConfig(tempDir, "ui.json")
     const section = config.getSection("components")
-    // NOTE: source code bug - path starting with "/" produces ".//path"
-    // getSectionPath prepends "./" to "/path" resulting in ".//path"
     expect(section?.path).toBe(".//custom-components")
```
claude-for-yamada-ui: **Issue:** The removed `NOTE` comments documented a real bug in `getSectionPath` (packages/cli/src/utils/config.ts:93-100): when the user-supplied path starts with `"/"`, the function does `path = `./${path}``, producing `.//custom-components` instead of a normalized path. The source bug still exists in this PR — only the comment was removed. Dropping the NOTE deletes the only in-repo record of the defect while silently canonicalizing the buggy output `.//custom-components` as "expected". The deletion does not earn its place: if nothing is being fixed, the comment should stay; if something is being fixed, the source should change and the assertion should become `./custom-components`.

**Fix:** Either keep the NOTE comments (since the bug is still present), or actually fix `getSectionPath` in `packages/cli/src/utils/config.ts` so the `"/"`-prefixed branch yields `./custom-components`, and update the assertion on line 131 to `"./custom-components"` accordingly. A test-only PR should not mask a known source bug.

claude-for-yamada-ui (changes requested):
- 必要性: The removed `NOTE` comments documented a real bug in `getSectionPath` (packages/cli/src/utils/config.ts:93-100): when the user-supplied path starts with `"/"`, the function does `path = `./${path}``, producing `.//custom-components` instead of a normalized path. The source bug still exists in this PR — only the comment was removed. Dropping the NOTE deletes the only in-repo record of the defect while silently canonicalizing the buggy output `.//custom-components` as "expected". The deletion does not earn its place: if nothing is being fixed, the comment should stay; if something is being fixed, the source should change and the assertion should become `./custom-components`.

packages/cli/src/utils/config.test.ts
codex-for-yamada-ui: **Issue:** This assertion still expects the malformed value `.//custom-components`, because `getSectionPath` currently prepends `./` twice for leading-slash inputs. After this diff removes the note that explained that source bug, the test becomes misleading: its name says "should normalize path with leading slash" even though it is intentionally locking in non-normalized output.

**Fix:** Restore the explanatory note, or update the test so its name and expectation describe the same behavior. If the intended behavior is true normalization, fix `getSectionPath` and change the expectation in the same PR.

codex-for-yamada-ui (changes requested):
- Test: This assertion still expects the malformed value `.//custom-components`, because `getSectionPath` currently prepends `./` twice for leading-slash inputs. After this diff removes the note that explained that source bug, the test becomes misleading: its name says "should normalize path with leading slash" even though it is intentionally locking in non-normalized output.

packages/cli/src/utils/config.test.ts
composer-for-yamada-ui: **Issue:** The removed comments were the only in-test explanation for why the expectation is `.//custom-components`. That value follows directly from `getSectionPath` prepending `./` to a path that already starts with `/` (`packages/cli/src/utils/config.ts` lines 93–99), which is easy to misread as a mistaken assertion once the comment is gone. The test title still claims “normalize,” while the result is a double slash, so deleting the note increases confusion without replacing it.

**Fix:** Restore a short comment above the assertion (or rename the test to describe current behavior, e.g. that a leading `/` yields a `./` + `/…` join), so the expectation stays obviously intentional.

composer-for-yamada-ui (changes requested):
- Necessity: The removed comments were the only in-test explanation for why the expectation is `.//custom-components`. That value follows directly from `getSectionPath` prepending `./` to a path that already starts with `/` (`packages/cli/src/utils/config.ts` lines 93–99), which is easy to misread as a mistaken assertion once the comment is gone. The test title still claims “normalize,” while the result is a double slash, so deleting the note increases confusion without replacing it.

packages/cli/src/utils/config.test.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #6963, opened 2026-04-24
docs(chart): clarify accessibility keyboard behavior

hirotomoyamada:
Closes: N/A

## AI used

## Description

Updates the accessibility documentation for chart components so the keyboard behavior matches the current Recharts accessibility layer behavior.

## Current behavior (updates)

The chart docs said arrow keys move focus to the next or previous data point, and the ARIA table listed `role="application"` without documenting keyboard focusability.

## New behavior

The docs now explain that arrow keys update the active data point while focus remains on the chart surface, that `Enter` toggles the active data point tooltip, and that the chart `svg` is focusable with `tabIndex={0}`.

No.

## Additional Information

Docs-only change. Format, lint, typecheck, and tests were not run per docs-only task instructions.


hirotomoyamada:
Closing this aggregate chart docs PR because the requested work has been split into one PR per chart component.

PR #6975, opened 2026-04-24
ci(quality): run lint and typecheck for playgrounds

### LINKED ISSUE #6973: ci(quality): run lint and typecheck for playgrounds
### Description

The CI quality workflow (`.github/workflows/quality.yml`) does not run `lint` or `typecheck` against the `playgrounds/**` tree, even though root scripts for it already exist.

`packages/**` and `www/**` are covered in the `lint` and `typescript` jobs of `.github/workflows/quality.yml`, but `playgrounds/**` is not. As a result, TypeScript or lint regressions introduced in the following playgrounds are not caught on PRs and can land on `main`:

- `playgrounds/next/app`
- `playgrounds/next/pages`
- `playgrounds/react-router/framework`
- `playgrounds/vite`
- `playgrounds/tanstack/router`
- `playgrounds/tanstack/start`

The root `package.json` already defines `lint:playgrounds`, `typecheck:playgrounds`, and `quality:playgrounds`, so the fix is purely a CI integration gap.

### Proposed Solution or API

Add a `playgrounds` key to the `files_yaml` of the `lint` and `typescript` jobs in `.github/workflows/quality.yml`, and add steps that run `pnpm lint:playgrounds` / `pnpm typecheck:playgrounds` when files under `playgrounds/**` change.

Watched paths:

- `playgrounds/**/*.{ts,tsx,mts}`
- `playgrounds/**/(package.json|tsconfig.json)`
- `turbo.json`
- `.github/workflows/quality.yml`

### Alternatives

Always running the playgrounds lint/typecheck unconditionally. Rejected because it would pay CI time on PRs unrelated to playgrounds, diverging from the existing selective pattern.

Yes

### AI used

hirotomoyamada:
Closes #6973

## AI used

## Description

Add `playgrounds` coverage to the `lint` and `typescript` jobs of `.github/workflows/quality.yml` so regressions in `playgrounds/**` are caught in CI.

## Current behavior (updates)

The `lint` and `typescript` jobs only look at `packages/**` and `www/**`. Changes under `playgrounds/**` (next/app, next/pages, react-router/framework, vite, tanstack/router, tanstack/start) skip lint and typecheck entirely, even though root scripts `lint:playgrounds` and `typecheck:playgrounds` already exist.

## New behavior

Each of the `lint` and `typescript` jobs now has a `playgrounds` entry in `files_yaml` and an extra step that runs `pnpm lint:playgrounds` / `pnpm typecheck:playgrounds` when files under `playgrounds/**` change. Unrelated PRs are unaffected because the steps are gated by `playgrounds_any_changed`.

Watched paths (both jobs):

- `turbo.json`
- `packages/workspace/src/oxlint/**` (lint job) / `packages/workspace/src/typescript/**` (typescript job)
- `packages/workspace/package.json`
- `playgrounds/**/(package.json|tsconfig.json)`
- `playgrounds/**/*.{ts,tsx,mts}`
- `.github/workflows/quality.yml`

No. CI-only change with no runtime or API impact.

## Additional Information

Verified locally before pushing:

- `pnpm lint:playgrounds` — all 6 playgrounds pass (`0 warnings, 0 errors`).
- `pnpm typecheck:playgrounds` — all 6 playgrounds pass.


.github/workflows/quality.yml
```diff
@@ -101,6 +108,10 @@ jobs:
         if: steps.changed-files.outputs.www_any_changed == 'true'
         run: pnpm turbo run lint --filter www
 
+      - name: Run lint
+        if: steps.changed-files.outputs.playgrounds_any_changed == 'true'
+        run: pnpm lint:playgrounds
+
```
taroj1205: I feel like if we are also going to be doing lint for playgrounds as well, it might be better to simply just do
```bash
pnpm turbo run lint --affected
```
so we won't be running 3 lint steps in worst case which will take long time.

.github/workflows/quality.yml
```diff
@@ -142,6 +160,10 @@ jobs:
         if: steps.changed-files.outputs.www_any_changed == 'true'
         run: pnpm turbo run typecheck --filter www
 
+      - name: Run typecheck
+        if: steps.changed-files.outputs.playgrounds_any_changed == 'true'
+        run: pnpm typecheck:playgrounds
+
   format:
```
taroj1205, line 170: Same here,
I feel like if we are also going to be doing typecheck for playgrounds as well, it might be better to simply just do
```bash
pnpm turbo run typecheck --affected
```
so we won't be running 3 typechecks steps in worst case which will take long time.

.github/workflows/quality.yml
hirotomoyamada: I’ll create that in a separate PR. I’ll leave it out of the scope of this one.

.github/workflows/quality.yml
hirotomoyamada: I’ll create that in a separate PR. I’ll leave it out of the scope of this one.

.github/workflows/quality.yml
```diff
@@ -142,6 +160,10 @@ jobs:
         if: steps.changed-files.outputs.www_any_changed == 'true'
         run: pnpm turbo run typecheck --filter www
 
+      - name: Run typecheck
+        if: steps.changed-files.outputs.playgrounds_any_changed == 'true'
+        run: pnpm typecheck:playgrounds
```
claude-for-yamada-ui: **Issue:** The new `Run typecheck` step (`pnpm typecheck:playgrounds`) consistently exceeds the TypeScript job's `timeout-minutes: 10` budget. Two consecutive CI runs on this branch were cancelled at ~15m (10m job timeout + ~5m cleanup grace), with no log output from the playgrounds step before the kill. The previous step (`www:typecheck`) finished at ~05:05:44, leaving ~7 minutes for playgrounds, yet nothing was emitted before cancellation — the step is either hanging or massively slower than expected in a fresh CI environment (no turbo cache, no per-playground build artifacts beyond the committed `routeTree.gen.ts`). The PR description's "all 6 playgrounds pass locally" does not reflect CI behavior because local runs hit cached state. Without diagnostics or a fix, this PR breaks the TypeScript check on every PR that touches `playgrounds/**` (and on this PR itself).

**Fix:** Reproduce the CI behavior with a clean turbo cache locally (`pnpm typecheck:playgrounds --force` or with `TURBO_FORCE=1` and node_modules cleared) to find which playground hangs, then either (a) fix the underlying typecheck (e.g. ensure `react-router typegen` / Next typegen / TanStack route generation runs before `tsgo` where required, or pin `tsgo`/`@typescript/native-preview` if that is the cause), (b) raise `timeout-minutes` on the TypeScript job to a measured value, and/or (c) add `--output-logs=full` (or `--summarize`) to `pnpm typecheck:playgrounds` so future hangs surface a culprit instead of silent kills. Do not merge until a CI run completes green.

.github/workflows/quality.yml
```diff
@@ -129,6 +140,13 @@ jobs:
               - www/tsconfig.json
               - www/turbo.json
               - .github/workflows/quality.yml
+            playgrounds:
+              - turbo.json
+              - packages/workspace/src/typescript/**
+              - packages/workspace/package.json
+              - playgrounds/**/(package.json|tsconfig.json)
```
claude-for-yamada-ui: **Issue:** The pattern `playgrounds/**/(package.json|tsconfig.json)` only matches files literally named `tsconfig.json`. The Vite playground splits its compiler options into `playgrounds/vite/tsconfig.app.json` and `playgrounds/vite/tsconfig.node.json` (the root `tsconfig.json` is just a project-references stub with empty `files`). A change that breaks types via `tsconfig.app.json` / `tsconfig.node.json` will not flip `playgrounds_any_changed`, so the typecheck step (and the same pattern at line 95 for lint) will be skipped exactly when it is needed. This silently regresses the very coverage the PR claims to add.

**Fix:** Broaden the pattern to cover sibling tsconfigs, e.g. `playgrounds/**/tsconfig*.json` (or explicitly add `playgrounds/**/tsconfig.{app,node}.json`) in both the lint and typescript `files_yaml` blocks.

claude-for-yamada-ui (changes requested):
- 正確性: The new `Run typecheck` step (`pnpm typecheck:playgrounds`) consistently exceeds the TypeScript job's `timeout-minutes: 10` budget. Two consecutive CI runs on this branch were cancelled at ~15m (10m job timeout + ~5m cleanup grace), with no log output from the playgrounds step before the kill. The previous step (`www:typecheck`) finished at ~05:05:44, leaving ~7 minutes for playgrounds, yet nothing was emitted before cancellation — the step is either hanging or massively slower than expected in a fresh CI environment (no turbo cache, no per-playground build artifacts beyond the committed `routeTree.gen.ts`). The PR description's "all 6 playgrounds pass locally" does not reflect CI behavior because local runs hit cached state. Without diagnostics or a fix, this PR breaks the TypeScript check on every PR that touches `playgrounds/**` (and on this PR itself).
- 正確性: The pattern `playgrounds/**/(package.json|tsconfig.json)` only matches files literally named `tsconfig.json`. The Vite playground splits its compiler options into `playgrounds/vite/tsconfig.app.json` and `playgrounds/vite/tsconfig.node.json` (the root `tsconfig.json` is just a project-references stub with empty `files`). A change that breaks types via `tsconfig.app.json` / `tsconfig.node.json` will not flip `playgrounds_any_changed`, so the typecheck step (and the same pattern at line 95 for lint) will be skipped exactly when it is needed. This silently regresses the very coverage the PR claims to add.

.github/workflows/quality.yml
codex-for-yamada-ui: **Issue:** The newly added playground typecheck runs `pnpm typecheck:playgrounds` inside the existing `TypeScript` job, but the reported CI run reaches this new step and then produces no further output until the job is killed by the timeout. The surrounding workflow keeps the TypeScript job timeout at 10 minutes, while this command fans out across every playground via `turbo run typecheck --filter=./playgrounds/**`, including generated-type steps such as the React Router playground. As added, the required CI path for playground changes does not reliably finish, so PRs touching playgrounds or this workflow can be blocked by a timeout rather than a useful typecheck result.

**Fix:** Make the playground typecheck CI path demonstrably complete: either split playground typechecks into separate filtered steps/jobs with an appropriate timeout and clearer logs, or increase/adjust the job timeout only after confirming the command finishes within it on GitHub Actions.

codex-for-yamada-ui (changes requested):
- Correctness: The newly added playground typecheck runs `pnpm typecheck:playgrounds` inside the existing `TypeScript` job, but the reported CI run reaches this new step and then produces no further output until the job is killed by the timeout. The surrounding workflow keeps the TypeScript job timeout at 10 minutes, while this command fans out across every playground via `turbo run typecheck --filter=./playgrounds/**`, including generated-type steps such as the React Router playground. As added, the required CI path for playground changes does not reliably finish, so PRs touching playgrounds or this workflow can be blocked by a timeout rather than a useful typecheck result.

.github/workflows/quality.yml
```diff
@@ -101,6 +108,10 @@ jobs:
         if: steps.changed-files.outputs.www_any_changed == 'true'
         run: pnpm turbo run lint --filter www
 
+      - name: Run lint
+        if: steps.changed-files.outputs.playgrounds_any_changed == 'true'
+        run: pnpm lint:playgrounds
```
composer-for-yamada-ui: **Issue:** The new step reuses the same `name: Run lint` as the two existing lint steps. The hunk for the TypeScript job also adds a third `name: Run typecheck` at lines 163–165. GitHub logs and failure annotations are harder to match to a specific step when multiple steps share the same `name` within one job.

**Fix:** Set distinct `name` values, for example `Run lint (playgrounds)` and `Run typecheck (playgrounds)`, in line with the two ranges above.

.github/workflows/quality.yml
composer-for-yamada-ui: **Issue:** This step runs `pnpm typecheck:playgrounds` (`turbo run typecheck --filter=./playgrounds/**` per the repo root `package.json`), which can run a large turbo graph. The same hunk’s included context (`.github/workflows/quality.yml:119-119` in the current file) still leaves the TypeScript job at `timeout-minutes: 10`, and the three conditional typecheck steps run in sequence. That combination can explain a job hitting the time limit; recent CI is reported to stall on this step and then be killed, which matches a wall-clock or hang scenario without a higher job timeout, step-level `timeout-minutes`, or more verbose logging (for example `TURBO_LOG_LEVEL` or turbo flags passed through the script) to see progress.

**Fix:** Increase the `typescript` job’s `timeout-minutes` (or add a `timeout-minutes` on this step) so a full `typecheck:playgrounds` can finish, and add explicit progress logging if the step is otherwise silent, so the job either completes or fails with a visible error instead of hanging until timeout.

composer-for-yamada-ui (changes requested):
- 一貫性: The new step reuses the same `name: Run lint` as the two existing lint steps. The hunk for the TypeScript job also adds a third `name: Run typecheck` at lines 163–165. GitHub logs and failure annotations are harder to match to a specific step when multiple steps share the same `name` within one job.
- 正確性: This step runs `pnpm typecheck:playgrounds` (`turbo run typecheck --filter=./playgrounds/**` per the repo root `package.json`), which can run a large turbo graph. The same hunk’s included context (`.github/workflows/quality.yml:119-119` in the current file) still leaves the TypeScript job at `timeout-minutes: 10`, and the three conditional typecheck steps run in sequence. That combination can explain a job hitting the time limit; recent CI is reported to stall on this step and then be killed, which matches a wall-clock or hang scenario without a higher job timeout, step-level `timeout-minutes`, or more verbose logging (for example `TURBO_LOG_LEVEL` or turbo flags passed through the script) to see progress.

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui: Thanks for raising the timeouts and adding `--output-logs=full --summarize`, but CI is still red: the latest run hits the playgrounds step at 05:37:46 and is killed at ~06:02 (~25 min) with zero log output from `pnpm typecheck:playgrounds`. Neither the job-level `timeout-minutes: 20` nor the step-level `timeout-minutes: 15` ever fires — the cancellation is coming from the workflow concurrency / runner kill, not from your timeouts, which strongly suggests the process is wedged (deadlock or non-flushing native binary), not just slow. `--output-logs=full --summarize` only takes effect once turbo prints anything, so it cannot help diagnose a hang that happens before the first byte. Concrete next steps before this can merge: (1) reproduce locally with `rm -rf node_modules .turbo && pnpm install && pnpm typecheck:playgrounds` (no cache) and confirm whether it ever finishes; (2) bisect which playground hangs by running `turbo run typecheck --filter=./playgrounds/vite`, `--filter=./playgrounds/next/app`, etc. one at a time; (3) try `turbo run typecheck --filter=./playgrounds/** --concurrency=1 --log-order=stream` so output is flushed per-task instead of buffered; (4) consider whether `tsgo` (`@typescript/native-preview`) is the culprit — falling back to `tsc` in the playground `typecheck` script would isolate that. Until a CI run goes green this PR cannot land.

codex-for-yamada-ui: This is not resolved yet. The latest diff adds clearer step names, a 20-minute TypeScript job timeout, a 15-minute playground step timeout, and full Turbo logs, but the attached latest TypeScript job is still failing and the log excerpt never shows `Run typecheck (playgrounds)` completing. Please keep this open until a CI run demonstrates that the new playground typecheck path actually finishes green.

composer-for-yamada-ui (dismissed):

.github/workflows/quality.yml
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #7005, opened 2026-04-29
ci: remove security workflow

hirotomoyamada:
Closes #

## AI used

## Description

Remove `.github/workflows/security.yml` (the `pinact-action` job).

## Current behavior (updates)

The `Security` workflow runs `pinact-action` on PRs that touch `.github/**/*.yml` to detect GitHub Actions referenced by mutable tags instead of commit SHAs.

## New behavior

The workflow is removed. SHA pinning is already enforced through other means:

- Every action currently used in `.github/workflows/*.yml` is already pinned to a commit SHA.
- Dependabot is configured for `package-ecosystem: github-actions` and updates SHA-pinned references in place, so they will not silently regress to tag references.
- The `Pinact` job was not registered in the repository ruleset, so a failure did not block merge — it provided no real gate.

The residual risk (a contributor adding a new tag-referenced action) is small enough to handle in code review, so the workflow's maintenance cost is no longer justified.

No.

## Additional Information


claude-for-yamada-ui (approved):

codex-for-yamada-ui (dismissed):
**File:** `.github/workflows/security.yml`

**Perspective:** 破壊的変更

**Issue:** This PR removes the only CI workflow that enforces pinned GitHub Actions (`suzuki-shunsuke/pinact-action`) for workflow changes, and no equivalent enforcement exists in the remaining `.github/workflows/*.yml`. That is a security regression in repository governance.

**Fix:** Keep this workflow, or move an equivalent pinning check into another existing workflow with the same PR trigger scope (`.github/**/*.yml`, `.github/**/*.yaml`) so action pinning remains enforced.

composer-for-yamada-ui (dismissed):
**File:** `.github/workflows/security.yml`

**Perspective:** Breaking changes

**Issue:** This deletion removes the repository’s only Pinact CI step (`grep` over the PR HEAD worktree finds no remaining `pinact` references). Workflow YAML edits under `.github/**` therefore lose PR-time enforcement that kept third-party actions pinned (see adjacent `on.pull_request.paths` at lines 10–12 in the same deleted file). Weekly Dependabot `github-actions` updates do not replace that PR-gated pinning check.

**Fix:** Keep `security.yml`, or move an equivalent Pinact job (same branch filters and `paths` gating for `.github/**/*.yml` and `.github/**/*.yaml`) into another workflow so those PRs still run Pinact with `skip_push: true` (or another agreed pinning policy).

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7006, opened 2026-05-01
Filter out undefined style props before passing

hirotomoyamada:
Closes # 

## AI used

## Description

If you do not exclude the object returned when the value is `undefined`, inconsistencies will occur in subsequent processing.

```tsx
const popoverStyleProps = usePopoverStyleProps(rest)
const props = useDatePicker({ matchWidth: true, ...popoverStyleProps })
```

In this case, the default value of `matchWidth` is overridden.

No


packages/react/src/components/popover/popover.tsx
```diff
@@ -123,13 +123,10 @@ export const usePopoverStyleProps = (props: UsePopoverStyleProps = {}) => {
...
-    flip,
-    gutter,
-    matchWidth,
-    placement,
-    strategy,
-  }
+  return useMemo(
+    () => filterUndefined({ flip, gutter, matchWidth, placement, strategy }),
+    [flip, gutter, matchWidth, placement, strategy],
+  )
```
claude-for-yamada-ui, line 129: **Issue:** This PR fixes a real behavioral bug (`usePopoverStyleProps` clobbering consumer-provided defaults with explicit `undefined` entries) but no regression test is added. CLAUDE.md mandates "Tests are required: Always write tests when fixing bugs or adding new features." `popover.test.tsx` is unchanged in the diff, and there is no `renderHook`-based assertion (e.g., that `usePopoverStyleProps({})` returns an object without the `flip`/`gutter`/`matchWidth`/`placement`/`strategy` keys, or that an undefined prop does not overwrite a default after spreading) that would have caught the regression this PR addresses. Without such a test, a future refactor that drops `filterUndefined` would silently reintroduce the bug.

**Fix:** Add a test in `packages/react/src/components/popover/popover.test.tsx` (or a new `use-popover-style-props.test.tsx`) that uses `renderHook(() => usePopoverStyleProps({ flip: undefined }))` and asserts the returned object does not contain a `flip` key (e.g., `expect("flip" in result.current).toBe(false)`), and ideally a complementary test that confirms a real value is still returned for explicitly-set props.

claude-for-yamada-ui (dismissed):
- テスト: This PR fixes a real behavioral bug (`usePopoverStyleProps` clobbering consumer-provided defaults with explicit `undefined` entries) but no regression test is added. CLAUDE.md mandates "Tests are required: Always write tests when fixing bugs or adding new features." `popover.test.tsx` is unchanged in the diff, and there is no `renderHook`-based assertion (e.g., that `usePopoverStyleProps({})` returns an object without the `flip`/`gutter`/`matchWidth`/`placement`/`strategy` keys, or that an undefined prop does not overwrite a default after spreading) that would have caught the regression this PR addresses. Without such a test, a future refactor that drops `filterUndefined` would silently reintroduce the bug.

packages/react/src/components/popover/popover.tsx
```diff
@@ -123,13 +123,10 @@ export const usePopoverStyleProps = (props: UsePopoverStyleProps = {}) => {
...
-  return {
-    flip,
-    gutter,
-    matchWidth,
-    placement,
-    strategy,
-  }
+  return useMemo(
+    () => filterUndefined({ flip, gutter, matchWidth, placement, strategy }),
+    [flip, gutter, matchWidth, placement, strategy],
```
codex-for-yamada-ui, line 128: **Issue:** The new `filterUndefined` is applied only to `styleProps`. If a consumer passes an explicit `undefined` (e.g. `placement={undefined}`), that key still exists on `props`, and this function does not remove it from the object later passed to root prop merging. So the change does not fully guarantee the stated behavior for explicit `undefined`.

**Fix:** Filter `undefined` on the final object passed into root prop resolution (not only the derived style object), e.g. apply `filterUndefined` to the merged `{ ...props, ...styleProps }` payload before calling `useRootComponentProps`.

packages/react/src/components/popover/popover.tsx
codex-for-yamada-ui, line 129: **Issue:** This is a behavior change in prop-resolution semantics, but no regression test was added in the diff to prove that consumer defaults are preserved when explicit `undefined` is passed.

**Fix:** Add a focused regression test in `packages/react/src/components/popover/popover.test.tsx` that sets a consumer default for popover style props and asserts it is not cleared by explicitly passing `undefined`.

codex-for-yamada-ui (dismissed):
- Correctness: The new `filterUndefined` is applied only to `styleProps`. If a consumer passes an explicit `undefined` (e.g. `placement={undefined}`), that key still exists on `props`, and this function does not remove it from the object later passed to root prop merging. So the change does not fully guarantee the stated behavior for explicit `undefined`.
- Testing: This is a behavior change in prop-resolution semantics, but no regression test was added in the diff to prove that consumer defaults are preserved when explicit `undefined` is passed.

packages/react/src/components/popover/popover.tsx
```diff
@@ -123,13 +123,10 @@ export const usePopoverStyleProps = (props: UsePopoverStyleProps = {}) => {
...
 
-  return {
-    flip,
-    gutter,
-    matchWidth,
-    placement,
-    strategy,
-  }
+  return useMemo(
+    () => filterUndefined({ flip, gutter, matchWidth, placement, strategy }),
```
composer-for-yamada-ui, line 127: **Issue:** `usePopoverStyleProps` の挙動変更に対し、このPRの差分にはユニット／コンポーネントテストの追加・更新がなく、`popover.test.tsx` にも当該フックやスプレッド後のマージ挙動を検証するケースがない。

**Fix:** `mergeProps` が未定義値を上書きしない前提で、`placement` 等を省略せずに `undefined` が載っていた際にコンテキスト既定が残ること（および関連する `flip` / `gutter` など）を `popover.test.tsx` でレンダリングまたはフック経由で検証するテストを追加する。

.changeset/popover-filter-undefined-style-props.md
```diff
@@ -0,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+`usePopoverStyleProps` now drops undefined entries so consumer-provided defaults survive when an explicit `undefined` prop is passed in.
```
composer-for-yamada-ui: **Issue:** チェンジセットは undefined のみ除去と読めるが、利用している `filterUndefined` は `@yamada-ui/utils` で `null` と `undefined` の両方を落とすため、公開説明と実装が一致しない。

**Fix:** チェンジセット本文を「`null` と `undefined` のエントリを除去する」旨に修正するか、意図が undefined のみなら `undefined` のみを除去するヘルパーに差し替える。

composer-for-yamada-ui (dismissed):
- テスト: `usePopoverStyleProps` の挙動変更に対し、このPRの差分にはユニット／コンポーネントテストの追加・更新がなく、`popover.test.tsx` にも当該フックやスプレッド後のマージ挙動を検証するケースがない。
- 正確性: チェンジセットは undefined のみ除去と読めるが、利用している `filterUndefined` は `@yamada-ui/utils` で `null` と `undefined` の両方を落とすため、公開説明と実装が一致しない。

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7012, opened 2026-05-01
fix(date-picker): branch onKeyDown by mode and fix locale handling

hirotomoyamada:
Closes #

## AI used

## Description

Fix two bugs in `DatePicker`'s `onKeyDown` handler and `stringToDate` helper that caused Enter and Backspace to misbehave when no value was set yet, and that mishandled the `locale` argument when parsing input strings.

## Current behavior (updates)

- `onKeyDown` branched on the current `value` shape (`isDate` / `isArray` / `isObject`). When no value was set yet (e.g. `value` is `undefined` for single mode, or the calendar was just cleared), Enter and Backspace fell through to the wrong branch and did nothing.
- `stringToDate` called `dayjs(value, locale)`, which mistakenly interprets the second argument as a parsing format string instead of a locale. Localized inputs were not parsed correctly.

## New behavior

- `onKeyDown` now branches on the actual mode (`range` / `multiple` / `single`) so Enter and Backspace work even when no value is set yet:
  - **Enter (single)**: parses the typed string and sets the value, even when the previous value was `undefined`.
  - **Enter (range / multiple)**: unchanged behavior, but no longer depends on `value` already being set.
  - **Backspace (single)**: lets the input handle the keystroke as normal text editing (no `preventDefault`).
  - **Backspace (range / multiple)**: unchanged behavior (clears last item / both range inputs when the input is empty).
- `stringToDate` now applies the locale via `dayjs(value).locale(locale)` so the locale is honored when parsing.
- Tests updated to reflect the new behavior:
  - Removed the redundant second `_user.click(input)` from Enter tests; the second click was inadvertently closing the calendar via the combobox's `onClick`, which made the `!open` early-return swallow the Enter handler.
  - Switched the multiple-mode Enter test and blur tests to focus the input directly, and disabled `openOnChange/Click/Focus` for blur tests so that tabbing actually leaves the field instead of being intercepted by the calendar.
  - Updated the range field click tests to invoke `fieldEl.click()` directly so `ev.target` is the field rather than a child input.
  - Renamed the Backspace single-date test and updated its expectation to reflect normal text editing.

No.

## Additional Information

All 288 date-picker browser tests pass across chromium / firefox / webkit.


claude-for-yamada-ui (approved):

packages/react/src/components/date-picker/use-date-picker.tsx
```diff
@@ -307,7 +307,8 @@ export const useDatePicker = <
     (value: string, start: Date = minDate, end: Date = maxDate) => {
       let date = parseDate?.(value)
 
-      if (!date && dayjs(value).isValid()) date = dayjs(value, locale).toDate()
+      if (!date && dayjs(value).isValid())
+        date = dayjs(value).locale(locale).toDate()
```
codex-for-yamada-ui, line 311: **Issue:** The PR changes the locale parsing fallback path (`dayjs(value).locale(locale)`), but there is no added/updated regression test that verifies this locale-dependent behavior. The touched `date-picker.test.tsx` changes cover Enter/Backspace/focus flows with default-locale-style inputs, not the new locale parsing branch.

**Fix:** Add a focused test in `packages/react/src/components/date-picker/date-picker.test.tsx` that uses a non-default locale and verifies Enter commits the value through this fallback path (input -> parse -> formatted value update), so this bug cannot regress silently.

codex-for-yamada-ui (changes requested):
- テスト: The PR changes the locale parsing fallback path (`dayjs(value).locale(locale)`), but there is no added/updated regression test that verifies this locale-dependent behavior. The touched `date-picker.test.tsx` changes cover Enter/Backspace/focus flows with default-locale-style inputs, not the new locale parsing branch.

packages/react/src/components/date-picker/use-date-picker.tsx
```diff
@@ -624,6 +623,15 @@ export const useDatePicker = <
...
               if (align === "start") endInputRef.current?.focus()
+            } else {
+              const date = stringToDate(inputValue)
+
+              if (!date) return
+
+              ev.preventDefault()
+
+              setInputValue(dateToString(date) as MaybeInputValue<Range>)
+              setValue(date as MaybeDateValue<Multiple, Range>)
```
composer-for-yamada-ui, line 634: **Issue:** 単一選択の `Enter` ハンドラの `else` で `setValue(date)` に置き換わっており、複数選択時の `Enter`（`setValue` 内で `updateMaybeDateValue(date, max)(prev)` を使う）と異なり、既存の `Date` を保持したまま同じ日付文字列を確定した場合に `updateMaybeDateValue` が行う `isSameDate` チェックによる選択解除（`undefined` へのトグル）が失われる。

**Fix:** `setValue(date …)` を `setValue((prev) => updateMaybeDateValue(date, max)(prev) as MaybeDateValue<Multiple, Range>)` に変更し、キーボード確定でも `updateMaybeDateValue` の単一値分岐と揃える。意図的な挙動変更なら変更内容とユーザー影響を changeset に明記すること。

composer-for-yamada-ui (dismissed):
- 正確性: 単一選択の `Enter` ハンドラの `else` で `setValue(date)` に置き換わっており、複数選択時の `Enter`（`setValue` 内で `updateMaybeDateValue(date, max)(prev)` を使う）と異なり、既存の `Date` を保持したまま同じ日付文字列を確定した場合に `updateMaybeDateValue` が行う `isSameDate` チェックによる選択解除（`undefined` へのトグル）が失われる。

composer-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

PR #7287, opened 2026-05-04
test(button): recategorize and clean up browser tests

hirotomoyamada:
Closes #7169

## AI used

- [ ] I did not use AI to create this PR.
- [x] (If there is no check above) I checked the generated content before submitting.

## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/button` according to the rule introduced in #7139.

## Current behavior (updates)

Three `*.test.browser.{ts,tsx}` files lived under `packages/react/src/components/button/`:

- `button.test.browser.tsx` — 10 tests, all jsdom-friendly except `a11y`.
- `button-group.test.browser.tsx` — 5 tests, only the `getComputedStyle` assertion needed a real browser.
- `icon-button.test.browser.tsx` — 5 tests, all jsdom-friendly except `a11y`.

Several of these tests did not contribute to coverage (pure metadata reads such as `Button.displayName`, or repeated render assertions covered by sibling cases).

## New behavior

Each test was re-categorized using the five-axis checklist in [`test-categorization.md`](.agents/rules/test-categorization.md).

**jsdom (`#test`)**

- `button.test.tsx` — 7 tests (a11y, icon, loading, aria, type, no type, disabled)
- `button-group.test.tsx` — 1 test (a11y)
- `icon-button.test.tsx` — 2 tests (a11y, children)

**browser (`#test/browser`)**

- `button-group.test.browser.tsx` — 1 test (`getComputedStyle` for the `attached` column style)
- `button.test.browser.tsx` and `icon-button.test.browser.tsx` are removed because no browser-only assertion remained.

Removed tests that did not contribute to coverage:

- `sets displayName correctly` × 3 (no rendering, pure metadata read)
- `renders HTML tag correctly` × 3 (same render path as `sets className correctly`)
- `sets className correctly` × 3 (covered by sibling assertions in the same render tree)

## Is this a breaking change (Yes/No):

No. Test-only change. Public API unchanged.

## Additional Information

Verified locally:

- `pnpm react test:jsdom --run src/components/button/` — 10 tests pass
- `pnpm react test:browser --run src/components/button/` — 1 test × 3 engines pass
- Coverage on `packages/react/src/components/button/` unchanged at 96% statements / 88.09% branches / 100% functions / 95.91% lines.

Sub-issue of #7141.

PR #7288, opened 2026-05-04
fix(vitest): use istanbul coverage for browser projects to support firefox and webkit

hirotomoyamada:
Closes #

## AI used

## Description

Use `istanbul` as the coverage provider for the React Browser Mode projects (`browser:chromium`, `browser:firefox`, `browser:webkit`) while keeping `v8` for everything else (jsdom, cli, utils). Vitest's `v8` coverage relies on Chrome DevTools Protocol, which is only available in Chromium — so `vitest --project=browser:firefox --coverage` (and `:webkit`) crashes during prepare/run with `browserContext.newCDPSession: CDP session is only available in Chromium`. `istanbul` instruments source via Babel and does not need CDP, so it works for every browser instance.

## Current behavior (updates)

- The shared Vitest config (`packages/workspace/src/vitest/config.ts`) sets `coverage.provider = "v8"`, and the React `test:coverage:{chromium,webkit,firefox}` scripts run `vitest --coverage` against a single browser project.
- Before the per-browser project split, three browser instances ran inside one project, so `v8` silently skipped non-Chromium instances (only Chromium contributed coverage). After the split each browser runs alone, and `--coverage` against Firefox/WebKit now fails outright (e.g. PR #7287's Firefox 1/4 / Webkit 1/4 failed once a `*.test.browser.tsx` landed in shard 1).
- jsdom / cli / utils still benefit from `v8` (faster, no Babel instrumentation, stable historical coverage values).

## New behavior

- Keep `provider: "v8"` as the default in `packages/workspace/src/vitest/config.ts` so jsdom, cli, and utils continue to use v8.
- Override `--coverage.provider=istanbul` in each browser-specific script in `packages/react/package.json` (`test:coverage:chromium` / `:firefox` / `:webkit`). Vitest's coverage settings are workspace-wide, so the override is done at the CLI flag level rather than inside `defineProject`.
- Add `@vitest/coverage-istanbul` alongside the existing `@vitest/coverage-v8` in the root `package.json`.

Verified locally:

- `pnpm react test:coverage:jsdom --run src/utils/ref.test.ts` → `Coverage enabled with v8`
- `pnpm react test:coverage:chromium --run src/components/button/button-group.test.browser.tsx` → `Coverage enabled with istanbul`
- `pnpm react test:coverage:firefox --run src/components/button/button-group.test.browser.tsx` → `Coverage enabled with istanbul` (was failing with CDP error before)
- `pnpm react test:coverage:webkit --run src/components/button/button-group.test.browser.tsx` → `Coverage enabled with istanbul` (was failing with CDP error before)

No.

## Additional Information

- `codecov.yml` and `.github/workflows/codecov.yml` are untouched. Both providers emit `coverage-final.json` in the same shape, and the existing `unittests` flag merges per-line OR across jobs server-side.
- Coverage numbers from browser jobs will shift slightly (istanbul vs. v8 statement/branch counting), but jsdom / cli / utils numbers remain on the v8 baseline so historical comparisons there are intact.
- Fixes the failing Firefox 1/4 and Webkit 1/4 jobs observed on PR #7287.


claude-for-yamada-ui (dismissed):

pnpm-lock.yaml
```diff
@@ -4394,8 +4397,13 @@ packages:
     peerDependencies:
       vitest: 4.1.4
 
+  '@vitest/coverage-istanbul@4.1.5':
+    resolution: {integrity: sha512-X4kQMDEWh9mA0IiLuigtdYv4kXe+W8KLTbucoz15lbyZRPAxT5l+hu0JizI7Am050+G9vQnB7QJNgYi2LnwV4w==, tarball: https://registry.npmjs.org/@vitest/coverage-istanbul/-/coverage-istanbul-4.1.5.tgz}
+    peerDependencies:
+      vitest: 4.1.5
```
codex-for-yamada-ui, line 4403: **Issue:** This PR adds `@vitest/coverage-istanbul@4.1.5` with peer requirement `vitest: 4.1.5`, while the same lockfile still resolves `vitest` to 4.1.4 in the importer graph. That creates a peer-version mismatch in the test tooling.

**Fix:** Align Vitest versions. Either upgrade `vitest` (and related `@vitest/*` packages) to 4.1.5, or pin `@vitest/coverage-istanbul` to a 4.1.4-compatible version so the peer requirement exactly matches.

codex-for-yamada-ui (changes requested):
- Correctness: This PR adds `@vitest/coverage-istanbul@4.1.5` with peer requirement `vitest: 4.1.5`, while the same lockfile still resolves `vitest` to 4.1.4 in the importer graph. That creates a peer-version mismatch in the test tooling.

packages/react/package.json
```diff
@@ -71,9 +71,9 @@
     "test:dev:firefox": "vitest --watch --ui --project=browser:firefox",
     "test:coverage": "vitest run --coverage",
```
composer-for-yamada-ui, line 72: **Issue:** `test:coverage` runs all Vitest projects (see `vitest.config.ts` `projects`: jsdom plus `browser:*`) but, unlike `test:coverage:chromium` / `webkit` / `firefox` on lines 74–76, it does not pass `--coverage.provider= istanbul`. The aggregated script therefore does not share the same coverage setup as the browser-specific coverage scripts changed in this PR.

**Fix:** Align `test:coverage` with browser coverage (for example run separate `vitest` invocations for jsdom vs browser with the appropriate provider, set `coverage.provider` per project in `vitest.config.ts`, or document that full-repo coverage must not use `test:coverage` and only the per-target scripts are supported).

pnpm-lock.yaml
composer-for-yamada-ui, line 4403: **Issue:** The added `packages` entry for `@vitest/coverage-istanbul@4.1.5` declares `peerDependencies: vitest: 4.1.5`, while the lockfile still resolves `vitest` to `4.1.4` (see the same diff’s snapshot ` '@vitest/coverage-istanbul@4.1.5(vitest@4.1.4)':` wiring `vitest: 4.1.4(...)`). The workspace therefore depends on a coverage plugin whose declared peer range does not match the installed Vitest version.

**Fix:** Pin `@vitest/coverage-istanbul` to a release that matches Vitest `4.1.4` (for example constraint `4.1.4` in root `package.json` next to `@vitest/coverage-istanbul`), or bump Vitest (catalog) and related `@vitest/*` packages to `4.1.5` so peers align; then regenerate the lockfile.

composer-for-yamada-ui (dismissed):
- Consistency: `test:coverage` runs all Vitest projects (see `vitest.config.ts` `projects`: jsdom plus `browser:*`) but, unlike `test:coverage:chromium` / `webkit` / `firefox` on lines 74–76, it does not pass `--coverage.provider= istanbul`. The aggregated script therefore does not share the same coverage setup as the browser-specific coverage scripts changed in this PR.
- Consistency: The added `packages` entry for `@vitest/coverage-istanbul@4.1.5` declares `peerDependencies: vitest: 4.1.5`, while the lockfile still resolves `vitest` to `4.1.4` (see the same diff’s snapshot ` '@vitest/coverage-istanbul@4.1.5(vitest@4.1.4)':` wiring `vitest: 4.1.4(...)`). The workspace therefore depends on a coverage plugin whose declared peer range does not match the installed Vitest version.

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

Issue #7141, opened 2026-05-04
Recategorize browser tests in packages/react and remove or consolidate redundant cases (1/2)


hirotomoyamada:
## Description

Recategorize the existing 181 `*.test.browser.{ts,tsx}` files in `packages/react/` according to the test categorization rule introduced in #7139, and along the way prune redundant tests and consolidate test files so each component owns a single test file.

This is part 1 of 2. Part 2 is tracked in #7286. The split exists because GitHub caps a parent issue at 100 sub-issues, and the per-folder breakdown produces 130 sub-issues. This part covers `components/accordion` through `components/slider`. The rest is in #7286.

## Problem Statement / Justification

PR #7139 adds [.agents/rules/test-categorization.md](https://github.com/yamada-ui/yamada-ui/blob/main/.agents/rules/test-categorization.md) as the single source of truth for choosing between jsdom and browser test environments, and enforces the `#test` ↔ `#test/browser` import boundary at lint time. The PR explicitly leaves the migration of the existing browser-suffixed files out of scope, deferring it to follow-up PRs grouped per folder.

This issue tracks that follow-up work. Beyond the recategorization itself, two related cleanups are folded in to avoid touching the same files multiple times:

- Tests that do not contribute to coverage (for example duplicated `displayName` assertions, or checks already covered elsewhere) should be removed.
- Test files should follow a "one component, one test file" rule. Internal hook tests like `use-accordion.test.ts` should be merged into the parent component's test file (`accordion`), since the hook is exercised through the component anyway. There are 24 such `use-*.test.*` files under `packages/react/src/components/` today.

## Proposed Solution

Create one sub-issue per folder under `packages/react/src/{core,hooks,components}/`. Each sub-issue:

- Re-categorizes its `*.test.browser.{ts,tsx}` files using the five-axis checklist in [test-categorization.md](https://github.com/yamada-ui/yamada-ui/blob/main/.agents/rules/test-categorization.md) (DOM read / event handling / observer / focus / browser API). Tests that do not need a real browser move to `*.test.{ts,tsx}` and `#test`.
- Merges sibling `use-*.test.*` files into the parent component's test file.
- Removes redundant tests and consolidates scattered ones inside the same file.

Companion: #7286. Depends on #7139.

## AI used

- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

PR #7139, opened 2026-05-04
docs(testing): add test categorization rule and enforce import boundaries

hirotomoyamada:
Closes #

## AI used

- [ ] I did not use AI to create this PR.
- [x] (If there is no check above) I checked the generated content before submitting.

## Description

Adds a categorization rule that tells contributors which Vitest project a `packages/react` test belongs in (jsdom / all-browser / single-browser), and enforces the `#test` ↔ `#test/browser` import boundary at lint time so the file-name suffix introduced in #7013 cannot drift from the helper actually imported.

## Current behavior (updates)

- [.agents/rules/browser-testing.md](.agents/rules/browser-testing.md) lists the four file-name suffixes but does not say *when* to choose each, and the per-browser variants (`*.test.{chromium,firefox,webkit}.*`) have no usage guidance. Contributors have to infer the categorization from the existing inventory.
- The `#test` / `#test/browser` import boundary exists only as convention; nothing fails CI when a jsdom test imports `#test/browser` (or vice versa).
- `createNoRestrictedImportConfig` in [packages/workspace/src/oxlint/import.ts](packages/workspace/src/oxlint/import.ts) takes `[directory, ...patterns][]` and applies the patterns to `<directory>/**/*.{ts,cts,mts,tsx}`. It cannot target a subset of files inside a directory (e.g. only browser-test files).

## New behavior

- New [.agents/rules/test-categorization.md](.agents/rules/test-categorization.md): single source of truth for which file-name to choose, with a five-axis checklist (DOM read / event handling / observer / focus / browser API) and per-browser engine-branching guidance.
- [.agents/rules/browser-testing.md](.agents/rules/browser-testing.md) lead-in updated so each per-browser variant explains *when to use it*, and points at test-categorization.md for the full decision.
- [AGENTS.md](AGENTS.md) (= `CLAUDE.md` symlink) rule index gains a `Test Categorization` entry.
- `createNoRestrictedImportConfig` signature changes from `[directory, ...patterns][]` to `[files, ...patterns][]`, so callers pass the file glob list directly. This unlocks targeting test-file subsets.
- New `sharedUnitTestFiles` and `sharedBrowserTestFiles` exports in [packages/workspace/src/oxlint/shared.ts](packages/workspace/src/oxlint/shared.ts). `sharedTestFiles` is preserved as the union for future reuse.
[...]

PR #7344, opened 2026-05-05
test(alert): recategorize and clean up browser tests

hirotomoyamada:
Closes #7159

## AI used

- [ ] I did not use AI to create this PR.
- [x] (If there is no check above) I checked the generated content before submitting.

## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/alert` according to the rule introduced in #7139.

## Current behavior (updates)

One `*.test.browser.{ts,tsx}` file lived under `packages/react/src/components/alert/`:

- `alert.test.browser.tsx` — 9 tests. None of the assertions read the real DOM, dispatched real events, used observers/focus, or branched on the engine.

Several of these tests did not contribute to coverage (pure metadata reads such as `Alert.Root.displayName`, repeated render assertions covered by sibling cases, and a `role="alert"` assertion already exercised by the default-`colorScheme` test).

## New behavior

Each test was re-categorized using the five-axis checklist in [`test-categorization.md`](.agents/rules/test-categorization.md). `a11y(...)` was moved to jsdom because the assertion is a pure render + axe pass.

**jsdom (`#test`)**

- `alert.test.tsx` — 5 tests (a11y, default `loadingScheme`, custom `loadingScheme`, default `colorScheme` from `status`, `Alert.Icon` `as` prop)

**browser (`#test/browser`)**

- `alert.test.browser.tsx` is removed because no browser-only assertion remained.

Removed tests that did not contribute to coverage (verified empirically by deleting each candidate and re-running coverage):

- `sets displayName correctly` (no rendering, pure metadata read)
- `sets className correctly` (className covered by sibling render assertions in the same file)
- `renders HTML tag correctly` (same render path as sibling tests)
- `should have role='alert'` (`role="alert"` already asserted by `uses status as default colorScheme when colorScheme is not provided`)

## Is this a breaking change (Yes/No):

No. Test-only change. Public API unchanged.

## Additional Information

Verified locally:

- `pnpm react test:jsdom --run src/components/alert/` — 5 tests pass
- `pnpm react test:browser --run src/components/alert/` — no browser tests remain for this component (every assertion runs in jsdom)
- `pnpm react lint` — passes (no `#test` ↔ `#test/browser` boundary violations)

Coverage on `packages/react/src/components/alert/` (per source file, line / branch / function / statement; combined = covered in either project):

| File | Project | Baseline | Final |
| --- | --- | --- | --- |
[...]

PR #7349, opened 2026-05-05
test(badge): recategorize and clean up browser tests

### LINKED ISSUE #7164: Recategorize and clean up browser tests for the `badge` component
## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/badge` according to the rule from #7139.

## Files in scope

- `packages/react/src/components/badge/badge.test.browser.tsx`

## Work to do

   ```sh
   pnpm react test:coverage:jsdom --coverage.include='src/components/badge/**' --coverage.reporter=text src/components/badge/
   pnpm react test:coverage:chromium --coverage.include='src/components/badge/**' --coverage.reporter=text src/components/badge/
   ```

## Acceptance criteria

- `pnpm react test:jsdom --run src/components/badge/` passes
- `pnpm react test:browser --run src/components/badge/` passes
- Per-source-file combined coverage (line / branch / function / statement) on `packages/react/src/components/badge/` is at least equal to the baseline captured in step 1, with both numbers posted in the PR description

## AI used

hirotomoyamada:
Closes #7164

## AI used

## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/badge` according to the rule introduced in #7139.

## Current behavior (updates)

One `*.test.browser.{ts,tsx}` file lived under `packages/react/src/components/badge/`:

- `badge.test.browser.tsx` — 3 tests, all jsdom-friendly (no DOM read, no event handling, no observer, no focus, no browser API).

The original tests were:

- `sets displayName correctly` — pure metadata read on `Badge.displayName`.
- `sets className correctly` — `toHaveClass("ui-badge")` attribute check.
- `renders HTML tag correctly` — `tagName` check.

## New behavior

Each test was re-categorized using the five-axis checklist in [`test-categorization.md`](.agents/rules/test-categorization.md). Since none of the original assertions required a real browser engine, all tests were eligible to move to jsdom.

**jsdom (`#test`)**

- `badge.test.tsx` — 1 test (`renders component correctly` via `a11y(<Badge>Badge</Badge>)`).

**browser (`#test/browser`)**

- `badge.test.browser.tsx` is removed because no browser-only assertion remained.

- `sets displayName correctly` (no rendering, pure metadata read; even an empty test file produced 100% coverage on this component because the import alone executes the module).
- `sets className correctly` (covered by the same render path as the remaining a11y test).
- `renders HTML tag correctly` (same render path as `sets className correctly`).

Empirical verification: replacing the file with a minimal `expect(Badge).toBeDefined()` test still yielded 100% statements / 100% lines / 100% functions / 100% branches on both `badge.tsx` and `badge.style.ts`. Replacing it with a single `a11y(...)` test under jsdom maintains the same combined coverage.

## Additional Information

Verified locally:

- `pnpm react test:jsdom --run src/components/badge/` — 1 test passes
- `pnpm react test:browser --run src/components/badge/` — no test files (expected, browser file deleted)
- `pnpm react lint` — 0 warnings, 0 errors

### Coverage (per source file, line / branch / function / statement)

| File             | Project  | Baseline                 | Final                    |
| ---------------- | -------- | ------------------------ | ------------------------ |
| `badge.style.ts` | jsdom    | 0% / 100% / 100% / 0%    | 100% / 100% / 100% / 100% |
| `badge.tsx`      | jsdom    | 0% / 100% / 100% / 0%    | 100% / 100% / 100% / 100% |
| `badge.style.ts` | chromium | 100% / 100% / 100% / 100% | 0% / 100% / 100% / 0%    |
| `badge.tsx`      | chromium | 100% / 100% / 100% / 100% | 0% / 100% / 100% / 0%    |

Combined per-source-file coverage (covered in either project): 100% / 100% / 100% / 100% — equal to baseline.

Sub-issue of #7141.


packages/react/src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { a11y } from "#test"
+import { Badge } from "./"
+
+describe("<Badge />", () => {
+  test("renders component correctly", async () => {
+    await a11y(<Badge>Badge</Badge>)
+  })
+})
```
claude-for-yamada-ui: **Issue:** The three previously existing assertions (displayName === "Badge", toHaveClass("ui-badge"), tagName === "SPAN") are not retained. The test-categorization rule explicitly lists displayName / className / element attribute checks as the kind that should go to jsdom — not be deleted. Sibling recategorization PRs in this same series keep them in jsdom: e.g. alert.test.tsx (PR #7344) and action-bar.test.tsx (PR #7342) both retain "sets displayName correctly" and "sets className correctly". The new file leaves only a11y(<Badge>Badge</Badge>), and a11y only runs axe — it does not assert displayName, className, or rendered tag, so a regression that renames the class from ui-badge, changes the rendered tag away from span, or breaks Badge.displayName would no longer be caught by any test in the repository. The PR description's "100% line coverage even with an empty file" argument conflates line coverage with contract coverage; these are public-API regression guards, not coverage filler.

**Fix:** Restore the structural tests in jsdom form, mirroring the pattern used in alert.test.tsx / action-bar.test.tsx. Add inside the existing describe("<Badge />"): "sets displayName correctly" (expect Badge.displayName toBe "Badge"), "sets className correctly" (render Badge then assert toHaveClass "ui-badge"), and "renders HTML tag correctly" (render Badge then assert tagName toBe "SPAN"). Import render and screen from #test alongside the existing a11y import.

claude-for-yamada-ui (changes requested):
- Consistency / Test: The three previously existing assertions (displayName === "Badge", toHaveClass("ui-badge"), tagName === "SPAN") are not retained. The test-categorization rule explicitly lists displayName / className / element attribute checks as the kind that should go to jsdom — not be deleted. Sibling recategorization PRs in this same series keep them in jsdom: e.g. alert.test.tsx (PR #7344) and action-bar.test.tsx (PR #7342) both retain "sets displayName correctly" and "sets className correctly". The new file leaves only a11y(<Badge>Badge</Badge>), and a11y only runs axe — it does not assert displayName, className, or rendered tag, so a regression that renames the class from ui-badge, changes the rendered tag away from span, or breaks Badge.displayName would no longer be caught by any test in the repository. The PR description's "100% line coverage even with an empty file" argument conflates line coverage with contract coverage; these are public-API regression guards, not coverage filler.

packages/react/src/components/badge/badge.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { a11y } from "#test"
+import { Badge } from "./"
+
+describe("<Badge />", () => {
+  test("renders component correctly", async () => {
+    await a11y(<Badge>Badge</Badge>)
```
codex-for-yamada-ui, line 22: **Issue:** This replacement keeps only an a11y smoke test and drops prior assertions for `displayName`, root `className` (`ui-badge`), and rendered tag (`SPAN`). With no production change in `Badge`, the diff reduces regression protection instead of preserving equivalent coverage.

**Fix:** Keep the jsdom move, but restore the removed assertions in this new file (displayName, className, and root tag) so test coverage remains equivalent to the deleted browser test.

codex-for-yamada-ui (changes requested):
- テスト: This replacement keeps only an a11y smoke test and drops prior assertions for `displayName`, root `className` (`ui-badge`), and rendered tag (`SPAN`). With no production change in `Badge`, the diff reduces regression protection instead of preserving equivalent coverage.

packages/react/src/components/badge/badge.test.tsx
composer-for-yamada-ui, line 22: **Issue:** Deleting `badge.test.browser.tsx` removed checks for `Badge.displayName`, the `ui-badge` class, and the default host tag (`SPAN`), while the new jsdom suite only calls `a11y` here. `.agents/rules/test-categorization.md` says assertions on `displayName`, `className`, and similar attributes belong in `*.test.tsx`, not browser tests—so those cases should move to jsdom, not disappear. Compared with nearby patterns (for example `toggle-group.test.tsx`, which pairs `a11y` with explicit `displayName` and `className` tests), this change leaves previously covered contracts without any test.

**Fix:** Add jsdom tests in `badge.test.tsx` (via `#test`’s `render` / `screen` or equivalent) that reintroduce the removed assertions: `displayName`, root `className` including `ui-badge`, and the expected default element type, keeping them separate and narrow per unit-testing guidance.

composer-for-yamada-ui (changes requested):
- テスト: Deleting `badge.test.browser.tsx` removed checks for `Badge.displayName`, the `ui-badge` class, and the default host tag (`SPAN`), while the new jsdom suite only calls `a11y` here. `.agents/rules/test-categorization.md` says assertions on `displayName`, `className`, and similar attributes belong in `*.test.tsx`, not browser tests—so those cases should move to jsdom, not disappear. Compared with nearby patterns (for example `toggle-group.test.tsx`, which pairs `a11y` with explicit `displayName` and `className` tests), this change leaves previously covered contracts without any test.

packages/react/src/components/badge/badge.test.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/badge/badge.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7398, opened 2026-05-05
test(indicator): recategorize and clean up browser tests

### LINKED ISSUE #7212: Recategorize and clean up browser tests for the `indicator` component
## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/indicator` according to the rule from #7139.

## Files in scope

- `packages/react/src/components/indicator/indicator.test.browser.tsx`

## Work to do

   ```sh
   pnpm react test:coverage:jsdom --coverage.include='src/components/indicator/**' --coverage.reporter=text src/components/indicator/
   pnpm react test:coverage:chromium --coverage.include='src/components/indicator/**' --coverage.reporter=text src/components/indicator/
   ```

## Acceptance criteria

- `pnpm react test:jsdom --run src/components/indicator/` passes
- `pnpm react test:browser --run src/components/indicator/` passes
- Per-source-file combined coverage (line / branch / function / statement) on `packages/react/src/components/indicator/` is at least equal to the baseline captured in step 1, with both numbers posted in the PR description

## AI used

hirotomoyamada:
Closes #7212

## AI used

## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/indicator` according to the rule introduced in #7139.

## Current behavior (updates)

`packages/react/src/components/indicator/indicator.test.browser.tsx` held 17 tests, but only two genuinely needed a real browser:

- `should apply ping prop with object configuration` reads computed CSS variables via `getComputedStyle`.
- `should apply offset prop` reads computed CSS variables via `getComputedStyle`.

Every other test asserted on attributes, slot `className`, presence in the DOM, or `data-*` props — all of which jsdom resolves correctly. The `a11y(...)` check was also imported from `#test/browser`, but the rule places `a11y` in jsdom by default.

The file additionally contained two tests that did not contribute to coverage:

- `sets displayName` — pure metadata read with no rendering.
- `should apply ping prop with boolean value` — the same render path as the `data-numeric` and slot-className tests.

## New behavior

Each test was re-categorized using the five-axis checklist in [`test-categorization.md`](.agents/rules/test-categorization.md).

**jsdom (`#test`)** — `indicator.test.tsx`, 13 tests:

- `renders component correctly` (a11y)
- `renders label and applies slot classNames`
- `renders numeric label with overflow indicator`
- `renders numeric label without overflow`
- `renders zero when showZero is true`
- `does not render label when showZero is false and label is 0`
- `does not render label when disabled`
- `sets data-numeric attribute for numeric labels`
- `does not set data-numeric attribute for non-numeric labels`
- `renders ping label inside the float wrapper`
- `renders the float wrapper containing the label for placement prop`
- `forwards floatProps to the float element`
- `forwards className through labelProps`

**browser (`#test/browser`)** — `indicator.test.browser.tsx`, 2 tests:

- `applies ping CSS variables when ping is an object` (`getComputedStyle`)
- `applies offset CSS variables` (`getComputedStyle`)

- `sets displayName` (pure metadata read, no render).
- `should apply ping prop with boolean value` (render path identical to the kept slot-className and data-numeric tests; no `getComputedStyle` assertion).

The `container.querySelector(...).toBeNull()` patterns were replaced with `expect(screen.queryByTestId(...)).not.toBeInTheDocument()`, and `data-testid` queries were swapped for `screen.getByText` / `screen.getByTestId` to follow the unit-testing rules. The `placement` test now asserts the wrapper relationship instead of touching `getComputedStyle`, since the offset variables are already covered by the dedicated browser test.

## Additional Information

Verified locally:

- `pnpm react test:jsdom --run src/components/indicator/` — 13 tests pass.
- `pnpm react test:browser --run src/components/indicator/` — 2 tests x 3 engines = 6 tests pass.
- `pnpm react lint` — passes.

Coverage on `packages/react/src/components/indicator/`:

| Project   | Stage    | Stmts            | Branches        | Funcs         | Lines            |
| --------- | -------- | ---------------- | --------------- | ------------- | ---------------- |
| jsdom     | baseline | 0% (0/17)        | 0% (0/13)       | 0% (0/3)      | 0% (0/17)        |
| chromium  | baseline | 100% (16/16)     | 100% (13/13)    | 100% (3/3)    | 100% (16/16)     |
| jsdom     | final    | 88.23% (15/17)   | 92.3% (12/13)   | 100% (3/3)    | 88.23% (15/17)   |
| chromium  | final    | 93.75% (15/16)   | 61.53% (8/13)   | 100% (3/3)    | 93.75% (15/16)   |

Combined per-source-file coverage (a line/branch covered in either project counts as covered) is 100% statements / 100% branches / 100% functions / 100% lines on `indicator.tsx`, equal to the baseline. The chromium project no longer exercises the overflow path (line 115) because it lives in the jsdom test, and the jsdom project does not exercise the `isObject(ping)` branch (lines 143-145) because it lives in the browser test — each gap is filled by the other project.

Sub-issue of #7141.

### DIFF
--- a/packages/react/src/components/indicator/indicator.test.browser.tsx
+++ b/packages/react/src/components/indicator/indicator.test.browser.tsx
@@ -1,143 +1,8 @@
-import { a11y, page, render } from "#test/browser"
+import { page, render } from "#test/browser"
 import { Indicator } from "."
 
 describe("<Indicator />", () => {
-  test("renders component correctly", async () => {
-    await a11y(
-      <Indicator label="New">
-        <div />
-      </Indicator>,
-    )
-  })
-
-  test("sets `displayName`", () => {
-    expect(Indicator.displayName).toBe("IndicatorRoot")
-  })
-
-  test("sets `className` correctly", async () => {
-    await render(
-      <Indicator data-testid="indicator" label="New">
-        <div />
-      </Indicator>,
-    )
-    const el = page.getByText("New").element()
-    expect(el).toHaveClass("ui-indicator__label")
-    expect(el.parentElement).toHaveClass("ui-indicator__float")
-    expect(el.parentElement?.parentElement).toHaveClass("ui-indicator__root")
-  })
-
-  test("should render", async () => {
-    await render(
-      <Indicator label="New">
-        <div />
-      </Indicator>,
-    )
-    await expect.element(page.getByText("New")).toBeInTheDocument()
-  })
-
-  test("should render (with overflowCount)", async () => {
-    await render(
-      <Indicator
-        label={100}
-        overflowCount={99}
-        labelProps={{ "data-testid": "indicator" }}
-      >
-        <div />
-      </Indicator>,
-    )
-    await expect.element(page.getByTestId("indicator")).toHaveTextContent("+")
-  })
-
-  test("should not render zero when showZero is false", async () => {
-    const { container } = await render(
-      <Indicator
-        label={0}
-        showZero={false}
-        labelProps={{ "data-testid": "indicator" }}
-      >
-        <div />
-      </Indicator>,
-    )
-    expect(container.querySelector('[data-testid="indicator"]')).toBeNull()
-  })
-
-  test("should not render", async () => {
-    const { container } = await render(
-      <Indicator
-        disabled
-        label="New"
-        labelProps={{ "data-testid": "indicator" }}
-      >
-        <div />
-      </Indicator>,
-    )
-    expect(container.querySelector('[data-testid="indicator"]')).toBeNull()
-  })
-
-  test("should render zero when showZero is true", async () => {
-    await render(
-      <Indicator label={0} showZero labelProps={{ "data-testid": "indicator" }}>
-        <div />
-      </Indicator>,
-    )
-    await expect.element(page.getByTestId("indicator")).toHaveTextContent("0")
-  })
-
-  test("should render numeric label without overflow", async () => {
-    const { container } = await render(
-      <Indicator
-        label={50}
-        overflowCount={99}
-        labelProps={{ "data-testid": "indicator" }}
-      >
-        <div />
-      </Indicator>,
-    )
-    await expect.element(page.getByTestId("indicator")).toHaveTextContent("50")
-    expect(container.textContent).not.toContain("+")
-  })
-
-  test("should set data-numeric attribute for numeric labels", async () => {
-    await render(
-      <Indicator label={10} labelProps={{ "data-testid": "indicator" }}>
-        <div />
-      </Indicator>,
-    )
-    await expect
-      .element(page.getByTestId("indicator"))
-      .toHaveAttribute("data-numeric")
-  })
-
-  test("should not set data-numeric attribute for non-numeric labels", async () => {
-    await render(
-      <Indicator label="New" labelProps={{ "data-testid": "indicator" }}>
-        <div />
-      </Indicator>,
-    )
-    await expect
-      .element(page.getByTestId("indicator"))
-      .not.toHaveAttribute("data-numeric")
-  })
-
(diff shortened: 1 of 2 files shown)


claude-for-yamada-ui (dismissed):

packages/react/src/components/indicator/indicator.test.browser.tsx
```diff
@@ -1,143 +1,8 @@
...
-    expect(label).toBeInTheDocument()
-    // Verify the label element exists (ping affects pseudo-element styling)
-    expect(label.parentElement).toHaveClass("ui-indicator__float")
-  })
-
-  test("should apply ping prop with object configuration", async () => {
+  test("applies ping CSS variables when ping is an object", async () => {
     await render(
       <Indicator
         data-testid="indicator-root"
```
codex-for-yamada-ui: **Issue:** This modified browser component test file still renders `<Indicator />` but no longer includes any `a11y(...)` check. That removes the browser-level accessibility regression guard that existed before, and it conflicts with the browser testing rule requiring `a11y(...)` in modified browser component test files that render component DOM.

**Fix:** Reintroduce at least one `a11y(...)` test case in `indicator.test.browser.tsx` for the default rendered state (and add additional state cases only if the accessible tree changes).

codex-for-yamada-ui (changes requested):
- Accessibility: This modified browser component test file still renders `<Indicator />` but no longer includes any `a11y(...)` check. That removes the browser-level accessibility regression guard that existed before, and it conflicts with the browser testing rule requiring `a11y(...)` in modified browser component test files that render component DOM.

packages/react/src/components/indicator/indicator.test.tsx
```diff
@@ -0,0 +1,158 @@
...
+describe("<Indicator />", () => {
+  test("renders component correctly", async () => {
+    await a11y(
+      <Indicator label="New">
+        <div />
+      </Indicator>,
+    )
+  })
+
+  test("renders label and applies slot classNames", () => {
```
composer-for-yamada-ui, line 17: **Issue:** The browser suite used to assert `Indicator.displayName === "IndicatorRoot"`, but that check was removed and not reintroduced in the new jsdom file, so the public `displayName` contract is no longer covered anywhere under `indicator/`. This sits next to other rendering tests in the same `describe` (starting at the next test on this line) and contradicts the repo rule that `displayName` assertions belong in `*.test.{ts,tsx}` (jsdom), not only in browser tests.

**Fix:** Add a small test in `indicator.test.tsx` (for example immediately after the `a11y` test or alongside the first structural render test) with `expect(Indicator.displayName).toBe("IndicatorRoot")` so the migration preserves prior coverage.

composer-for-yamada-ui (changes requested):
- テスト: The browser suite used to assert `Indicator.displayName === "IndicatorRoot"`, but that check was removed and not reintroduced in the new jsdom file, so the public `displayName` contract is no longer covered anywhere under `indicator/`. This sits next to other rendering tests in the same `describe` (starting at the next test on this line) and contradicts the repo rule that `displayName` assertions belong in `*.test.{ts,tsx}` (jsdom), not only in browser tests.

packages/react/src/components/indicator/indicator.test.browser.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/components/indicator/indicator.test.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7441, opened 2026-05-06
test(slot): recategorize and clean up browser tests

### LINKED ISSUE #7256: Recategorize and clean up browser tests for the `slot` component
## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/slot` according to the rule from #7139.

## Files in scope

- `packages/react/src/components/slot/slot.test.browser.tsx`

## Work to do

   ```sh
   pnpm react test:coverage:jsdom --coverage.include='src/components/slot/**' --coverage.reporter=text src/components/slot/
   pnpm react test:coverage:chromium --coverage.include='src/components/slot/**' --coverage.reporter=text src/components/slot/
   ```

## Acceptance criteria

- `pnpm react test:jsdom --run src/components/slot/` passes
- `pnpm react test:browser --run src/components/slot/` passes
- Per-source-file combined coverage (line / branch / function / statement) on `packages/react/src/components/slot/` is at least equal to the baseline captured in step 1, with both numbers posted in the PR description

Depends on #7139. Sub-issue of #7286.

## AI used

hirotomoyamada:
Closes #7256

## AI used

## Description

Recategorize, prune, and consolidate the browser tests in `packages/react/src/components/slot` according to the rule introduced in #7139.

## Current behavior (updates)

One `*.test.browser.{ts,tsx}` file lived under `packages/react/src/components/slot/`:

- `slot.test.browser.tsx` — 9 tests. None of them required a real browser engine: every assertion was a render plus attribute / text / error check, with no DOM read (`getBoundingClientRect`, `getComputedStyle`, `scrollTop`, ...), event handling, observer, focus, or browser API usage.

One test (`sets displayName correctly`) did not contribute to coverage (pure metadata read with no rendering).

## New behavior

Each test was re-categorized using the five-axis checklist in [`test-categorization.md`](.agents/rules/test-categorization.md). Every test in `slot.test.browser.tsx` is jsdom-friendly, so the file is removed and the surviving tests are migrated to a new `slot.test.tsx`.

**jsdom (`#test`)**

- `slot.test.tsx` — 7 tests
  - `renders component correctly` (a11y)
  - `merges props onto child element`
  - `renders null when Slot has multiple children without Slottable`
  - `renders with Slottable - merges child element props`
  - `renders null when Slottable child is not a valid element`
  - `throws when Slottable has multiple children elements`
  - `<Slottable />` `renders its children`

**browser (`#test/browser`)**

- `slot.test.browser.tsx` is removed because no browser-only assertion remained.

- `sets displayName correctly` (no rendering, pure metadata read)
- `renders null when Slot has no children` (covered by sibling `Children.count(children) > 1 ? Children.only(null) : null` branches in the same render path)

Each removal was verified empirically: deleting the test and re-running coverage held line / branch / function / statement coverage at 100%.

## Additional Information

Verified locally:

- `pnpm react test:jsdom --run src/components/slot/` — 7 tests pass
- `pnpm react test:browser --run src/components/slot/` — no test files found (expected; every test is jsdom-friendly)
- `pnpm react lint` — passes

Coverage on `packages/react/src/components/slot/`:

| Metric     | Baseline (browser) | Final (jsdom)    |
| ---------- | ------------------ | ---------------- |
| Statements | 100% (19/19)       | 100% (19/19)     |
| Branches   | 100% (14/14)       | 100% (14/14)     |
| Functions  | 100% (4/4)         | 100% (4/4)       |
| Lines      | 100% (18/18)       | 100% (18/18)     |

Sub-issue of #7141.

### DIFF
--- a/packages/react/src/components/slot/slot.test.tsx
+++ b/packages/react/src/components/slot/slot.test.tsx
@@ -38,6 +38,13 @@ describe("<Slot />", () => {
     spy.mockRestore()
   })
 
+  test("renders null when Slot has no children", () => {
+    const { container } = render(<Slot />)
+
+    const visibleContent = container.querySelector(":not([hidden])")
+    expect(visibleContent).toBeNull()
+  })
+
   test("renders with Slottable - merges child element props", () => {
     const Button: FC<
       HTMLAttributes<HTMLButtonElement> & {
--- a/packages/react/src/components/slot/slot.test.tsx
+++ b/packages/react/src/components/slot/slot.test.tsx
@@ -1,5 +1,5 @@
 import type { FC, HTMLAttributes, ReactNode } from "react"
-import { a11y, page, render } from "#test/browser"
+import { a11y, render, screen } from "#test"
 import { Slot, Slottable } from "./slot"
 
 describe("<Slot />", () => {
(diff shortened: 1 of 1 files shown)


claude-for-yamada-ui (dismissed):

packages/react/src/components/slot/slot.test.tsx
```diff
@@ -11,46 +11,34 @@ describe("<Slot />", () => {
...
 
-  test("renders null when Slot has no children", async () => {
-    const { container } = await render(<Slot />)
-    const visibleContent = container.querySelector(":not([hidden])")
-
-    expect(visibleContent).toBeNull()
-  })
-
-  test("renders with Slottable - merges child element props", async () => {
+  test("renders with Slottable - merges child element props", () => {
```
codex-for-yamada-ui, line 48: **Issue:** The migration removes the `renders null when Slot has no children` case without adding an equivalent assertion, so the `Slot` path where `slottable` is absent and `children` is empty (`Slot` fallback to `SlotClone` returning `null`) is no longer directly covered.

**Fix:** Re-add an equivalent jsdom test for `<Slot />` with no children (or add another test that explicitly exercises the same no-children fallback path and asserts no visible output).

codex-for-yamada-ui (changes requested):
- Test: The migration removes the `renders null when Slot has no children` case without adding an equivalent assertion, so the `Slot` path where `slottable` is absent and `children` is empty (`Slot` fallback to `SlotClone` returning `null`) is no longer directly covered.

packages/react/src/components/slot/slot.test.tsx
composer-for-yamada-ui, line 48: **Issue:** The suite no longer asserts that `<Slot />` with no children yields no visible content. `SlotClone` in `slot.tsx` still returns `null` when there is no valid child (`isValidElement` false and not multiple children), so the previous regression test matched production behavior and this diff removes that coverage without replacing it elsewhere in the diff.

**Fix:** Reintroduce a jsdom test equivalent to the removed case (render `<Slot />`, query `:not([hidden])` or assert container is empty as before) immediately after the multiple-children error test and before the Slottable merge test.

composer-for-yamada-ui (changes requested):
- Testing (その変更が新たに導入する分岐・契約・回帰経路が、既存のテストでカバーされているか): The suite no longer asserts that `<Slot />` with no children yields no visible content. `SlotClone` in `slot.tsx` still returns `null` when there is no valid child (`isValidElement` false and not multiple children), so the previous regression test matched production behavior and this diff removes that coverage without replacing it elsewhere in the diff.

packages/react/src/components/slot/slot.test.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7479, opened 2026-05-08
Remove timeout from `useClipboard` hook

hirotomoyamada:
Closes # 

## AI used

## Description

Remove timeout from `useClipboard` hook.

No

### DIFF
--- a/www/components/copy-button.tsx
+++ b/www/components/copy-button.tsx
@@ -17,7 +17,7 @@ export interface CopyButtonProps extends ButtonProps {
 
 export function CopyButton({ value, onClick, ...rest }: CopyButtonProps) {
   const t = useTranslations("component.copyButton")
-  const { copied, onCopy } = useClipboard(value, 4000)
+  const { copied, onCopy } = useClipboard(value)
 
   return (
     <Tooltip
--- a/www/components/llms-button-group.tsx
+++ b/www/components/llms-button-group.tsx
@@ -47,7 +47,7 @@ export function LlmsButtonGroup({
     sm: "end-center",
   })
   const t = useTranslations("component.llmsButtonGroup")
-  const { copied, onCopy } = useClipboard(md, 4000)
+  const { copied, onCopy } = useClipboard(md)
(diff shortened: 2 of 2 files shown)


claude-for-yamada-ui (approved):

www/components/copy-button.tsx
```diff
@@ -17,7 +17,7 @@ export interface CopyButtonProps extends ButtonProps {
 
 export function CopyButton({ value, onClick, ...rest }: CopyButtonProps) {
   const t = useTranslations("component.copyButton")
-  const { copied, onCopy } = useClipboard(value, 4000)
+  const { copied, onCopy } = useClipboard(value)
```
composer-for-yamada-ui, line 20: **Issue:** Dropping the second argument makes `useClipboard` fall through to `@yamada-ui/react` defaults (`packages/react/src/hooks/use-clipboard/index.ts` sets `timeout` to **1500** when omitted via `timeout = 1500`). The previous explicit `4000` kept the tooltip/icon “copied” state ~4s on these controls; now it resets ~2.5× faster (`LlmsButtonGroup` mirrors this at www/components/llms-button-group.tsx:50).

**Fix:** Decide whether shortened feedback is intentional. If readers still need longer confirmation, restore an explicit duration (for example `useClipboard(value, 4000)` or `useClipboard(value, { timeout: 4000 })`) in both touched call sites so behavior matches documented UX intent.

codex-for-yamada-ui (approved):

composer-for-yamada-ui (changes requested):
- Correctness: Dropping the second argument makes `useClipboard` fall through to `@yamada-ui/react` defaults (`packages/react/src/hooks/use-clipboard/index.ts` sets `timeout` to **1500** when omitted via `timeout = 1500`). The previous explicit `4000` kept the tooltip/icon “copied” state ~4s on these controls; now it resets ~2.5× faster (`LlmsButtonGroup` mirrors this at www/components/llms-button-group.tsx:50).

composer-for-yamada-ui (approved):

PR #7483, opened 2026-05-08
chore(www): run dev and build tasks with turbo

### LINKED ISSUE #7482: Use Turbo to run www dev and build tasks
## Description

Use Turborepo task configuration to run the `www` development and build tasks instead of relying on `npm-run-all` scripts in `www/package.json`.

The `www` package currently uses `run-p dev:*` to start `dev:next` and `dev:content` concurrently. Turborepo supports the `with` task option for long-running tasks that should run alongside each other, so this workflow can be represented in the existing `www/turbo.json` package configuration.

The same package also uses `run-s build:*` for production builds. The build order can be represented in Turbo directly by making the Next build task depend on content generation.

## Proposed Solution or API

- Change `www#dev` to invoke Turbo for the Next dev task.
- Mark `www#dev:next` and `www#dev:content` as persistent, uncached tasks.
- Configure `www#dev:next` with `with: ["www#dev:content"]` so Velite watch starts alongside Next dev.
- Change `www#build` to invoke Turbo for the Next build task.
- Configure `www#build:next` to depend on `www#build:content` so Velite content generation runs before `next build`.
- Remove `npm-run-all` from the `www` dependency graph.

## Alternatives

Keep `run-p dev:*` and `run-s build:*` in `www/package.json`.

Yes

## AI used

## Additional Information

Turbo documentation for `with`: https://turborepo.dev/docs/reference/configuration#with

taroj1205:
- Closes #7482

## AI used

## Description

Run the `www` development and build tasks through Turbo instead of using `run-p dev:*` and `run-s build:*`.

## Current behavior (updates)

`www#dev` uses `run-p` to start `dev:next` and `dev:content` concurrently.

`www#build` uses `run-s build:*` to run the package build scripts.

## New behavior

`www#dev` delegates to Turbo, and `www#dev:next` uses Turbo's `with` configuration to run `www#dev:content` alongside it as persistent uncached tasks.

`www#build` delegates to Turbo, and `www#build:next` explicitly depends on `www#build:content` so content generation still runs before the Next production build.

`npm-run-all` is removed from the `www` dependency graph.

No.

## Additional Information

Verified with:

- `pnpm www dev --dry=json`
- `pnpm www dev`
- `pnpm www build --dry=json`
- pre-commit hooks for `2dec47e7a1`

Not run:

- full `pnpm www build`


hirotomoyamada (commented):
Why do we need to switch to Turbo? I don't see the need for it.

Also, if you're going to do this, you should do the same for the build.

taroj1205:
@hirotomoyamada 
Good point. It is not strictly needed.

The benefit is that both `dev` and `build` can use Turbo for orchestration, so we can remove `npm-run-all` from `www`. I've updated the build script as well.

I have updated the PR and issue accordingly.

hirotomoyamada (changes requested):
Please resolve the conflict.

hirotomoyamada (approved):

PR #7536, opened 2026-05-12
test(use-field-sizing): migrate tests to browser mode

### LINKED ISSUE #7513: Migrate use-field-sizing tests to Vitest Browser Mode
### Description

Migrate the `use-field-sizing` hook tests to Vitest Browser Mode and drop the `getBoundingClientRect` mocks.

The following test file currently runs under jsdom via `#test`:

- [index.test.tsx](packages/react/src/hooks/use-field-sizing/index.test.tsx)

The hook relies on real element measurement via `getBoundingClientRect`. Under jsdom this returns zero, so the tests today spy on the prototype to feed fake widths. In a real browser the measurement works correctly and the spy is no longer needed — this matches the cleanup performed in #7503 for `slider` / `color-selector` / `rating`.

### Proposed Solution or API

1. Replace `import { ... } from "#test"` with `import { page, render } from "#test/browser"` (add `a11y` for cases that render component DOM).
2. Rewrite assertions to use Browser Mode APIs (`page.getBy*`, `await expect.element(...)`, `user.type(...)`).
3. Drop `vi.spyOn(HTMLElement.prototype, "getBoundingClientRect")` (and any similar layout spy). Size the test input via explicit `style` / props instead and assert against the real measured width.
4. Verify: `pnpm react test:browser --run src/hooks/use-field-sizing/`.

Reference: #7503 (the same mock-removal pattern applied to `slider` and `color-selector`).

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #7506

hirotomoyamada:
Closes #7513

## AI used

## Description

Migrated `use-field-sizing` tests from jsdom to Vitest Browser Mode and removed layout mocks.

## Current behavior (updates)

`use-field-sizing` tests were implemented as jsdom tests and relied on `getBoundingClientRect` spies.

## New behavior

- Replaced the jsdom test file with `index.test.browser.tsx`.
- Switched test APIs to `#test/browser` (`a11y`, `page`, `render`).
- Removed `getBoundingClientRect` spying and asserted width behavior from real browser measurements.
- Kept hidden text rendering, default value, width initialization, and rerender width update checks.

No

## Additional Information

Coverage check on `src/hooks/use-field-sizing/` (before/after):
- Before: `11/12` statements (`91.67%`)
- After: `11/12` statements (`91.67%`)
- Delta: `0.00%` (within ±5% tolerance)


packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <input ref={ref} aria-label="field sizing input" data-testid="input" />
+      {text}
+    </div>
+  )
+}
+
+describe("useFieldSizing", () => {
+  test("passes a11y checks", async () => {
+    await a11y(<Component value="hello" />)
+  })
```
claude-for-yamada-ui: **Issue:** The `a11y` test is placed in browser mode, but per `.agents/references/review-anti-patterns/a11y-in-browser-mode.md`, `a11y(...)` is a static DOM/attribute check that runs correctly under jsdom. The audited tree here (a `<div>` containing a labeled `<input>` and a hidden `<span>`) does not require real layout, real events, observers, focus, or browser-only APIs to be audited. Running it in browser mode triples its runtime cost (chromium/firefox/webkit) for no additional signal.

**Fix:** Move the a11y test back to a jsdom file (`index.test.tsx`) using `import { a11y } from "#test"`. Keep only the layout-dependent assertions (those reading `style.width` populated via `getBoundingClientRect`) in `index.test.browser.tsx`.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const span = page.getByText("hello")
+
+    await expect.element(span).toHaveAttribute("aria-hidden")
+    await expect.element(span).toHaveStyle({
+      opacity: "0",
+      overflow: "hidden",
+      position: "absolute",
+      whiteSpace: "nowrap",
+      zIndex: "-1",
+    })
```
claude-for-yamada-ui: **Issue:** The static contract assertions for the hidden span (`aria-hidden`, inline style: opacity/overflow/position/whiteSpace/zIndex) do not depend on any browser-mode criterion in `.agents/rules/test-categorization.md` (no live layout reads, no events, no observers, no focus, no browser APIs). They are attribute/style assertions that belong in jsdom. The same split is followed in nearby hooks such as `use-value` (pure logic in `*.test.tsx`, browser-dependent in `*.test.browser.tsx`).

**Fix:** Restore a jsdom test file (`index.test.tsx`) for the static-contract tests and keep `index.test.browser.tsx` for behavior that genuinely needs a real browser (input-width derivation from the measured span).

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const input = page.getByTestId("input")
+    const beforeWidth = Number.parseFloat(input.element().style.width)
+
+    rerender(<Component value="hello world" />)
+
+    await vi.waitFor(() => {
+      const afterWidth = Number.parseFloat(input.element().style.width)
+
+      expect(afterWidth).toBeGreaterThan(beforeWidth)
+    })
```
claude-for-yamada-ui, line 58: **Issue:** The migration weakens the sizing contract. The previous tests asserted concrete propagation (`input.style.width` equals the mocked span width: `100px` then `200px` after rerender). The new tests only assert `width > 0` and `after > before`, which can hold even if the input's width were not actually derived from the measured span (any positive non-decreasing source would pass). This is a coverage regression for the hook's central contract: "input width tracks the measured text width".

**Fix:** Tighten the assertion so the input width equals the measured span width — e.g., read the rendered span's `getBoundingClientRect().width` (or `offsetWidth`) and assert `Number.parseFloat(input.style.width) === <measuredSpanWidth>`. Alternatively, restore a jsdom test that mocks `getBoundingClientRect` and asserts exact equality, while keeping a browser smoke test for runtime coverage.

claude-for-yamada-ui (changes requested):
- Consistency: The `a11y` test is placed in browser mode, but per `.agents/references/review-anti-patterns/a11y-in-browser-mode.md`, `a11y(...)` is a static DOM/attribute check that runs correctly under jsdom. The audited tree here (a `<div>` containing a labeled `<input>` and a hidden `<span>`) does not require real layout, real events, observers, focus, or browser-only APIs to be audited. Running it in browser mode triples its runtime cost (chromium/firefox/webkit) for no additional signal.
- Test: The static contract assertions for the hidden span (`aria-hidden`, inline style: opacity/overflow/position/whiteSpace/zIndex) do not depend on any browser-mode criterion in `.agents/rules/test-categorization.md` (no live layout reads, no events, no observers, no focus, no browser APIs). They are attribute/style assertions that belong in jsdom. The same split is followed in nearby hooks such as `use-value` (pure logic in `*.test.tsx`, browser-dependent in `*.test.browser.tsx`).
- Test: The migration weakens the sizing contract. The previous tests asserted concrete propagation (`input.style.width` equals the mocked span width: `100px` then `200px` after rerender). The new tests only assert `width > 0` and `after > before`, which can hold even if the input's width were not actually derived from the measured span (any positive non-decreasing source would pass). This is a coverage regression for the hook's central contract: "input width tracks the measured text width".

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui: **Issue:** Static contract checks (attribute/style assertions for the hidden span) were moved entirely into a browser-mode file, even though the repository test-categorization rule says attribute-level assertions should be in jsdom (`*.test.tsx`). This diverges from the established split used in nearby hooks (for example, `use-value` keeps static logic in jsdom and browser-dependent behavior in `*.test.browser.tsx`).

**Fix:** Restore a jsdom test file for static contracts (`aria-hidden`, inline style contract, basic render contract) and keep browser-mode tests only for behavior that depends on real layout/DOM measurement.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui, line 58: **Issue:** The migration weakened the key sizing contract verification. The old tests validated concrete width propagation (`100px` then `200px` after rerender via mocked `getBoundingClientRect`), but the new tests only assert `> 0` and `after > before`. That can pass even if width is not actually derived from the measured span width.

**Fix:** Add a deterministic assertion that the input width equals the measured span width (for example, by restoring a jsdom contract test with controlled `getBoundingClientRect` values, or by explicitly asserting equality against a controlled measurement source), while keeping browser tests for runtime smoke coverage.

codex-for-yamada-ui (changes requested):
- Consistency: Static contract checks (attribute/style assertions for the hidden span) were moved entirely into a browser-mode file, even though the repository test-categorization rule says attribute-level assertions should be in jsdom (`*.test.tsx`). This diverges from the established split used in nearby hooks (for example, `use-value` keeps static logic in jsdom and browser-dependent behavior in `*.test.browser.tsx`).
- Test: The migration weakened the key sizing contract verification. The old tests validated concrete width propagation (`100px` then `200px` after rerender via mocked `getBoundingClientRect`), but the new tests only assert `> 0` and `after > before`. That can pass even if width is not actually derived from the measured span width.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/hooks/use-field-sizing/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7537, opened 2026-05-12
test(use-focus): migrate tests to browser mode

### LINKED ISSUE #7514: Migrate use-focus tests to Vitest Browser Mode
### Description

Migrate the `use-focus` hook tests to Vitest Browser Mode and drop the focus / utility mocks.

The following test file currently runs under jsdom via `#test`:

- [index.test.tsx](packages/react/src/hooks/use-focus/index.test.tsx)

The file mocks `../../utils` to stub `getFirstFocusableElement` and spies on `button.focus()`. In jsdom this is necessary because focus management does not behave like a real browser. In a real browser via Vitest Browser Mode, the actual focusable-element traversal and `HTMLElement.focus()` work correctly, so the mocks become unnecessary and the test can assert focus through `toHaveFocus()` instead.

### Proposed Solution or API

1. Replace `import { ... } from "#test"` with `import { page, render } from "#test/browser"` (add `a11y` for cases that render component DOM).
2. Rewrite assertions to use Browser Mode APIs (`page.getBy*`, `await expect.element(...).toHaveFocus()`).
3. Drop mocks that are no longer needed in a real browser:
   - `vi.mock("../../utils", async (importOriginal) => ...)` that stubs `getFirstFocusableElement` — let the real implementation run against the real DOM tree.
   - `vi.spyOn(button, "focus")` — assert focus via `toHaveFocus()` / `document.activeElement` instead.
4. Verify: `pnpm react test:browser --run src/hooks/use-focus/`.

Reference: #7503 (mock cleanup pattern), [.agents/rules/browser-testing.md](.agents/rules/browser-testing.md).

### Alternatives

_No response_

Yes

### AI used

### Additional Information

Related: #7506

hirotomoyamada:
Closes #7514

## AI used

## Description

Migrated `use-focus` tests from jsdom to Vitest Browser Mode and removed utility/focus mocking.

## Current behavior (updates)

`use-focus` tests were implemented as jsdom tests and depended on utility/focus mocks.

## New behavior

- Renamed to `index.test.browser.tsx` and switched to `#test/browser` APIs.
- Removed `../../utils` mock and direct `focus()` spy assertions.
- Replaced assertions with browser-native focus behavior checks (`toHaveFocus`, active element checks).
- Kept coverage across `useFocusOnShow` and `useFocusOnPointerDown` branches, including Safari/non-Safari paths.

No

## Additional Information

Coverage check on `src/hooks/use-focus/` (before/after):
- Before: `44/44` statements (`100.00%`)
- After: `43/44` statements (`97.73%`)
- Delta: `-2.27%` (within ±5% tolerance)

### DIFF
--- a/packages/react/src/hooks/use-focus/index.test.browser.tsx
+++ b/packages/react/src/hooks/use-focus/index.test.browser.tsx
@@ -190,145 +190,73 @@ describe("useFocusOnShow", () => {
 })
 
 describe("useFocusOnPointerDown", () => {
-  const defaultPlatform = window.navigator.platform
-  const defaultVendor = window.navigator.vendor
-
-  beforeAll(() => {
-    Object.defineProperty(window.navigator, "platform", {
-      value: "MacOS",
-      writable: true,
-    })
-    Object.defineProperty(window.navigator, "vendor", {
-      value: "Apple Computer, Inc.",
-      writable: true,
-    })
-  })
-
-  afterAll(() => {
-    Object.defineProperty(window.navigator, "platform", {
-      value: defaultPlatform,
-      writable: false,
-    })
-    Object.defineProperty(window.navigator, "vendor", {
-      value: defaultVendor,
-      writable: false,
-    })
-  })
-
-  const Component: FC<Omit<UseFocusOnMouseDownProps, "ref">> = (props) => {
+  const Component: FC<
+    Omit<UseFocusOnMouseDownProps, "ref"> & { preventNativeFocus?: boolean }
+  > = ({ preventNativeFocus = true, ...props }) => {
     const ref = useRef<HTMLDivElement>(null)
-    const buttonRef = useRef<HTMLButtonElement>(null)
+    const targetRef = useRef<HTMLDivElement>(null)
     useFocusOnPointerDown({
       ref,
-      elements: [buttonRef],
+      elements: [targetRef],
       enabled: true,
       ...props,
     })
 
     return (
       <div ref={ref} data-testid="container">
-        <button ref={buttonRef} data-testid="button">
-          Button
-        </button>
+        <div
+          ref={targetRef}
+          data-testid="target"
+          tabIndex={-1}
+          onPointerDown={
+            preventNativeFocus ? (ev) => ev.preventDefault() : undefined
+          }
+        >
+          Target
+        </div>
       </div>
     )
   }
 
-  const dispatchPointerDown = (target: Element) => {
-    target.dispatchEvent(
-      new PointerEvent("pointerdown", {
-        bubbles: true,
-        cancelable: true,
-      }),
-    )
+  const stubNavigator = (overrides: { platform?: string; vendor?: string }) => {
+    const descriptors: { [key: string]: PropertyDescriptor | undefined } = {
+      platform: Object.getOwnPropertyDescriptor(window.navigator, "platform"),
+      vendor: Object.getOwnPropertyDescriptor(window.navigator, "vendor"),
+    }
+
+    for (const [key, value] of Object.entries(overrides)) {
+      Object.defineProperty(window.navigator, key, {
+        configurable: true,
+        value,
+      })
+    }
+
+    return () => {
+      for (const [key, descriptor] of Object.entries(descriptors)) {
+        if (descriptor) {
+          Object.defineProperty(window.navigator, key, descriptor)
+        } else {
+          Reflect.deleteProperty(window.navigator, key)
+        }
+      }
+    }
   }
 
-  test("prevents default behavior and focuses on the target element", async () => {
-    await render(<Component />)
-    const button = page.getByTestId("button")
-
-    dispatchPointerDown(button.element())
-
-    await expect.element(button).toHaveFocus()
-  })
-
-  test("does not focus when disabled", async () => {
-    await render(<Component enabled={false} />)
-    const button = page.getByTestId("button")
-
-    dispatchPointerDown(button.element())
-
-    await expect.element(button).not.toHaveFocus()
-  })
-
-  test("does not move focus when target is already active", async () => {
-    await render(<Component />)
-    const button = page.getByTestId("button")
-    const buttonElement = button.element() as HTMLElement
-
-    buttonElement.focus()
-    dispatchPointerDown(buttonElement)
-
-    await expect.element(button).toHaveFocus()
-  })
-
-  test("uses ref as fallback when elements is not provided", async () => {
-    await render(<Component elements={undefined} />)
-    const button = page.getByTestId("button")
-
-    dispatchPointerDown(button.element())
-
-    await expect.element(button).toHaveFocus()
-  })
-
-  test("supports raw HTMLElement in elements and ignores null entries", async () => {
-    const { rerender } = await render(<Component />)
-    const button = page.getByTestId("button")
-    const buttonElement = button.element() as HTMLElement
-
-    rerender(<Component elements={[buttonElement]} />)
-    dispatchPointerDown(buttonElement)
-
-    await expect.element(button).toHaveFocus()
-
-    buttonElement.blur()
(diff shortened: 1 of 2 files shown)


packages/react/src/hooks/use-focus/index.test.browser.tsx
```diff
@@ -235,10 +204,6 @@ describe("useFocusOnPointerDown", () => {
     })
   })
 
-  afterEach(() => {
-    vi.restoreAllMocks()
-  })
-
   afterAll(() => {
     Object.defineProperty(window.navigator, "platform", {
       value: defaultPlatform,
```
claude-for-yamada-ui: **Issue:** The `useFocusOnPointerDown` suite forces `navigator.vendor`/`platform` to mimic Safari so that the hook's `isSafari()` branch is exercised, then runs in `*.test.browser.tsx`, which fans out to chromium/firefox/webkit. Per `.agents/rules/test-categorization.md`, hooks that branch on `isSafari()`/`isWebKit()` belong in an engine-specific file (e.g. `*.test.webkit.tsx`). The current categorization is the proximate cause of the in-scope Chromium CI failure: 3 tests time out on `toHaveFocus()` because Chromium's pointer/focus pipeline does not match the assumptions baked into the Safari path.

**Fix:** Move the `useFocusOnPointerDown` describe block (or its Safari-gated tests) into `index.test.webkit.tsx` so the suite runs only under WebKit, and keep `*.test.browser.tsx` for behavior that must hold across all engines. The "does not focus on non-safari browsers" case can stay in a chromium/firefox file or in the shared browser file.

packages/react/src/hooks/use-focus/index.test.browser.tsx
```diff
@@ -269,71 +234,70 @@ describe("useFocusOnPointerDown", () => {
...
-    act(() => fireEvent.pointerDown(button))
 
-    await waitFor(() => {
-      expect(button).not.toHaveFocus()
-    })
+    const container = page.getByTestId("container").element() as HTMLElement
+    dispatchPointerDown(container)
+
+    await expect.element(button).not.toHaveFocus()
   })
```
claude-for-yamada-ui: **Issue:** `dispatchPointerDown` directly calls `target.dispatchEvent(new PointerEvent("pointerdown", ...))`. In real browser mode, a synthetic `PointerEvent` does not produce the same focus side effects as a user gesture. The hook's `target.focus()` runs inside the synthetic handler, but the assertions `await expect.element(button).toHaveFocus()` then race against the browser's focus model and time out (this is exactly the Chromium failure pattern reported in CI: "Matcher did not succeed in time" on `dispatchPointerDown(button.element())` at lines 250, 279, 290). The previous jsdom suite hid this by virtue of the simulated event loop.

**Fix:** Drive interactions through the browser-mode user/interaction APIs exposed by `#test/browser` (e.g. `userEvent`/page interactions provided by the helpers under `packages/react/test/browser/`), or otherwise trigger pointerdown via the real input pipeline so the focus assertions resolve deterministically. Synthetic `dispatchEvent` should only remain for branches that genuinely cannot be reached via real interactions.

claude-for-yamada-ui (changes requested):
- Test: The `useFocusOnPointerDown` suite forces `navigator.vendor`/`platform` to mimic Safari so that the hook's `isSafari()` branch is exercised, then runs in `*.test.browser.tsx`, which fans out to chromium/firefox/webkit. Per `.agents/rules/test-categorization.md`, hooks that branch on `isSafari()`/`isWebKit()` belong in an engine-specific file (e.g. `*.test.webkit.tsx`). The current categorization is the proximate cause of the in-scope Chromium CI failure: 3 tests time out on `toHaveFocus()` because Chromium's pointer/focus pipeline does not match the assumptions baked into the Safari path.
- Correctness: `dispatchPointerDown` directly calls `target.dispatchEvent(new PointerEvent("pointerdown", ...))`. In real browser mode, a synthetic `PointerEvent` does not produce the same focus side effects as a user gesture. The hook's `target.focus()` runs inside the synthetic handler, but the assertions `await expect.element(button).toHaveFocus()` then race against the browser's focus model and time out (this is exactly the Chromium failure pattern reported in CI: "Matcher did not succeed in time" on `dispatchPointerDown(button.element())` at lines 250, 279, 290). The previous jsdom suite hid this by virtue of the simulated event loop.

packages/react/src/hooks/use-focus/index.test.browser.tsx
codex-for-yamada-ui: **Issue:** The Safari-specific `useFocusOnPointerDown` behavior is tested in a `*.test.browser.tsx` file, so the suite runs on Chromium/Firefox/WebKit even though this branch is explicitly `isSafari()`-gated. In this PR, Chromium CI already fails in the Safari-path tests (`toHaveFocus()` timeouts), which means the current categorization does not produce stable, engine-appropriate coverage for this branch.

**Fix:** Split the Safari-gated `useFocusOnPointerDown` suite into a WebKit-only test file (for example `index.test.webkit.tsx`) and keep cross-browser assertions in `*.test.browser.tsx` only for behavior that must hold across all engines.

packages/react/src/hooks/use-focus/index.test.browser.tsx
```diff
@@ -269,71 +234,70 @@ describe("useFocusOnPointerDown", () => {
     )
   }
 
+  const dispatchPointerDown = (target: Element) => {
+    target.dispatchEvent(
+      new PointerEvent("pointerdown", {
+        bubbles: true,
+        cancelable: true,
+      }),
+    )
```
codex-for-yamada-ui: **Issue:** The new helper dispatches synthetic DOM events directly (`target.dispatchEvent(new PointerEvent(...))`) instead of using browser interaction APIs. In this file, those direct synthetic `pointerdown` calls are exactly where the failing focus assertions are anchored (`dispatchPointerDown(button.element())`), so the tests are not reliably exercising the real interaction path in browser mode.

**Fix:** Replace direct `dispatchEvent` usage with interaction-driven input from `render()`/browser user utilities (the repo’s browser-test pattern), so pointer/focus behavior is triggered through the actual browser event pipeline and the focus assertions become stable.

codex-for-yamada-ui (changes requested):
- Test: The Safari-specific `useFocusOnPointerDown` behavior is tested in a `*.test.browser.tsx` file, so the suite runs on Chromium/Firefox/WebKit even though this branch is explicitly `isSafari()`-gated. In this PR, Chromium CI already fails in the Safari-path tests (`toHaveFocus()` timeouts), which means the current categorization does not produce stable, engine-appropriate coverage for this branch.
- Correctness: The new helper dispatches synthetic DOM events directly (`target.dispatchEvent(new PointerEvent(...))`) instead of using browser interaction APIs. In this file, those direct synthetic `pointerdown` calls are exactly where the failing focus assertions are anchored (`dispatchPointerDown(button.element())`), so the tests are not reliably exercising the real interaction path in browser mode.

packages/react/src/hooks/use-focus/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

packages/react/src/hooks/use-focus/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

packages/react/src/hooks/use-focus/index.test.browser.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7549, opened 2026-05-13
feat(mergeProps): default `mergeAria` to true

hirotomoyamada:
## Description

Follows up on #7546. Changes the default of `mergeAria` from `false` to `true`, so that `mergeProps` concatenates space-separated WAI-ARIA list attributes (`aria-controls`, `aria-describedby`, `aria-flowto`, `aria-labelledby`, `aria-owns`) by default. Pass `mergeAria: false` to opt back into the previous last-wins behavior.

## AI used

No

### DIFF
--- a/.changeset/bright-aria-merge.md
+++ b/.changeset/bright-aria-merge.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": minor
+---
+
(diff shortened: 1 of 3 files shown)


hirotomoyamada:
Replaced by force-pushing the equivalent change directly onto `v2.3`.

PR #7565, opened 2026-05-14
refactor(color-selector): use mergeProps with mergeAria for aria attribute merging

### LINKED ISSUE #7564: refactor(color-selector): use mergeProps with mergeAria for aria attribute merging
### Description

Refactor `useColorSelector` to use `mergeProps` with the `mergeAria` option (added in #7546) instead of manual `cx()` merging for aria attributes.

`useColorSelector` currently destructures aria attributes from props and merges them manually with `cx()`. Now that `mergeProps` natively supports aria attribute merging via the `mergeAria` option (default `true`), this manual handling is redundant.

Yes

### AI used

### Additional Information

Follow-up to #7546.

hirotomoyamada:
Closes #7564

## AI used

## Description

Refactored `useColorSelector` to use `mergeProps` with `mergeAria` for aria attribute merging.

## Current behavior (updates)

## New behavior

No

## Additional Information

Follow-up to #7546.


packages/react/src/components/color-selector/use-color-selector.ts
```diff
@@ -284,13 +283,16 @@ export const useColorSelector = ({
...
   const getColorSwatchGroupProps: PropGetter = useCallback(
-    ({ "aria-labelledby": ariaLabelledby, ...props } = {}) => ({
-      "aria-labelledby": cx(ariaLabelledby, swatchGroupLabelId),
-      "data-disabled": dataAttr(disabled),
-      "data-readonly": dataAttr(readOnly),
-      role: "listbox",
-      ...props,
-    }),
+    (props = {}) =>
+      mergeProps(
```
codex-for-yamada-ui, line 287: **Issue:** This refactor reverses the order of merged `aria-labelledby` IDs. Previously user-provided labels were placed before the internal swatch group label via `cx(ariaLabelledby, swatchGroupLabelId)`, but `mergeProps` now produces `swatchGroupLabelId` first and user labels second. The order is observable because accessible names are computed in `aria-labelledby` token order, so a pure refactor can change what screen readers announce.

**Fix:** Preserve the previous merge order for `aria-labelledby` while still allowing the other props to merge/override as before, or keep the original explicit `cx(ariaLabelledby, swatchGroupLabelId)` behavior.

codex-for-yamada-ui (changes requested):
- This refactor reverses the order of merged `aria-labelledby` IDs. Previously user-provided labels were placed before the internal swatch group label via `cx(ariaLabelledby, swatchGroupLabelId)`, but `mergeProps` now produces `swatchGroupLabelId` first and user labels second. The order is observable because accessible names are computed in `aria-labelledby` token order, so a pure refactor can change what screen readers announce.

composer-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

packages/react/src/components/color-selector/use-color-selector.ts
hirotomoyamada: I don't think there's any particular accessibility issue even if the order of the `id` attributes is reversed and the order in which they are read by screen readers changes.

codex-for-yamada-ui: I still think this should be preserved. `aria-labelledby` is an ordered IDREF list, and the accessible name is computed by concatenating referenced labels in that token order. So this refactor changes the exposed accessible name from the previous user-label-first order to the internal-label-first order for consumers that pass their own `aria-labelledby`.

Even if both orders are technically readable, this is still an observable behavior change in a refactor and can change what assistive technologies announce. Please keep the previous `cx(ariaLabelledby, swatchGroupLabelId)` ordering for `aria-labelledby` while leaving the rest of the prop merge behavior unchanged.

hirotomoyamada: We plan to clearly state in the release notes for the minor update that the order in which screen readers read the text may change. Therefore, this issue does not pose a particular problem for owners.

codex-for-yamada-ui (approved):

PR #7567, opened 2026-05-14
refactor(dropzone): use mergeProps with mergeAria for aria attribute merging

### LINKED ISSUE #7566: refactor(dropzone): use mergeProps with mergeAria for aria attribute merging
### Description

Refactor `useDropzone` to use `mergeProps` with the `mergeAria` option (added in #7546) instead of manual `cx()` merging for aria attributes.

`useDropzone` currently destructures aria attributes from props and merges them manually with `cx()`. Now that `mergeProps` natively supports aria attribute merging via the `mergeAria` option (default `true`), this manual handling is redundant.

Yes

### AI used

### Additional Information

Follow-up to #7546.

hirotomoyamada:
Closes #7566

## AI used

## Description

Refactored `useDropzone` to use `mergeProps` with `mergeAria` for aria attribute merging.

## Current behavior (updates)

## New behavior

No

## Additional Information

Follow-up to #7546.


packages/react/src/components/dropzone/use-dropzone.ts
```diff
@@ -278,17 +271,22 @@ export const useDropzone = (props: UseDropzoneProps = {}) => {
...
+      getOriginalInputProps(
+        mergeProps(
+          {
+            id,
+            name,
+            disabled,
+            readOnly,
+            ...ariaProps,
+            ...dataProps,
+            ...props,
```
composer-for-yamada-ui: **Issue:** The `...props` spread inside the first argument to `mergeProps` is redundant because `props` is also passed as the second argument. This causes non-aria properties (like `className`) to be duplicated. For example, `className` would be set from `...props` in the first object, then `cx(props.className, props.className)` would be called when merging with the second `props` argument, resulting in the same className appearing twice.

**Fix:** Remove `...props` from the first object passed to `mergeProps`. The first object should only contain the internal defaults (`id`, `name`, `disabled`, `readOnly`, `...ariaProps`, `...dataProps`, `"aria-labelledby": labelledbyId`), and `props` should only be passed as the second argument to `mergeProps` for proper merging.

claude-for-yamada-ui: **Issue:** `...props` is spread into the first `mergeProps` argument (line 284) AND also passed as the second argument (line 287). When the user passes event handlers (e.g., `onClick`) to `getInputProps`, `mergeProps` with `mergeEvent: true` will call `handlerAll(result[key], props[key])` where both are the same user function, causing the handler to fire twice. This is inconsistent with the existing `getRootProps` pattern (lines 237-251) which does NOT spread user props into the first `mergeProps` argument.

**Fix:** Remove `...props` from the first `mergeProps` argument on line 284, matching the `getRootProps` pattern. User props should only appear as the second `mergeProps` argument where they are properly merged/overridden.

codex-for-yamada-ui: **Issue:** `props` is spread into the first `mergeProps` argument and then passed again as the second argument. Any mergeable input prop, such as `onChange`, `onClick`, `className`, `style`, or ARIA token list props, is therefore merged with itself; event handlers can run twice and class/style/ARIA values can be duplicated or reordered compared with the previous behavior.

**Fix:** Pass the caller props only once. For example, remove `...props` from the first object and, if the old `aria-labelledby` order must be preserved, merge the internal labelled-by id as a separate trailing prop bag: `mergeProps({ id, name, disabled, readOnly, ...ariaProps, ...dataProps }, props, { "aria-labelledby": labelledbyId })()`.

composer-for-yamada-ui (changes requested):
- The `...props` spread inside the first argument to `mergeProps` is redundant because `props` is also passed as the second argument. This causes non-aria properties (like `className`) to be duplicated. For example, `className` would be set from `...props` in the first object, then `cx(props.className, props.className)` would be called when merging with the second `props` argument, resulting in the same className appearing twice.

claude-for-yamada-ui (changes requested):
- `...props` is spread into the first `mergeProps` argument (line 284) AND also passed as the second argument (line 287). When the user passes event handlers (e.g., `onClick`) to `getInputProps`, `mergeProps` with `mergeEvent: true` will call `handlerAll(result[key], props[key])` where both are the same user function, causing the handler to fire twice. This is inconsistent with the existing `getRootProps` pattern (lines 237-251) which does NOT spread user props into the first `mergeProps` argument.

codex-for-yamada-ui (changes requested):
- `props` is spread into the first `mergeProps` argument and then passed again as the second argument. Any mergeable input prop, such as `onChange`, `onClick`, `className`, `style`, or ARIA token list props, is therefore merged with itself; event handlers can run twice and class/style/ARIA values can be duplicated or reordered compared with the previous behavior.

packages/react/src/components/dropzone/use-dropzone.ts
codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

codex-for-yamada-ui: Fixed.

composer-for-yamada-ui (dismissed):

codex-for-yamada-ui (dismissed):

claude-for-yamada-ui (dismissed):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

PR #7577, opened 2026-05-14
refactor(pin-input): use mergeProps with mergeAria for aria attribute merging

### LINKED ISSUE #7576: refactor(pin-input): use mergeProps with mergeAria for aria attribute merging
### Description

Refactor `usePinInput` to use `mergeProps` with the `mergeAria` option (added in #7546) instead of manual `cx()` merging for aria attributes.

`usePinInput` currently destructures aria attributes from props and merges them manually with `cx()`. Now that `mergeProps` natively supports aria attribute merging via the `mergeAria` option (default `true`), this manual handling is redundant.

Yes

### AI used

### Additional Information

Follow-up to #7546.

hirotomoyamada:
Closes #7576

## AI used

## Description

Refactored `usePinInput` to use `mergeProps` with `mergeAria` for aria attribute merging.

## Current behavior (updates)

## New behavior

No

## Additional Information

Follow-up to #7546.


packages/react/src/components/pin-input/use-pin-input.ts
```diff
@@ -297,27 +297,34 @@ export const usePinInput = (props: UsePinInputProps = {}) => {
...
+            placeholder:
+              focusedIndex === index && !readOnly && !props.readOnly
+                ? ""
+                : placeholder,
+            readOnly,
+            value: values[index] || "",
+            ...filterUndefined(props),
+            id: `${id}${index ? `-${index}` : ""}`,
+          },
+          props,
```
codex-for-yamada-ui: **Issue:** `props` is applied twice: it is spread into the first object and then passed again as the second `mergeProps` argument. When callers pass an array `css` prop, the first merge stores that same array in `result.css`, and the second merge hits the `mergeCSS` array branch and pushes the array into itself. This creates a circular CSS value and matches the CI `RangeError: Maximum call stack size exceeded` in `createCSS`. The same duplication also drops the internal `labelId` when `aria-labelledby` is supplied and can duplicate user event handlers.

**Fix:** Do not spread `props` into the first argument. Keep only the hook-owned base props there, then pass `filterUndefined(props)` once as the next `mergeProps` argument before the internal handlers so aria, css, className, and events are merged exactly once.

claude-for-yamada-ui: **Issue:** `getInputProps` passes `props` as both a spread in the first argument (`...filterUndefined(props)`) and as the second argument (`props` directly) to `mergeProps`. This causes `mergeProps` to apply its special merge logic for `css` (creating `[obj, obj]`), `style` (deep merge), and other keys on the same values. The CSS array or deep-merged style interacts incorrectly with yamada-ui's CSS processing, causing `RangeError: Maximum call stack size exceeded` in `createCSS` → `transformInterpolation`. This breaks all pin-input tests across JSDOM, Firefox, Chromium, and Webkit runners.

**Fix:** Remove the second argument (`props`) from the `mergeProps` call since its values are already included in the first argument via `...filterUndefined(props)`. If event/aria merging from caller props is needed, handle those explicitly rather than passing the full `props` object as a separate merge source.

codex-for-yamada-ui (changes requested):
- `props` is applied twice: it is spread into the first object and then passed again as the second `mergeProps` argument. When callers pass an array `css` prop, the first merge stores that same array in `result.css`, and the second merge hits the `mergeCSS` array branch and pushes the array into itself. This creates a circular CSS value and matches the CI `RangeError: Maximum call stack size exceeded` in `createCSS`. The same duplication also drops the internal `labelId` when `aria-labelledby` is supplied and can duplicate user event handlers.

claude-for-yamada-ui (changes requested):
- `getInputProps` passes `props` as both a spread in the first argument (`...filterUndefined(props)`) and as the second argument (`props` directly) to `mergeProps`. This causes `mergeProps` to apply its special merge logic for `css` (creating `[obj, obj]`), `style` (deep merge), and other keys on the same values. The CSS array or deep-merged style interacts incorrectly with yamada-ui's CSS processing, causing `RangeError: Maximum call stack size exceeded` in `createCSS` → `transformInterpolation`. This breaks all pin-input tests across JSDOM, Firefox, Chromium, and Webkit runners.

composer-for-yamada-ui (dismissed):

packages/react/src/components/pin-input/use-pin-input.ts
codex-for-yamada-ui: Fixed by removing the duplicate props spread from the base object and passing filtered caller props only once before the internal props.

codex-for-yamada-ui: Fixed by ensuring getInputProps no longer merges the same props object twice, avoiding duplicate css/style/event merging.

composer-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

claude-for-yamada-ui (approved):

PR #7589, opened 2026-05-14
refactor(slider): use mergeProps with mergeAria for aria attribute merging

### LINKED ISSUE #7588: refactor(slider): use mergeProps with mergeAria for aria attribute merging
### Description

Refactor `useSlider` to use `mergeProps` with the `mergeAria` option (added in #7546) instead of manual `cx()` merging for aria attributes.

`useSlider` currently destructures aria attributes from props and merges them manually with `cx()`. Now that `mergeProps` natively supports aria attribute merging via the `mergeAria` option (default `true`), this manual handling is redundant.

Yes

### AI used

### Additional Information

Follow-up to #7546.

hirotomoyamada:
Closes #7588

## AI used

## Description

Refactored `useSlider` to use `mergeProps` with `mergeAria` for aria attribute merging.

## Current behavior (updates)

## New behavior

No

## Additional Information

Follow-up to #7546.


packages/react/src/components/slider/use-slider.ts
```diff
@@ -394,38 +392,39 @@ export const useSlider = <Y extends [number, number] | number = number>(
...
+            }
+          : {
+              "aria-valuenow": value,
+              "aria-valuetext":
+                ariaValueText ??
+                getAriaValueText?.(value, index) ??
+                value.toString(),
+              "data-end": dataAttr(index === 0),
+            },
+        rest,
```
codex-for-yamada-ui, line 425: **Issue:** Placing `rest` after the computed value props lets `thumbProps` or `Slider.Thumb` props override `aria-valuenow`, `aria-valuetext`, `data-start`, and `data-end`. Before this refactor those props were assigned after `rest`, so the thumb always exposed the current slider value and correct range markers. A stale `aria-valuenow` override would make the visual slider and screen reader value diverge after value changes.

**Fix:** Keep user props before the computed thumb state, then merge the internal keyboard handler last. For example, merge base/aria props, `rest`, then the range/single computed props, and merge `aria-labelledby` in the same final position needed to preserve the previous label order.

codex-for-yamada-ui (changes requested):
- Placing `rest` after the computed value props lets `thumbProps` or `Slider.Thumb` props override `aria-valuenow`, `aria-valuetext`, `data-start`, and `data-end`. Before this refactor those props were assigned after `rest`, so the thumb always exposed the current slider value and correct range markers. A stale `aria-valuenow` override would make the visual slider and screen reader value diverge after value changes.

claude-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

packages/react/src/components/slider/use-slider.ts
hirotomoyamada: Just like with other components, it should be possible to override the default value using `rest`. The reason is that if `rest` is moved to the front, users will no longer be able to override the default. This is not ideal.

codex-for-yamada-ui (approved):

PR #7600, opened 2026-05-15
test(browser): stabilize remaining flaky date-picker/select/tooltip/popover tests

hirotomoyamada:
Closes #

## AI used

## Description

Stabilizes the four browser tests that continued to flake across recent PRs after #7550. Root causes were distinct in each case; this PR addresses each one rather than retrying with longer timeouts.

## Current behavior (updates)

- `<DatePicker /> > handles Enter key on single date input` (Firefox): intermediate `toHaveValue("2024-01-15")` / `toHaveFocus()` assertions race the parse-on-fill, sometimes reading the formatted value before the assertion runs.
- `<DatePicker /> > handles input change for range with existing start value` / `handles parseDate` (Firefox): per-keystroke `user.type` feeds partially valid strings into the parse pipeline; for `parseDate` this also surfaces an unhandled \`RangeError: date value is not finite in DateTimeFormat.format()\`.
- `<Select /> > selects and deselects values in multiple mode` (Firefox): Playwright's actionability check trips on \"element not stable\" because the option list re-renders between multi-selections.
- `<Tooltip /> > handles openDelay and closeDelay` / `does not open on touch pointer events` (Firefox): residual pointer position from prior tests either suppresses the new \`user.hover\` (no pointermove fires) or races the synthetic touch dispatch.
- `<Popover /> > should return focus to trigger after escape key` (WebKit): focus restore after popover close doesn't complete within the default 1s \`expect.poll\` window on WebKit.

## New behavior

- date-picker: drop the racy intermediate assertions on the Enter-key path; switch range/parseDate tests from `user.type` to `user.fill` so the parse pipeline only sees the final string.
- select: pass `force: true` on the multi-select option clicks (matches the existing convention in the rest of this file).
- tooltip: render a fixed-position `pointer-reset` anchor and park the cursor on it before each scenario (mirrors the [use-hover](../blob/main/packages/react/src/hooks/use-hover/use-hover.test.browser.tsx) pattern).
- popover: replace the focus-return `expect.poll` with `vi.waitFor` at \`{ timeout: 3000 }\` so WebKit's focus restore has time to settle.

No. Test-only changes.

## Additional Information

Local verification on the worktree (post-rebase to \`main\`):

| File | Firefox | Chromium | WebKit |
|---|---|---|---|
| date-picker.test.browser.tsx | 5/5 ✅ | 1/1 ✅ | 1/1 ✅ |
| select.test.browser.tsx | 5/5 ✅ | 1/1 ✅ | 1/1 ✅ |
| tooltip.test.browser.tsx | 5/5 ✅ | 1/1 ✅ | 1/1 ✅ |
| popover.test.browser.tsx | 1/1 ✅ | 1/1 ✅ | 5/5 ✅ |

### DIFF
--- a/packages/react/src/components/date-picker/date-picker.test.browser.tsx
+++ b/packages/react/src/components/date-picker/date-picker.test.browser.tsx
@@ -137,6 +137,7 @@ describe("<DatePicker />", () => {
     const input = page.getByRole("textbox").first()
     await user.click(input)
     await user.fill(input, "2024-01-15")
+    await expect.element(input).toHaveValue("2024-01-15")
     await user.keyboard("{Enter}")
 
     await vi.waitFor(async () => {
--- a/packages/react/src/components/date-picker/date-picker.test.browser.tsx
+++ b/packages/react/src/components/date-picker/date-picker.test.browser.tsx
@@ -137,11 +137,11 @@ describe("<DatePicker />", () => {
     const input = page.getByRole("textbox").first()
     await user.click(input)
     await user.fill(input, "2024-01-15")
-    await expect.element(input).toHaveValue("2024-01-15")
-    await expect.element(input).toHaveFocus()
     await user.keyboard("{Enter}")
 
-    await expect.element(input).toHaveValue("January 15, 2024")
+    await vi.waitFor(async () => {
+      await expect.element(input).toHaveValue("January 15, 2024")
+    })
   })
 
   test("handles Enter key on range date start input", async () => {
@@ -865,8 +865,7 @@ describe("<DatePicker />", () => {
 
     const inputs = page.getByRole("textbox")
     await user.click(inputs.nth(1))
-    await user.clear(inputs.nth(1))
-    await user.type(inputs.nth(1), "2024-01-20")
+    await user.fill(inputs.nth(1), "2024-01-20")
 
     await expect.element(inputs.nth(1)).toHaveValue("2024-01-20")
   })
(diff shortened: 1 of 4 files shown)


claude-for-yamada-ui (dismissed):

composer-for-yamada-ui (dismissed):

packages/react/src/components/date-picker/date-picker.test.browser.tsx
```diff
@@ -137,11 +137,11 @@ describe("<DatePicker />", () => {
     const input = page.getByRole("textbox").first()
     await user.click(input)
     await user.fill(input, "2024-01-15")
-    await expect.element(input).toHaveValue("2024-01-15")
-    await expect.element(input).toHaveFocus()
     await user.keyboard("{Enter}")
 
-    await expect.element(input).toHaveValue("January 15, 2024")
+    await vi.waitFor(async () => {
+      await expect.element(input).toHaveValue("January 15, 2024")
```
codex-for-yamada-ui, line 144: **Issue:** This test no longer proves that formatting is triggered by the Enter key. After removing the pre-Enter value check, it can pass even if formatting happens on input change, which weakens the Enter-key contract coverage of this case.

**Fix:** Re-introduce a pre-Enter assertion (for example `toHaveValue("2024-01-15")`) before `user.keyboard("{Enter}")`, then keep the post-Enter formatted-value assertion.

codex-for-yamada-ui (changes requested):
- Testing: This test no longer proves that formatting is triggered by the Enter key. After removing the pre-Enter value check, it can pass even if formatting happens on input change, which weakens the Enter-key contract coverage of this case.

packages/react/src/components/date-picker/date-picker.test.browser.tsx
codex-for-yamada-ui: Fixed.

claude-for-yamada-ui (approved):

codex-for-yamada-ui (approved):

composer-for-yamada-ui (approved):

Issue #6707, opened 2026-04-13
Create ESLint custom plugin for Yamada UI


chida09:
## Description

Upgrade ESLint from 9 to 10 and develop a custom ESLint plugin for Yamada UI. The plugin will be developed under `packages/lint` as `@yamada-ui/lint` and published as an npm package so that Yamada UI users can integrate it into their own projects. It follows the same package structure and release flow as existing published packages (e.g., `@yamada-ui/cli`).

## Problem Statement/Justification

Currently, there is no way to detect or auto-fix the following issues via lint in projects using Yamada UI:

- Hard-coded values are used in style props instead of design tokens (e.g., `14px`)
- Shorthand prop usage is not consistent across projects
- Generic components are used when more specific ones exist (e.g., `Flex` with `direction="column"` instead of `VStack`)

Reference: https://github.com/yukukotani/eslint-plugin-chakra-ui

## Proposed Solution or API

Individual tasks are tracked as sub-issues.

## Alternatives

N/A

### Are you willing to participate in implementing this feature and create a pull request include the implementation?

Yes

### AI used

- [ ] I did not use AI to create this issue.
- [x] (If there is no check above) I checked the generated content before submitting.

## Additional Information

N/A

PR #7607, opened 2026-05-18
feat(lint): add @yamada-ui/eslint-plugin package with props-shorthand rule

### LINKED ISSUE #6710: Configurable shorthand prop restriction per project
## Description

Create an ESLint rule that allows projects to enable or disable shorthand prop restrictions. This should be configurable per project so that each team can enforce their preferred style.

Parent: #6707

### AI used

chida09:
Closes #6710

## AI used

## Description

Adds a new publishable workspace package `@yamada-ui/eslint-plugin` and its first rule `props-shorthand`, which enforces a consistent choice of shorthand or longhand style props on Yamada UI components with auto-fix.

## Current behavior (updates)

No lint rule exists to keep style prop usage consistent or to detect duplicates such as `<Box m={1} margin={2} />`.

## New behavior

- `preferred: "shorthand"` (default) auto-fixes longhand to shorthand; `preferred: "longhand"` does the reverse.
- Duplicate style props (shorthand + longhand, or two shorthands mapping to the same longhand) are reported as a single error per tag and only reference names actually written in JSX.
- Tracks named imports, the `styled` factory, default imports, and namespace imports from `@yamada-ui/react`. Other elements and spread attributes are skipped.
- Adds the `pnpm linter` workspace shortcut.

No.


bmthd (commented):
I think the current rule name may be slightly misleading.

`props-shorthand` sounds like a rule that specifically enforces shorthand props, but the actual behavior is “choose either shorthand or longhand and enforce consistency” via the `preferred` option.

For example, this configuration reads a little oddly:

```json id="9fr3u2"
"@yamada-ui/eslint-plugin/props-shorthand": [
  "error",
  { "preferred": "longhand" }
]
```

Because of that, would a more neutral name such as one of the following better reflect the rule responsibility?

* `style-prop-consistency`
* `consistent-style-props`
* `style-prop-notation`

The current implementation itself makes sense to me, especially since both styles are treated as first-class options.

I'd also be interested to hear what others think about the naming here.

packages/eslint-plugin/src/rules/props-shorthand/index.ts
bmthd, line 1: I wonder if this rule needs to depend on `@typescript-eslint/utils`.

As far as I can tell, this rule does not require type information. It only inspects imports, JSX opening elements, and JSX attributes, then reports/fixes prop names. If so, implementing it against the plain ESLint rule API may be enough.

My concern is compatibility with non-ESLint runtimes such as Oxlint. Oxlint has JS plugin support, but it is still alpha, and depending on `@typescript-eslint/*` may make the plugin more tightly coupled to the TypeScript ESLint ecosystem than necessary.

Would it be better to avoid `@typescript-eslint/utils` unless we need type-aware behavior, and keep this rule closer to the standard ESLint plugin API?

packages/eslint-plugin/src/rules/props-shorthand/index.test.ts
bmthd, line 1: Overall, I get the impression that the rule creation utilities and test infrastructure were built specifically for this single rule.

Since the milestone seems to assume that more rules will be added in the future, I think it may be worth thinking a bit more about how easy it will be to author and maintain additional rules going forward.

For example, it might make sense to:

* first introduce the shared rule/test infrastructure in a separate preparatory PR,
* expand this PR so the abstractions are clearly designed for future rules as well,
* or follow up with a dedicated refactor PR afterward.

Right now, some of the abstractions feel slightly too coupled to the current `props-shorthand` implementation, so I wanted to raise the concern before the plugin surface grows further.

chida09: Thanks for flagging this — you're right that the plugin is structured for a single rule today, and the parent issue #6707 plans for more rules (design tokens, component suggestions, etc.), so this concern is well-aimed.

That said, I'd like to avoid premature abstraction with only one concrete rule to base it on.

Keep this PR focused on landing `props-shorthand` with the standard one-folder-per-rule layout (matches typescript-eslint, eslint-plugin-react).

I also confirmed this offline with you directly, so no concrete changes will be made in this PR.

packages/eslint-plugin/src/rules/props-shorthand/index.ts
chida09: Fixed.
https://github.com/yamada-ui/yamada-ui/pull/7607/changes/50e36e3cd58ca1639922755a8f5397d130b35cb5

bmthd (commented):
imo-badge

I wonder if we should avoid depending on the TypeScript Compiler API here, especially with Oxlint support in mind.

The current implementation determines whether a component belongs to Yamada UI by resolving and analyzing its import source. While this improves accuracy, it also significantly increases the complexity of the rule and introduces a dependency on the TypeScript Compiler API.

Would it make sense to start with a simpler approach that only checks component names without resolving imports? It may introduce some false positives, but the tradeoff seems reasonable for a lint rule like this.

This would allow us to remove the TypeScript Compiler API dependency and keep the rule compatible with Oxlint. Given that Oxlint is quickly becoming a de facto choice in the ecosystem, I'd prefer not to exclude it unless we have a strong reason to do so.

For more advanced type-aware analysis, it might be better to revisit this once Corsa matures and stabilizes. For now, I think prioritizing a simpler implementation with broader tooling compatibility would be a better tradeoff.

PR #7661, opened 2026-05-25
ci(security): add zizmor workflow

### LINKED ISSUE #7660: Add zizmor security workflow
## Description

Add a GitHub Actions workflow that runs zizmor against workflow, composite action, and Dependabot configuration changes.

GitHub Actions definitions are part of the project security boundary. The repository currently has multiple workflows, composite actions, and Dependabot configuration, but no dedicated zizmor check to surface insecure workflow patterns during review.

## Proposed Solution or API

Create `.github/workflows/github-actions.yml` with a `zizmor` job that runs `zizmorcore/zizmor-action` with Advanced Security enabled for code scanning results.

## Alternatives

Rely only on manual workflow review or run zizmor locally. That is less consistent than a dedicated PR and base-branch workflow.

Yes

## AI used

## Additional Information

Reference: https://docs.zizmor.sh/integrations/#github-actions

taroj1205:
- Closes #7660

## AI used

## Description

Adds a dedicated GitHub Actions workflow that runs zizmor for workflow, composite action, and Dependabot configuration changes.

## Current behavior (updates)

GitHub Actions security analysis is not run as a dedicated repository workflow.

## New behavior

The new `GitHub Actions` workflow runs the lowercase `zizmor` job with `zizmorcore/zizmor-action`, Advanced Security enabled, minimal workflow permissions, and checkout credential persistence disabled.

No

## Additional Information

Verified with:

- `actionlint .github/workflows/github-actions.yml`
- `zizmor .github/workflows/github-actions.yml`
- `pnpm exec oxfmt --check .github/workflows/github-actions.yml`

The broader pre-PR helper was attempted, but repository-wide formatting is currently blocked by unrelated untracked local files outside this PR diff.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -1,4 +1,4 @@
-name: Security
+name: GitHub Actions
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


hirotomoyamada (changes requested):
Please incorporate this into `quality.yml`. It doesn't make sense to put it in a separate file. Also, since we're checking out the repository, I thought we needed to set `contents: read` in `permissions`—was I wrong?

taroj1205:
> Please incorporate this into `quality.yml`. It doesn't make sense to put it in a separate file. Also, since we're checking out the repository, I thought we needed to set `contents: read` in `permissions`—was I wrong?

@hirotomoyamada 
I separated it intentionally so the workflow can be gated with `on.<event>.paths`. If we put this inside `quality.yml`, we cannot add workflow-level `paths` without also skipping the normal Quality checks on non-GitHub-Actions PRs, so we would need a `changed-files` step/job-level filtering just for zizmor. A separate workflow keeps that simpler and only runs when Actions-related files change.

On `contents: read`, yes, I think you’re right that we should make it explicit. `checkout` happened to work as-is, but the job should declare the read permission it needs alongside `security-events: write`. I’ll update it to:

```yaml
permissions:
  contents: read
  security-events: write
```

Then make that permission change in the PR.

PR #7691, opened 2026-06-13
fix(icon): avoid fake default size token

### LINKED ISSUE #7690: `Icon` generates a fake `fontSizes.1em` CSS variable for default size

@yamada-ui/react

### Package Version

2.2.2

### Description

`Icon` generates a noisy and misleading CSS custom property for its default size.

For the default `Icon` styles, DevTools shows output like:

```css
width: var(--size);
height: var(--size);
vertical-align: middle;

--size: var(--fontSizes1em, 1em);
line-height: 1em;
```

The rendered size still falls back to `1em`, but `--fontSizes1em` is not a real theme token. This appears to be caused by `Icon` remapping `fontSize` to `--size` while the generic injected-vars path treats the literal `1em` as a `fontSizes` token candidate.

### Link to Reproduction

_No public reproduction yet. Observed in DevTools while inspecting `Icon` styles._

### Steps to Reproduce

1. Render an `Icon` without an explicit size/font size.
2. Inspect the generated Emotion CSS in DevTools.
3. See `--size: var(--fontSizes1em, 1em)` used by `width: var(--size)` and `height: var(--size)`.

### Browser

Chrome

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

Yes

No

### Additional Information

Investigation notes:

- `Icon` maps `fontSize` into `--size` and returns `boxSize: "{size}"` in `packages/react/src/components/icon/icon.tsx`.
- The default component style sets `fontSize: "1em"` in `packages/react/src/components/icon/icon.style.ts`.
- `fontSize` is configured with token family `fontSizes` in `packages/react/src/core/css/styles.ts`.
- `injectVars` wraps targeted tokenized props as `{fontSizes.1em, 1em}` in `packages/react/src/core/system/var.ts`.
- The CSS interpolation path converts that fallback form into `var(--fontSizes1em, 1em)` in `packages/react/src/core/css/css.ts`.

Likely fix direction: keep real `fontSizes` tokens resolving through the token system, but avoid tokenizing literal CSS length values such as `1em` when assigning the injected `--size` variable.

### AI used

hirotomoyamada:
Closes #7690

## AI used

## Description

Moves the default Icon size from fontSize to the --size variable so the literal `1em` is not treated as a fontSizes token candidate.

## Current behavior (updates)

Default Icon styles can emit `--size: var(--fontSizes1em, 1em)` even though `fontSizes.1em` is not a real token.

## New behavior

Default Icon styles set `--size: 1em` directly while preserving `boxSize: "{size}"` usage for width and height.

No

## Additional Information

- Verified with `pnpm react test:jsdom --run src/components/icon/`.

### DIFF
--- a/packages/react/src/components/icon/icon.style.ts
+++ b/packages/react/src/components/icon/icon.style.ts
@@ -2,10 +2,10 @@ import { defineComponentStyle } from "../../core"
 
 export const iconStyle = defineComponentStyle({
   base: {
+    "--size": "1em",
     color: "currentColor",
     display: "inline-block",
     flexShrink: 0,
-    fontSize: "1em",
     lineHeight: "1em",
   },
 })
--- a/.changeset/silver-icons-smile.md
+++ b/.changeset/silver-icons-smile.md
@@ -1,0 +1,5 @@
+---
+"@yamada-ui/react": patch
+---
+
+Avoid generating a fake font size token fallback for the default Icon size.


packages/react/src/components/icon/icon.style.ts
```diff
@@ -2,10 +2,10 @@ import { defineComponentStyle } from "../../core"
 
 export const iconStyle = defineComponentStyle({
   base: {
+    "--size": "1em",
     color: "currentColor",
     display: "inline-block",
     flexShrink: 0,
-    fontSize: "1em",
     lineHeight: "1em",
```
taroj1205, line 9: do you know what caused passing `1em` to `fontSize` in here generating a fake font-size token?

odendayoko (approved):

packages/react/src/components/icon/icon.style.ts
hirotomoyamada: I see. Isn't the issue that the generated CSS variables are just noise?

taroj1205 (approved):

t4isei (approved):

PR #7693, opened 2026-06-14
Enhance CSS compatibility by adding vendor-specific properties

### LINKED ISSUE #7692: Vendor-prefixed style props like `WebkitMaskSize` are forwarded to the DOM

@yamada-ui/react

### Package Version

2.2.2

### Description

Passing a vendor-prefixed CSS prop such as `WebkitMaskSize` directly to a Yamada UI component is accepted by the `CSSProps` type surface, but it is forwarded to the rendered DOM element instead of being consumed as a style prop.

For example:

```tsx
import { Button, MenuRoot, ZodiacLeoIcon } from "@yamada-ui/react"

function App() {
  return <Button WebkitMaskSize="cover">Button</Button>
}

export default App
```

Observed behavior: the rendered button receives a `WebkitMaskSize="cover"` DOM attribute.

Expected behavior: `WebkitMaskSize` should be treated as a CSS prop and serialized into generated CSS as `-webkit-mask-size: cover`, or it should not be accepted as a direct style prop if direct vendor props are intentionally unsupported.

The nested `css` prop handles the same value as CSS:

```tsx
import { Button, MenuRoot, ZodiacLeoIcon } from "@yamada-ui/react"

function App() {
  return (
    <Button
      css={{
        WebkitMaskSize: "cover",
      }}
    >
      Button
    </Button>
  )
}

export default App
```

Observed behavior for the nested `css` version: Emotion serializes the value into CSS as `-webkit-mask-size`.

This looks like a mismatch between the direct style-prop filtering table and the broader `CSSProps` type surface. Source inspection shows:

- `packages/react/src/core/css/index.types.ts` includes `VendorProps` from `CSS.VendorPropertiesFallback` in `CSSProps`.
- `packages/react/src/core/components/props.ts` filters direct DOM props with `cssProps`, which is built only from `conditionProperties` and generated `styleProperties`.
- `packages/react/src/core/css/styles.ts` includes unprefixed `maskSize`, but does not include `WebkitMaskSize`, `webkitMaskSize`, or `msWritingMode` in the generated style-prop list.

### Link to Reproduction

_No external reproduction yet. The code snippets above are the minimal reproduction._

### Steps to Reproduce

1. Render a Yamada UI app with the direct-prop example above.
2. Inspect the rendered `button` element in the DOM.
3. See that `WebkitMaskSize="cover"` is forwarded as a DOM attribute.
4. Replace it with the nested `css={{ WebkitMaskSize: "cover" }}` version.
5. See that the style is serialized into CSS as `-webkit-mask-size` instead of being forwarded as a DOM attribute.

### Browser

Chrome

### Operating System

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iPadOS
- [ ] iOS
- [ ] Android

No

No

### Additional Information

Related source areas inspected:

```ts
// packages/react/src/core/components/props.ts
export const cssProps = new Set<string>([
  ...conditionProperties,
  ...styleProperties,
])

export function createShouldForwardProp(forwardProps: string[] = []) {
  return function (prop: string): boolean {
    if (forwardProps.includes(prop)) return true
    if (prop.startsWith("--")) return false
    return !cssProps.has(prop)
  }
}
```

```ts
// packages/react/src/core/css/index.types.ts
type VendorProps = {
  [Y in keyof CSS.VendorPropertiesFallback]?: StyleValueWithCondition<
    AnyString | CSS.VendorPropertiesFallback[Y]
  >
}

export interface CSSProps
  extends StyleProps, ConditionProps, VendorProps, CSSVariableProps {
  css?: CSSObject | CSSObject[]
}
```

Focused duplicate searches for `WebkitMaskSize`, `webkitMaskSize`, `mask-size DOM style prop`, and `vendor style prop DOM` did not find an existing matching issue.

### AI used

hirotomoyamada:
Closes #7692

## AI used

## Description

Enhance CSS compatibility by adding vendor-specific properties.


chida09 (approved):

bmthd (approved):
LGTM！

koralle (approved):

PR #7700, opened 2026-06-14
fix(lefthook): allow ignored files in format hook

### LINKED ISSUE #7699: Allow lockfile-only commits to pass the pre-commit format hook
## Description

Allow commits that only modify files ignored by `oxfmt`, such as `pnpm-lock.yaml`, to pass the pre-commit format hook.

The pre-commit `format` job currently runs `pnpm exec oxfmt {staged_files} --write`. When the staged files are all ignored by `oxfmt`, for example a lockfile-only commit, `oxfmt` exits with `Expected at least one target file. All matched files may have been excluded by ignore rules.`

This blocks valid maintenance commits that only update `pnpm-lock.yaml`.

## Proposed Solution or API

Add `--no-error-on-unmatched-pattern` to the `oxfmt` command in `lefthook.yaml`:

```yaml
run: pnpm exec oxfmt {staged_files} --write --no-error-on-unmatched-pattern
```

This keeps formatting active for supported files while allowing ignored-only staged files to pass.

## Alternatives

A wrapper script could special-case the exact error, but the built-in `oxfmt` option is simpler and less fragile.

Yes

## AI used

## Additional Information

Related #6840

hirotomoyamada:
Closes #7699

## AI used

## Description

Adds `--no-error-on-unmatched-pattern` to the pre-commit format hook so ignored-only staged files can pass the hook.

## Current behavior (updates)

A commit that only stages files ignored by `oxfmt`, such as `pnpm-lock.yaml`, fails because `oxfmt` exits with an unmatched target error.

## New behavior

The format hook still formats supported staged files, but ignored-only staged files are treated as a successful no-op.

No

## Additional Information

Verified with:

- `pnpm exec oxfmt pnpm-lock.yaml --write --no-error-on-unmatched-pattern`
- `pnpm exec oxfmt lefthook.yaml pnpm-lock.yaml --check --no-error-on-unmatched-pattern`
- `pnpm exec lefthook run pre-commit --job format --file pnpm-lock.yaml --force`

### DIFF
--- a/lefthook.yaml
+++ b/lefthook.yaml
@@ -91,7 +91,7 @@ pre-commit:
 
     - name: format
       glob: "*.{js,cjs,mjs,jsx,ts,cts,mts,tsx,d.ts,md,mdx,yaml,yml,json,html,css}"
-      run: pnpm exec oxfmt {staged_files} --write
+      run: pnpm exec oxfmt {staged_files} --write --no-error-on-unmatched-pattern
       stage_fixed: true
 
 commit-msg:


lefthook.yaml
```diff
@@ -91,7 +91,7 @@ pre-commit:
 
     - name: format
       glob: "*.{js,cjs,mjs,jsx,ts,cts,mts,tsx,d.ts,md,mdx,yaml,yml,json,html,css}"
-      run: pnpm exec oxfmt {staged_files} --write
+      run: pnpm exec oxfmt {staged_files} --write --no-error-on-unmatched-pattern
```
taroj1205, line 94: What if we also exclude format from running if it's just lockfile change?

taroj1205 (approved):
Looks good to me, but with one question

lefthook.yaml
chida09: The exclusion is already managed in config.ts via ignorePatterns.
The flag just tells oxfmt "don't error when all files happen to be excluded" — no need to duplicate the list in the glob.

ref: https://github.com/yamada-ui/yamada-ui/blob/main/packages/workspace/src/oxfmt/config.ts#L13

chida09 (approved):

lefthook.yaml
taroj1205: by excluding it in the lefthook config, we can skip the entire oxc script from running when the only change is a lock file.

odendayoko (approved):

bmthd (approved):
LGTM

PR #7842, opened 2026-09-10
fix(tabs): enhance a11y for tabs stories

### LINKED ISSUE #7603: Enhance a11y for `Tabs`
name: Tabs

### Description
1 accessibility violation was found in the `Tabs`.

#### [Components / Tabs / Dynamic](https://yamada-ui.github.io/yamada-ui/?path=/story/components-tabs--dynamic)

Received:
Certain ARIA roles must contain particular children (aria-required-children)

Fix any of the following:
  Element has children which are not allowed: button

You can find more information on this issue here: 
https://dequeuniversity.com/rules/axe/4.13/aria-required-children?application=axeAPI

### Additional Information

iimtii:
Closes #7603

## AI used

## Description

The `Dynamic` story placed the `Add Tab` button directly inside `Tabs.List`, which renders with `role="tablist"`. Because `tablist` only permits children with `role="tab"`, axe reported an `aria-required-children` violation.

## Old behavior

`Tabs.List` owned a plain `button` element alongside the tabs, so assistive technologies could skip the `Add Tab` button entirely

Received:

```
Certain ARIA roles must contain particular children (aria-required-children)

Fix any of the following:
  Element has children which are not allowed: button
```

## New behavior

`Tabs.List` and the `Add Tab` button are now siblings inside a `Flex` wrapper, so the tablist owns only `role="tab"` elements.

The underline moved from `Tabs.List` to the wrapper (`borderBottomWidth="1px"` on `Flex`, `borderBottomWidth="0"` on `Tabs.List`), keeping the appearance unchanged. 

No component or style changes were required — the violation was only happening inside how story was written.

No.

### DIFF
--- a/packages/react/src/components/tabs/tabs.stories.tsx
+++ b/packages/react/src/components/tabs/tabs.stories.tsx
@@ -5,6 +5,7 @@ import { PropsTable } from "#storybook"
 import { useInterval } from "../../hooks/use-interval"
 import { uuid } from "../../utils"
 import { Button, IconButton } from "../button"
+import { Flex } from "../flex"
 import { Heading } from "../heading"
 import { PlusIcon, XIcon } from "../icon"
 import { Text } from "../text"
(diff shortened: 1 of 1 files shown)


iimtii:
will fix the build and push in this week

hirotomoyamada (approved):

